PR #1, opened 2020-10-14
Notary Library Prototype

shizhMSFT:
1. Able to sign and verify manifests.
2. Able to push and pull signatures to / from the OCI-compatible registry.


dlorenc:
It looks like this prototype uses an oci Index to store the signature, where the signature is serialized into the Config and the "link" is established as a downward reference via a manifest. Is this still the intended approach for the Notary v2 effort?

shizhMSFT:
This PR is closing as it will be succeeded by the `prototype-2` where the signature is a regular blob instead of a config blob. /cc @SteveLasker

PR #11, opened 2021-09-24
JWS-based Notation Go Library

### LINKED ISSUE #8: Move to generic registry and client
Opening this issue to track refactoring. We might not even need registry client and can be scoped to just the signature formats.

### LINKED ISSUE #9: Implement updated signature format - JWS as per specification

### LINKED ISSUE #10: Tracking issue for pending feedbacks/Issues in PR5
Tracking issue for multiple comments/issues that were called out in https://github.com/notaryproject/notation-go-lib/pull/5 and needs to be resolved.

shizhMSFT:
This PR rewrites the previous prototype-level notation go library, and aims for production quality code.

Change logs:

- Removed registry client so that this library only focuses on signing.
  - Closes #8.
- Updated the JWT signature format for prototyping to JWS signature format.
  - Candidate specification: https://github.com/notaryproject/notaryproject/pull/93.
  - Closes #9.
  - The JWS implementation is based on [golang-jwt/jwt v4](https://github.com/golang-jwt/jwt).
    - Plugable signing and verification methods.
  - The resulted JWT is reserialized to / from JWS JSON serialization by code re-written from [shizhMSFT/go-jwsutil](https://github.com/shizhMSFT/go-jwsutil).
  - New notation library interfaces.
- Integrated timestamping to the JWS signature format.
  - Related discussion on expiry: https://github.com/notaryproject/notaryproject/discussions/98
  - Developed package `crypto/timestamp` for fetching and verifing timestamp signatures from TSAs.
    - Code is re-written from [shizhMSFT/go-timestamp](https://github.com/shizhMSFT/go-timestamp).
- Resolves #10.
- Tested by unit tests.
- Documented all the packages.
- Added a `Makefile` for easier development.
- Added GitHub actions for continuous testing.
- Added dependency bot for dependency updates.


notation.go
```diff
@@ -0,0 +1,77 @@
...
+
+// Validate does basic validation on SignOptions.
+func (opts SignOptions) Validate() error {
+	if opts.Expiry.IsZero() {
+		return ErrExpiryNotSpecified
+	}
+	return nil
+}
+
+// Signer is an generic interface for signing an artifact.
```
northtyphoon, line 49: a generic

notation.go
```diff
@@ -0,0 +1,77 @@
...
+
+// VerifyOptions contains parameters for Verifier.Verify.
+type VerifyOptions struct{}
+
+// Validate does basic validation on VerifyOptions.
+func (opts VerifyOptions) Validate() error {
+	return nil
+}
+
+// Verifier is an generic interface for verifying an artifact.
```
northtyphoon, line 66: ditto

crypto/cryptoutil/cert.go
```diff
@@ -0,0 +1,31 @@
+package cryptoutil
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"os"
+)
+
+// ReadCertificateFile reads a certificate PEM file.
```
northtyphoon, line 9: Do we plan to support DER format?

crypto/cryptoutil/key.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+	switch block.Type {
+	case "PRIVATE KEY":
+		return x509.ParsePKCS8PrivateKey(block.Bytes)
+	case "EC PRIVATE KEY":
+		return x509.ParseECPrivateKey(block.Bytes)
+	case "RSA PRIVATE KEY":
+		return x509.ParsePKCS1PrivateKey(block.Bytes)
+	}
+	return nil, errors.New("unsupported PEM block type")
```
northtyphoon, line 34: do you want to include the type name in the error message?

crypto/jwsutil/signature.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"encoding/json"
+	"strings"
+)
+
+// Signature represents a detached signature.
+type Signature struct {
+	Protected   string          `json:"protected,omitempty"`
+	Unprotected json.RawMessage `json:"header,omitempty"`
```
northtyphoon, line 12: why the serialized name and property name are different?

crypto/timestamp/timestamptest/tsa.go
```diff
@@ -0,0 +1,269 @@
...
+
+// Certificate returns the certificate used by the server.
+func (tsa *TSA) Certificate() *x509.Certificate {
+	return tsa.cert
+}
+
+// Timestamp stamps the time with the given request.
+func (tsa *TSA) Timestamp(_ context.Context, req *timestamp.Request) (*timestamp.Response, error) {
+	// validate request
+	if req.Version != 1 {
```
northtyphoon, line 90: Can you define a const?

crypto/timestamp/timestamptest/tsa.go
```diff
@@ -0,0 +1,269 @@
...
+		return responseRejection, nil
+	}
+	if hashedMessage := req.MessageImprint.HashedMessage; len(hashedMessage) != hash.Size() {
+		return responseRejection, nil
+	}
+
+	// generate token info
+	policy := asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 4146, 2} // time-stamp-policies
+	switch hash {
+	case crypto.SHA1:
```
northtyphoon, line 104: Does 2 or 3 mean anything? Or is it just a random number?

crypto/timestamp/timestamptest/tsa.go
```diff
@@ -0,0 +1,269 @@
...
+	messageDigest, err := convertToRawASN1([]interface{}{infoDigest}, "set")
+	if err != nil {
+		return cms.SignedData{}, err
+	}
+	signingTime, err := convertToRawASN1([]interface{}{time.Now().UTC()}, "set")
+	if err != nil {
+		return cms.SignedData{}, err
+	}
+	signed := cms.SignedData{
+		Version: 3,
```
northtyphoon, line 192: Is the version defined somewhere?

internal/crypto/cms/errors.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// VerificationError indicates verification failures.
+type VerificationError struct {
+	Message string
+	Detail  error
+}
+
+// Error returns error message.
+func (e VerificationError) Error() string {
```
northtyphoon, line 48: The codes look similar. Is it possible to avoid the duplicated code?

crypto/cryptoutil/cert.go
shizhMSFT: PEM is a wrapper of DER cert. If it is a DER cert, just call `x509.ParseCertificate()` directly.

crypto/jwsutil/signature.go
shizhMSFT: It is defined in [RFC 7515 7.2.1](https://datatracker.ietf.org/doc/html/rfc7515#section-7.2.1).

crypto/timestamp/timestamptest/tsa.go
shizhMSFT: It is a fixed number defined in [RFC 3161 2.4.1](https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.1). Not meaningful to define a const for this.

crypto/timestamp/timestamptest/tsa.go
shizhMSFT: Yes, they means diffrent policies.
See https://www.globalsign.com/en/repository/GlobalSign_CP_v6.7_final.pdf at page 18.

crypto/timestamp/timestamptest/tsa.go
shizhMSFT: See [RFC 5652 5.1](https://datatracker.ietf.org/doc/html/rfc5652#section-5.1). More info can be found in the `internal/crypto/cms` package.

internal/crypto/cms/errors.go
shizhMSFT: They are two types of error. Technically, we can avoid the duplicated code but will make readability worse.

Actually, similar pattern can be found in the [go built-in library code](https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/encoding/asn1/asn1.go;l=34-47).

internal/crypto/cms/signed.go
```diff
@@ -0,0 +1,216 @@
...
+// WARNING: this function doesn't do any revocation checking.
+func (d *ParsedSignedData) Verify(opts x509.VerifyOptions) ([]SignerInfo, error) {
+	if len(d.Signers) == 0 {
+		return nil, ErrSignerNotFound
+	}
+	if len(d.Certificates) == 0 {
+		return nil, ErrCertificateNotFound
+	}
+
+	intermediates := x509.NewCertPool()
```
northtyphoon, line 71: nit: maybe you can add a comment to highlight this replacement part

internal/crypto/cms/signed.go
```diff
@@ -0,0 +1,216 @@
...
+		if err != nil {
+			return VerificationError{Message: "invalid signed attributes", Detail: err}
+		}
+		signed = encoded
+	}
+	if err := cert.CheckSignature(algorithm, signed, signer.Signature); err != nil {
+		return VerificationError{Detail: err}
+	}
+
+	// verify attributes if present
```
northtyphoon, line 127: verify attributes are present

internal/encoding/asn1/common.go
```diff
@@ -0,0 +1,116 @@
...
+		// short form
+		return int(b), nil
+	case b == 0x80:
+		// Indefinite-length method is not supported.
+		return 0, ErrUnsupportedLength
+	}
+
+	// long form
+	n := int(b & 0x7f)
+	if n > 4 {
```
northtyphoon, line 96: Does it assume int is 32 bits?

signature/jws/algorithm.go
```diff
@@ -0,0 +1,49 @@
...
+	case *ecdsa.PublicKey:
+		switch key.Curve.Params().BitSize {
+		case jwt.SigningMethodES256.CurveBits:
+			return jwt.SigningMethodES256, nil
+		case jwt.SigningMethodES384.CurveBits:
+			return jwt.SigningMethodES384, nil
+		case jwt.SigningMethodES512.CurveBits:
+			return jwt.SigningMethodES512, nil
+		default:
+			return nil, errors.New("ecdsa key not recognized")
```
northtyphoon, line 43: Do you want to include the key info in the error message?

signature/jws/signer.go
```diff
@@ -0,0 +1,182 @@
...
+		return nil, errors.New("nil signing key")
+	}
+	if keyID == "" {
+		return nil, errors.New("empty signer key ID")
+	}
+
+	return &Signer{
+		method: method,
+		key:    key,
+		keyID:  keyID,
```
northtyphoon, line 96: Question: How is the KeyID used for signing and verification process since it's optional?

signature/jws/verifier.go
```diff
@@ -0,0 +1,302 @@
...
+
+// maxTimestampAccuracy specifies the max acceptable accuracy for timestamp.
+const maxTimestampAccuracy = time.Minute
+
+// VerificationKey verifies artifacts against JWS signatures.
+type VerificationKey struct {
+	// id is the key ID.
+	id string
+
+	// value is the embedded singing key key.
```
northtyphoon, line 27: dup?

signature/jws/verifier.go
```diff
@@ -0,0 +1,302 @@
...
+	TSAVerifyOptions x509.VerifyOptions
+}
+
+// NewVerifier creates a verifier with a set of trusted verification keys.
+// Callers may be interested in options in the public field of the Verifier, especially
+// VerifyOptions for setting up trusted certificates.
+func NewVerifier(keys []*VerificationKey) *Verifier {
+	indexedKeys := make(map[string]*VerificationKey)
+	for _, key := range keys {
+		if key.id != "" {
```
northtyphoon, line 99: Does the code need to fail if the id is empty?

signature/jws/verifier.go
```diff
@@ -0,0 +1,302 @@
...
+		intermediates.AddCert(cert)
+	}
+	verifyOpts := v.VerifyOptions
+	verifyOpts.Intermediates = intermediates
+	if len(verifyOpts.KeyUsages) == 0 {
+		verifyOpts.KeyUsages = []x509.ExtKeyUsage{x509.ExtKeyUsageCodeSigning}
+	}
+
+	// verify the signing certificate
+	checkTimestamp := v.EnforceExpiryValidation
```
northtyphoon, line 190: It looks the option is to check the signature was signed when the signing cert was still valid. Maybe the name can be updated to reflect the intention. Expiry makes me think it is to check if the cert has expired.

signature/jws/verifier.go
```diff
@@ -0,0 +1,302 @@
...
+	if len(envelope.Signatures) != 1 {
+		return nil, errors.New("single signature envelope expected")
+	}
+	sig := envelope.Open()
+	return &sig, nil
+}
+
+// verifyTimestamp verifies the timestamp token and returns stamped time.
+func verifyTimestamp(contentBytes, tokenBytes []byte, opts x509.VerifyOptions) (time.Time, error) {
+	token, err := timestamp.ParseSignedToken(tokenBytes)
```
northtyphoon, line 283: if the tokenBytes doesn't have timestamp, will it return error?

internal/crypto/cms/signed.go
shizhMSFT: This replacement part has been mentioned in the documentation of this method. See line 52.

internal/crypto/cms/signed.go
shizhMSFT: We only verify attributes if they are present.

signature/jws/algorithm.go
shizhMSFT: No, I would like to keep it internal as the `BitSize` is not meaningful.

signature/jws/signer.go
shizhMSFT: As discussed in the recent meeting, the `KeyID` will be removed.

signature/jws/verifier.go
shizhMSFT: Can you elaborate more?

signature/jws/verifier.go
shizhMSFT: This name is defined in https://github.com/notaryproject/notaryproject/discussions/98. We may need to change the name in the spec first.

signature/jws/verifier.go
shizhMSFT: struct removed

northtyphoon (approved):

internal/crypto/cms/signed.go
```diff
@@ -0,0 +1,216 @@
...
+	var contentInfo ContentInfo
+	if _, err := asn1.Unmarshal(data, &contentInfo); err != nil {
+		return nil, SyntaxError{Message: "invalid content info", Detail: err}
+	}
+	if !oid.SignedData.Equal(contentInfo.ContentType) {
+		return nil, ErrExpectSignedData
+	}
+
+	var signedData SignedData
+	if _, err := asn1.Unmarshal(contentInfo.Content.Bytes, &signedData); err != nil {
```
priteshbandi, line 34: We should add validation for `SignedData#Version`

priteshbandi (commented):
LGTM, some minor comments

crypto/jwsutil/envelope.go
```diff
@@ -0,0 +1,53 @@
...
+import "encoding/json"
+
+// Envelope contains a common payload signed by multiple signatures.
+type Envelope struct {
+	Payload    string      `json:"payload,omitempty"`
+	Signatures []Signature `json:"signatures,omitempty"`
+}
+
+// Size returns the number of enclosed signatures.
+func (e Envelope) Size() int {
```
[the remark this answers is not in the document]
shizhMSFT: Envelope is about signatures. Just like `Heap.Size()` is about the number of the enclosed elements. We don't need something named `Heap.ItemCount()` / `Heap.ItemSize()`.

crypto/jwsutil/envelope.go
```diff
@@ -0,0 +1,53 @@
...
+// Open opens the evelope and returns the first or default complete signature.
+func (e Envelope) Open() CompleteSignature {
+	if len(e.Signatures) == 0 {
+		return CompleteSignature{
+			Payload: e.Payload,
+		}
+	}
+	return CompleteSignature{
+		Payload:   e.Payload,
+		Signature: e.Signatures[0],
```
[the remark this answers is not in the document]
shizhMSFT: No, the documentation clearly states `returns the first or default complete signature`. Therefore, the caller should check the `Size()` first. Actaully, the verifier in `verifier.go` does that.

internal/encoding/asn1/io.go
```diff
@@ -0,0 +1,51 @@
...
+
+import "io"
+
+// ValueReader is the interface for reading a value.
+type ValueReader interface {
+	io.Reader
+	io.ByteScanner
+}
+
+// ValueReader is the interface for writing a value.
```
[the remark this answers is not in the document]
shizhMSFT: Thanks for pointing out.

signature/jws/signer.go
```diff
@@ -0,0 +1,161 @@
...
+		CertChain: s.certChain,
+	}
+
+	// timestamp JWT
+	sig, err := jwsutil.ParseCompact(compact)
+	if err != nil {
+		return nil, err
+	}
+	if s.TSA != nil {
+		token, err := s.timestamp(ctx, sig.Signature.Signature)
```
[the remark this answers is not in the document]
shizhMSFT: By default `s.TSA` is `nil`. The signature is timestamped only if the users opt in by setting `s.TSA` to some `Timestamper`.

notation.go
```diff
@@ -0,0 +1,73 @@
...
+type Metadata struct {
+	// Identity describes the artifact identity.
+	Identity string
+
+	// Attributes contains user defined attributes.
+	Attributes map[string]interface{}
+}
+
+// SignOptions contains parameters for Signer.Sign.
+type SignOptions struct {
```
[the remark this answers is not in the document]
shizhMSFT: Timestamping is currently a per signer operation. Should we make it optional as per signature signing?

signature/jws/signer.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, err
+	}
+	if status := resp.Status; status.Status != pki.StatusGranted {
+		return nil, fmt.Errorf("tsa: %d: %v", status.Status, status.StatusString)
+	}
+	tokenBytes := resp.TokenBytes()
+
+	// verify the timestamp signature
+	if _, err := verifyTimestamp(decodedSig, tokenBytes, s.TSAVerifyOptions); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
shizhMSFT: Good point. Let me consider refactoring it.

internal/crypto/cms/signed.go
shizhMSFT: Accroding to [RFC 5652 5.1](https://datatracker.ietf.org/doc/html/rfc5652#section-5.1), the `SignedData` struct can handle all versions (1,3,4,5). I don't think we need to valid the versions unless we have strong security reasons.

signature/jws/signer.go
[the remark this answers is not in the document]
shizhMSFT: @priteshbandi After taking a look into it, I don't think we need those validations. It is simply because we don't have the request stored in the artifact signature. Even if we store the request in the artifact signature, it can be simply replaced by the adversary since the request is the unprotected header and is not signed.

signature/jws/signer.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	return NewSignerWithCertificateChain(method, key, certChain)
+}
+
+// NewSignerWithCertificateChain creates a signer with the specified signing method and
+// a signing key bundled with a certificate chain.
+// The relation of the provided siging key and its certificate chain is not verified,
+// and should be verified by the caller.
+func NewSignerWithCertificateChain(method jwt.SigningMethod, key crypto.PrivateKey, certChain []*x509.Certificate) (*Signer, error) {
```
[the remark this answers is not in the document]
shizhMSFT: The cert chain could be a partial chain.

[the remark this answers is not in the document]
shizhMSFT: @priteshbandi updated code to verify the signing cert.

notation.go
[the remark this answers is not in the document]
priteshbandi: One signer implementation(like file-based one) can be reused by multiple organization. Some organizations might not want to timestamp signature because of availability concerns or something else but other might prefer to timestamp signature.
So there should be a way to for user to disable/enable time-stamping on per signature basis. One way to achieve is that only add timestamp when user explicitly specifies it
* notation sign --timestamp=${serverURL} <-- signs and timestamps
* notation sign  <-- only signs

signature/jws/signer.go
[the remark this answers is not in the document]
priteshbandi: We do need to add more validation for timestamp response to avoid MitM, replay, etc. attacks and also to be a compliant client. The validations are more from timestamp security and correctness perspective. If the timestamp signature is bad then it is of no use for users.

Haven't gone through complete rfc3161 but here are some notable mentions:
* MitM and replay attack: https://datatracker.ietf.org/doc/html/rfc3161#section-4 point 4 and 5
* Policy: https://datatracker.ietf.org/doc/html/rfc3161#section-2.2
* nonce https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.1
* PKIStatus,PKIFailureInfo https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2

NiazFK (approved):
LGTM

signature/jws/signer.go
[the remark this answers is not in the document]
shizhMSFT: Let's address this after this PR.

priteshbandi (approved):
LGTM

PR #18, opened 2021-10-18
Restrict TSA verify options

### LINKED ISSUE #16: Dont allow signer-provider or verification-provider to override the TSAVerifyOptions.
We don't want signer-provider or verification-provider to override the [TSAVerifyOptions](https://pkg.go.dev/crypto/x509#VerifyOptions) because then user can override this value and start using non time-stamping cert(like ssl) for time-stamping.

Also as per [rfc3161#section-2.3 ](https://datatracker.ietf.org/doc/html/rfc3161#section-2.3)
`The corresponding certificate MUST contain only one instance of the extended key usage field extension as defined in [RFC2459] Section 4.2.1.13 with KeyPurposeID having value:
 id-kp-timeStamping.`

I would suggest to start with not exposing `TSAVerifyOptions` and then later if need arises we can expose this option with sane defaults.

_Originally posted by @priteshbandi in https://github.com/notaryproject/notation-go-lib/pull/15#discussion_r727777892_

shizhMSFT:
All verification options, except configurating the root certificate pool, are hidden from the timestamp token verifiers.

Resolves #16.

### DIFF
--- a/crypto/timestamp/http_test.go
+++ b/crypto/timestamp/http_test.go
@@ -3,6 +3,7 @@ package timestamp
 import (
 	"bytes"
 	"context"
+	"crypto"
 	"crypto/x509"
 	"encoding/asn1"
 	"io"
@@ -13,6 +14,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/notaryproject/notation-go-lib/internal/crypto/hashutil"
 	"github.com/notaryproject/notation-go-lib/internal/crypto/pki"
 )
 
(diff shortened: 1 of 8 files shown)


SteveLasker:
@shizhMSFT, can you resolve the merge conflict?

shizhMSFT:
@SteveLasker merge conflict resolved.

SteveLasker:
@priteshbandi, :eyes: please

priteshbandi (approved):
LGTM

NiazFK (commented):
LGTM

PR #37, opened 2022-04-26
Implement plugin manager

qmuntal:
This PR implements the common code of the [plugin-extensibility](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md) spec so it will be used by `notary` and others.

It also contains unit test and integration tests with ad-hoc plugins. Coverage is around 95%.

@SteveLasker

Signed-off-by: qmuntal <qmuntaldiaz@microsoft.com>

### DIFF
--- a/plugin/manager.go
+++ b/plugin/manager.go
@@ -1,0 +1,100 @@
+package plugin
+
+import (
+	"io/fs"
+	"os"
+	"os/exec"
+	"path"
+)
+
+// commander is defined for mocking purposes.
+type commander interface {
+	Output(string, ...string) ([]byte, error)
+}
+
+type rootedCommander struct {
+	root string
+}
+
+func (c rootedCommander) Output(name string, args ...string) ([]byte, error) {
+	cmd := &exec.Cmd{
+		Path: path.Join(c.root, name),
+		Args: append([]string{path.Base(name)}, args...),
+	}
+	return cmd.Output()
+}
+
+// Manager manages plugins installed on the system.
+type Manager struct {
+	fsys  fs.FS
+	cmder commander
+}
+
+// NewManager returns a new manager.
+func NewManager() (*Manager, error) {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return nil, err
+	}
+	pluginDir := path.Join(homeDir, ".notation", "plugins")
+	return &Manager{os.DirFS(pluginDir), rootedCommander{pluginDir}}, nil
+}
+
+// Get returns a plugin on the system by its name.
+// The plugin might be incomplete if p.Err is not nil.
+func (mgr *Manager) Get(name string) (*Plugin, error) {
+	fullname, ok := mgr.isCandidate(name)
+	if !ok {
+		return nil, ErrNotFound
+	}
+	p := newPlugin(mgr.cmder, fullname, name)
+	return p, nil
+}
+
+// List produces a list of the plugins available on the system.
+// Some plugins might be incomplete if their Err is not nil.
+func (mgr *Manager) List() ([]*Plugin, error) {
+	var plugins []*Plugin
+	fs.WalkDir(mgr.fsys, ".", func(dir string, d fs.DirEntry, _ error) error {
+		if dir == "." || !d.IsDir() {
+			return nil
+		}
+		p, err := mgr.Get(d.Name())
+		if err == nil {
+			plugins = append(plugins, p)
+		}
+		return fs.SkipDir
+	})
+	return plugins, nil
+}
+
+// Command returns an "os/exec".Cmd which when .Run() will execute the named plugin.
+// The error returned is ErrNotFound if no plugin was found.
+func (mgr *Manager) Command(name string, args ...string) (*exec.Cmd, error) {
+	p, err := mgr.Get(name)
+	if err != nil {
+		return nil, err
+	}
+	if p.Err != nil {
+		return nil, p.Err
+	}
+	return exec.Command(p.Path, args...), nil
+}
+
+func (mgr *Manager) isCandidate(name string) (fullname string, ok bool) {
+	fullname = path.Join(name, "notation-"+name)
+	fi, err := fs.Stat(mgr.fsys, addExeSuffix(fullname))
+	if err != nil {
+		// Ignore any file which we cannot Stat
+		// (e.g. due to permissions or anything else).
+		return "", false
+	}
+	switch fi.Mode().Type() {
+	case 0, fs.ModeSymlink:
+		// Regular file or symlink, keep going.
+		return fullname, true
+	default:
+		// Something else, ignore.
+		return "", false
+	}
+}
--- a/plugin/metadata.go
+++ b/plugin/metadata.go
@@ -1,0 +1,58 @@
+package plugin
+
+import "errors"
+
+// Metadata provided by the plugin.
+type Metadata struct {
+	Name                      string   `json:"name"`
+	Description               string   `json:"description"`
+	Version                   string   `json:"version"`
+	URL                       string   `json:"url"`
+	SupportedContractVersions []string `json:"supported-contract-versions"`
+	Capabilities              []string `json:"capabilities"`
+}
+
+// Validate checks if the metadata is correctly populated.
+func (m *Metadata) Validate() error {
+	if m.Name == "" {
+		return errors.New("name must not be empty")
+	}
+	if m.Description == "" {
+		return errors.New("description name must not be empty")
+	}
+	if m.Version == "" {
+		return errors.New("version must not be empty")
+	}
+	if m.URL == "" {
+		return errors.New("url must not be empty")
+	}
+	if len(m.Capabilities) == 0 {
+		return errors.New("capabilities must not be empty")
+	}
+	if len(m.SupportedContractVersions) == 0 {
+		return errors.New("supported contract versions must not be empty")
+	}
+	return nil
+}
+
+// HasCapability return true if the metadata states that the
+// capability is supported.
+func (m *Metadata) HasCapability(capability string) bool {
+	for _, c := range m.Capabilities {
+		if c == capability {
+			return true
+		}
+	}
+	return false
+}
+
+// SupportsContract return true if the metadata states that the
+// major contract version is supported.
+func (m *Metadata) SupportsContract(major string) bool {
+	for _, v := range m.SupportedContractVersions {
+		if v == major {
+			return true
+		}
+	}
+	return false
+}
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -1,0 +1,56 @@
+package plugin
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"runtime"
+)
+
+// Plugin represents a potential plugin with all it's metadata.
+type Plugin struct {
+	Metadata
+
+	Path string `json:",omitempty"`
+
+	// Err is non-nil if the plugin failed one of the candidate tests.
+	Err error `json:",omitempty"`
+}
+
+// ErrNotFound by Manager.Get when the plugin is not found.
+var ErrNotFound = errors.New("plugin not found")
+
+// newPlugin determines if the given candidate is valid
+// and returns a Plugin.
+func newPlugin(cmder commander, binPath, name string) *Plugin {
+	p := &Plugin{
+		Path: addExeSuffix(binPath),
+	}
+
+	meta, err := cmder.Output(p.Path, "get-plugin-metadata")
+	if err != nil {
+		p.Err = fmt.Errorf("failed to fetch metadata: %w", err)
+		return p
+	}
+
+	if err := json.Unmarshal(meta, &p.Metadata); err != nil {
+		p.Err = fmt.Errorf("metadata can't be decoded: %w", err)
+		return p
+	}
+	if p.Name != name {
+		p.Err = fmt.Errorf("metadata name %q is not valid, must be %q", name, p.Name)
+		return p
+	}
+	if err := p.Metadata.Validate(); err != nil {
+		p.Err = fmt.Errorf("invalid metadata: %w", err)
+		return p
+	}
+	return p
+}
+
+func addExeSuffix(s string) string {
+	if runtime.GOOS == "windows" {
+		s += ".exe"
+	}
+	return s
+}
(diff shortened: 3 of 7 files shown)


gokarnm:
cc: @rgnote @shizhMSFT for review.

plugin/metadata.go
```diff
@@ -0,0 +1,58 @@
...
+
+import "errors"
+
+// Metadata provided by the plugin.
+type Metadata struct {
+	Name                      string   `json:"name"`
+	Description               string   `json:"description"`
+	Version                   string   `json:"version"`
+	URL                       string   `json:"url"`
+	SupportedContractVersions []string `json:"supported-contract-versions"`
```
rgnote, line 11: >supported-contract-versions

Should it be `supportedContractVersions`?

plugin/plugin.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+const (
+	// NamePrefix is the prefix required on all plugin binary names.
+	NamePrefix = "notation-"
+
+	// MetadataSubcommandName is the name of the plugin subcommand
+	// which must be supported by every plugin and returns the
+	// plugin metadata.
+	MetadataSubcommandName = "get-plugin-metadata"
```
rgnote, line 17: Isn't this `discover`?

plugin/manager.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (c execCommander) Output(name string, args ...string) ([]byte, error) {
+	cmd := &exec.Cmd{
+		Path: name,
+		Args: append([]string{name}, args...),
+	}
+	return cmd.Output()
+}
+
+type rootedFS struct {
```
rgnote, line 28: nit: doc comment

rgnote (commented):
Overall, looks solid

jaredpar (approved):

plugin/metadata.go
qmuntal: Good catch!

plugin/plugin.go
qmuntal: We changed it to `get-plugin-metadata` in https://github.com/notaryproject/notaryproject/pull/150

plugin/manager.go
qmuntal: Done!

qmuntal:
@jaredpar @SteveLasker as the review process is going slower than expected, I've pushed a new batch of commits since. They don't add new functionality but refine the API after the learnings I got using it in https://github.com/notaryproject/notation/pull/168.

SteveLasker:
@qmuntal, question for how we can test a set of these changes. 
Just an idea (might be a bad one), should we create a branch in notation, notation-go and [azure/notation-azure-kv](https://github.com/azure/notation-azure-kv) to test these changes?

qmuntal:
> @qmuntal, question for how we can test a set of these changes. Just an idea (might be a bad one), should we create a branch in notation, notation-go and [azure/notation-azure-kv](https://github.com/azure/notation-azure-kv) to test these changes?

I like the idea! I'll come with something and share it with you all.

plugin/plugin.go
```diff
@@ -0,0 +1,160 @@
...
+	CapabilityEnvelopeGenerator Capability = "SIGNATURE_ENVELOPE_GENERATOR"
+)
+
+// GenerateSignatureRequest contains the parameters passed in a generate-signature request.
+// All parameters are required.
+type GenerateSignatureRequest struct {
+	ContractVersion string
+	KeyName         string
+	KeyID           string
+}
```
gokarnm: Is missing the `payload` field from [spec](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#user-content-signature-generator). I'm assuming that KeyName and KeyID are in place of KeyDefinition.

plugin/plugin.go
```diff
@@ -0,0 +1,160 @@
...
+	SigningAlgorithm string `json:"signingAlgorithm"`
+
+	// Ordered list of certificates starting with leaf certificate
+	// and ending with root certificate.
+	CertificateChain []string `json:"certificateChain"`
+}
+
+// GenerateEnvelopeRequest contains the parameters passed in a generate-envelop request.
+// All parameters are required.
+type GenerateEnvelopeRequest struct {
```
gokarnm, line 113: Missing the `payload` field defined in [spec](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#user-content-signature-envelope-generator).

plugin/manager/manager.go
```diff
@@ -0,0 +1,240 @@
...
+	"path"
+	"path/filepath"
+	"runtime"
+
+	"github.com/notaryproject/notation-go/plugin"
+)
+
+const (
+	// NamePrefix is the prefix required on all plugin binary names.
+	NamePrefix = "notation-"
```
gokarnm, line 21: Nit, rename to `PluginPrefix` ?

plugin/manager/manager.go
```diff
@@ -0,0 +1,240 @@
...
+func binName(name string) string {
+	return addExeSuffix(NamePrefix + name)
+}
+
+func binPath(fsys fs.FS, name string) string {
+	base := binName(name)
+	if fsys, ok := fsys.(rootedFS); ok {
+		return filepath.Join(fsys.root, name, base)
+	}
+	return filepath.Join(name, base)
```
gokarnm, line 229: I didn't understand what this code achieves, specially line 204. I'm new to Go.

gokarnm (changes requested):
Overall structure and organization looks great, added some comments. Will make another pass tomorrow for some areas I skipped. I'm new to Go, some high level questions.
1. Some of these error messages will be presented to users through the CLI, how do we differentiate them, and case and format them correctly with more context.
2. We need to support verbose logging through out the library and CLI, is there a standard way to do that in Go? With verbose logging, we can return concise error messages to the user, and they can enable verbose logging for more details/troubleshooting.

plugin/errors.go
```diff
@@ -0,0 +1,76 @@
...
+
+type ErrorCode string
+
+const (
+	// Any of the required request fields was empty,
+	// or a value was malformed/invalid.
+	ErrorCodeValidation ErrorCode = "VALIDATION_ERROR"
+
+	// The contract version used in the request is unsupported.
+	ErrorCodeUnsupportedContractVersion ErrorCode = "UNSUPPORTED_CONTRACT_VERSION"
```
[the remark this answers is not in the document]
qmuntal: Normally people will instantiate an `ErrorCode` variable using shorthand assignment `e := plugin.ErrorCodeValidation`, so there is not prefix repetition.

It is common to add common prefixes to enums to promote discoverability. Go does not have scoped enums, so the easiest way to know which enum values are available is to type the prefix and let autocompletion do the rest.

plugin/errors.go
```diff
@@ -0,0 +1,76 @@
+package plugin
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+)
+
+var ErrUnknownCommand = errors.New("not a plugin command")
```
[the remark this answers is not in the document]
qmuntal: Wops, this variable is an unused leftover, will remove it.

plugin/manager/manager.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// Manager manages plugins installed on the system.
+type Manager struct {
+	fsys  fs.FS
+	cmder commander
+}
+
+// NewManager returns a new manager.
+func NewManager() *Manager {
```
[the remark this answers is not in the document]
qmuntal: I would rather keep `NewManager` without parameters. Once we move the config management from `notation` to `notation-go` we will be able to root the manager to whatever path reports `config`, aka default plugins dir path. I expect this to be good enough for 95% of downstream users. If someone wants to use another config directory, we can provide `NewManagerWithPath(pluginsDir)` later on.

On the other hand, I can change the implementation now to use `os.UserConfigDir` instead of `os.UserHomeDir` so we support `XDG_CONFIG_HOME`.

plugin/manager/manager.go
```diff
@@ -0,0 +1,240 @@
...
+	plugin.Metadata
+
+	Path string `json:",omitempty"`
+
+	// Err is non-nil if the plugin failed one of the candidate tests.
+	Err error `json:",omitempty"`
+}
+
+// ErrNotFound is returned by Manager.Get and Manager.Run when the plugin is not found.
+var ErrNotFound = errors.New("plugin not found")
```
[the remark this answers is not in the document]
qmuntal: Good catch. I'll leave this as a sentinel error so callers can use it as `errors.Is(err, manager.ErrNotFound)`, but I'll decorate it with the plugin never wherever it is used using `fmt.Errorf("%s: %w", pluginName, ErrNotFound)`.

plugin/manager/manager.go
qmuntal: I've found that this variable is duplicated in `plugin` and `manager`. I'll remove the later and rename the former to just `Prefix`, so users will use it as `plugin.Prefix`

plugin/manager/manager.go
qmuntal: `if fsys, ok := fsys.(rootedFS); ok {` this is an interface type cast, being `ok` a flag which is true when `fsys` can be casted to `rootedFS`.

In production `fsys` type is always `rootedFS`, as this is what `NewManager` instantiates, but for testing purposes I'm also supporting generic `fsys` types, aka `fs.FS`, which are mostly instantiated from `testing/fstest` package.

qmuntal:
> Overall structure and organization looks great, added some comments. Will make another pass tomorrow for some areas I skipped. I'm new to Go, some high level questions.
> 
> 1. Some of these error messages will be presented to users through the CLI, how do we differentiate them, and case and format them correctly with more context.
> 2. We need to support verbose logging through out the library and CLI, is there a standard way to do that in Go? With verbose logging, we can return concise error messages to the user, and they can enable verbose logging for more details/troubleshooting.

Thanks for the review @gokarnm. My take is that `notation-go` should provide fine-grained error messaged which are good for displaying but also can be captured programmatically. If an error is missing some context, then we should add it.

What I wouldn't support is verbose levels in the returned errors. It either returns an error or not, leaving to the caller to decide if it should be logged as a debug, info, warning or whatever other level.

The build-in `log` package does not have support for using log levels, but there are multiple well-known packages with does support it, e.g. github.com/sirupsen/logrus.

plugin/manager/manager.go
gokarnm: Got it. Should we add a code comment here to make this clearer, that the alternate path is to support testing?

gokarnm:
> Thanks for the review @gokarnm. My take is that `notation-go` should provide fine-grained error messaged which are good for displaying but also can be captured programmatically. If an error is missing some context, then we should add it.

What about the error message casing, error messages which are returned to CLI users should be correctly cased. It seems to be a convention to use small case error messages in Go?

> What I wouldn't support is verbose levels in the returned errors. It either returns an error or not, leaving to the caller to decide if it should be logged as a debug, info, warning or whatever other level.

Agreed, I wasn't clear enough. I meant verbose level should log to stdout so that it's visible to in CLI or an upstream caller if they enable verbose logging. The verbose logging will need to be supported in notation-go and notation-core-go, as that's where bulk of the decision paths and validations will be. This need not be implemented right away.

plugin/manager/manager.go
```diff
@@ -0,0 +1,269 @@
...
+	fi, err := fs.Stat(fsys, path.Join(name, base))
+	if err != nil {
+		// Ignore any file which we cannot Stat
+		// (e.g. due to permissions or anything else).
+		return false
+	}
+	if fi.Mode().Type() != 0 {
+		// Ignore non-regular files.
+		return false
+	}
```
gokarnm, line 223: We should additionally check if the containing dir `~/.notation/plugins/{plugin-name}/notation-{plugin-name}` is also not a symlink, something that was missed in the spec.

plugin/manager/manager.go
```diff
@@ -0,0 +1,269 @@
...
+
+// List produces a list of the plugins available on the system.
+//
+// Some plugins might be incomplete if their Err is not nil.
+func (mgr *Manager) List(ctx context.Context) ([]*Plugin, error) {
+	var plugins []*Plugin
+	fs.WalkDir(mgr.fsys, ".", func(dir string, d fs.DirEntry, _ error) error {
+		if dir == "." || !d.IsDir() {
+			return nil
+		}
```
gokarnm, line 104: We should skip over plugin dirs that are symlinks, something that was missed in the spec.

gokarnm (approved):
LGTM for this iteration. Added a couple of comments to not allow symlinks for plugin dir, that can be addressed in next revision, when we add support for `describe-key`.

plugin/manager/manager.go
qmuntal: Done!

qmuntal:
> What about the error message casing, error messages which are returned to CLI users should be correctly cased. It seems to be a convention to use small case error messages in Go?

There are multiple options to transform and localize text, being https://pkg.go.dev/golang.org/x/text the most common.

> Agreed, I wasn't clear enough. I meant verbose level should log to stdout so that it's visible to in CLI or an upstream caller if they enable verbose logging. The verbose logging will need to be supported in notation-go and notation-core-go, as that's where bulk of the decision paths and validations will be. This need not be implemented right away.

If we need to log something from those libraries, which is still not clear to me if we should, we will find the way. For example, [http.Server](https://pkg.go.dev/net/http#Server) logs error into the optional `http.Server.ErrorLog` property.

plugin/manager/manager.go
qmuntal: This is exactly what it is being checked here. Non-regular files include symlinks and directories.

plugin/manager/manager.go
qmuntal: Done!

plugin/manager/manager.go
gokarnm: I'm bit confused here. Isn't this code only checking that the plugin executable (`notation-{plugin-name}`) itself is not a symlink, and not that additionally the enclosing directory is also not a symlink (`~/.notation/plugins/{plugin-name}` dir)?

gokarnm:
> If we need to log something from those libraries, which is still not clear to me if we should, we will find the way. For example, [http.Server](https://pkg.go.dev/net/http#Server) logs error into the optional `http.Server.ErrorLog` property.

Yes, that's what I'm looking for, similar to libraries that support emitting verbose logs to aid debugging issues. It would be useful for us to debug user reported errors as the logic in many of these areas, plugin and signature verification logic, is quite complex.

PR #39, opened 2022-04-27
Load and Validate Trust Policy documents

rgnote:
* Implementing a parser for trust policy documents as specified in https://github.com/notaryproject/notaryproject/pull/149/files?short_path=bab1489#diff-bab14897801fe47838fadd7f0ebd867e81ce66e098785175f4c1ec368bbe406a
* Added unit tests and manually tested locally 
cc: @shizhMSFT @SteveLasker

### DIFF
--- a/verification/policy.go
+++ b/verification/policy.go
@@ -96,7 +96,6 @@ func ValidatePolicyDocument(policyDoc *PolicyDocument) error {
 			i := strings.Index(statement.TrustStore, ":")
 			if i < 0 || !isPresent(statement.TrustStore[:i], supportedTrustStorePrefixes) {
 				return fmt.Errorf("trust policy statement %q uses an unsupported trust store type %q in trust store value %q", statement.Name, statement.TrustStore[:i], statement.TrustStore)
-				return fmt.Errorf("%q has a trust store with an unsupported trust store type", statement.Name)
 			}
 		}
 
@@ -109,7 +108,6 @@ func ValidatePolicyDocument(policyDoc *PolicyDocument) error {
 		// If there is a wildcard in trusted identies, there shouldn't be any other identities
 		if len(statement.TrustedIdentities) > 1 && isPresent(wildcard, statement.TrustedIdentities) {
 			return fmt.Errorf("trust policy statement %q uses a wildcard trusted identity '*', a wildcard identity cannot be used in conjunction with other values", statement.Name)
-			return errors.New("wildcard trusted identity can not be shared with other identities")
 		}
 	}
 
--- a/verification/policy.go
+++ b/verification/policy.go
@@ -1,0 +1,153 @@
+/*
+Package verification provides the utilities for handling verification related logic
+like Trust Stores and Trust Policies. Few utilities include loading, parsing, and validating
+trust policies and trust stores.
+*/
+package verification
+
+import (
+	"encoding/json"
+	"errors"
+	"io/ioutil"
+	"os"
+	"strings"
+)
+
+// PolicyDocument represent a trust_policy.json document
+type PolicyDocument struct {
+	// Version of the policy document
+	Version string `json:"version"`
+	// TrustPolicies include each policy statement
+	TrustPolicies []TrustPolicy `json:"trustPolicies"`
+}
+
+type TrustPolicy struct {
+	// Name of the policy statement
+	Name string `json:"name"`
+	// RegistryScopes that this policy statement affects
+	RegistryScopes []string `json:"registryScopes"`
+	// SignatureVerification setting for this policy statement
+	SignatureVerification string `json:"signatureVerification"`
+	// TrustStore this policy stament uses
+	TrustStore string `json:"trustStore,omitempty"`
+	// TrustedIdentities this policy statement pins
+	TrustedIdentities []string `json:"trustedIdentities,omitempty"`
+}
+
+func isPresent(val string, values []string) bool {
+	for _, v := range values {
+		if v == val {
+			return true
+		}
+	}
+	return false
+}
+
+// ValidatePolicyDocument validates a policy document according to it's version's rule set.
+// if any rule is violated, returns an error
+func ValidatePolicyDocument(policyDoc *PolicyDocument) error {
+	// Constants
+	wildcardScope := "*"
+	supportedPolicyVersions := []string{"1.0"}
+	supportedVerificationPresets := []string{"strict", "permissive", "audit", "skip"}
+	supportedTrustStorePrefixes := []string{"ca", "signingservice"}
+
+	// Validate Version
+	if !isPresent(policyDoc.Version, supportedPolicyVersions) {
+		return errors.New("Version '" + policyDoc.Version + "' is not supported")
+	}
+
+	// Validate the policy according to 1.0 rules
+	if len(policyDoc.TrustPolicies) == 0 {
+		return errors.New("Trust Policy document can not have zero statements")
+	}
+	var uniqueStatementNames []string
+	registryScopeCount := make(map[string]int)
+	for _, statement := range policyDoc.TrustPolicies {
+
+		// Verify statement name is valid
+		if statement.Name == "" {
+			return errors.New("Policy statement is missing a name")
+		} else {
+			if !isPresent(statement.Name, uniqueStatementNames) {
+				uniqueStatementNames = append(uniqueStatementNames, statement.Name)
+			}
+		}
+
+		// Verify registry scopes are valid
+		if len(statement.RegistryScopes) == 0 {
+			return errors.New("Policy statement has zero registry scopes")
+		} else {
+			if len(statement.RegistryScopes) > 1 && isPresent(wildcardScope, statement.RegistryScopes) {
+				return errors.New("Wildcard scope can not be shared with other registry scopes")
+			}
+			for _, scope := range statement.RegistryScopes {
+				registryScopeCount[scope]++
+			}
+		}
+
+		// Verify signature verification preset is valid
+		if !isPresent(statement.SignatureVerification, supportedVerificationPresets) {
+			return errors.New("SignatureVerification '" + statement.SignatureVerification + "' is not supported")
+		}
+
+		// strict and permissive verification needs trust store to verify authenticity
+		if statement.SignatureVerification == "strict" || statement.SignatureVerification == "permissive" {
+			if statement.TrustStore == "" {
+				return errors.New("Verification statement with strict or permissive preset is missing a trust store")
+			}
+		}
+
+		// Verify trust store type is valid
+		if statement.TrustStore != "" {
+			i := strings.Index(statement.TrustStore, ":")
+			if i < 0 || !isPresent(statement.TrustStore[:i], supportedTrustStorePrefixes) {
+				return errors.New("Statement '" + statement.Name + "' has a trust store with an unsupported trust store type")
+			}
+		}
+
+		// If there are trusted identitied, verify they are not empty
+		if len(statement.TrustedIdentities) > 0 {
+			for _, identity := range statement.TrustedIdentities {
+				if identity == "" {
+					return errors.New("Policy statement has an empty trusted identity")
+				}
+			}
+		}
+	}
+
+	// Verify unique policy statement names across the policy document
+	if len(policyDoc.TrustPolicies) != len(uniqueStatementNames) {
+		return errors.New("Multiple policy statements have the same name")
+	}
+
+	// Verify one policy statement per registry scope
+	for key := range registryScopeCount {
+		if registryScopeCount[key] > 1 {
+			return errors.New("Registry '" + key + "' is present in multiple statements")
+		}
+	}
+
+	// No errors
+	return nil
+}
+
+// LoadPolicyDocument loads a policy document from the given path
+// If successful, returns a pointer to the PolicyDocument. Otherwise, an error
+func LoadPolicyDocument(path string) (*PolicyDocument, error) {
+	jsonFile, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer jsonFile.Close()
+
+	byteValue, _ := ioutil.ReadAll(jsonFile)
+	var policyDoc *PolicyDocument
+
+	err = json.Unmarshal(byteValue, &policyDoc)
+	if err != nil {
+		return nil, err
+	}
+
+	return policyDoc, nil
+}
--- a/verification/policy.go
+++ b/verification/policy.go
@@ -6,14 +6,12 @@ trust policies and trust stores.
 package verification
 
 import (
-	"encoding/json"
 	"errors"
-	"io/ioutil"
-	"os"
+	"fmt"
 	"strings"
 )
 
-// PolicyDocument represent a trust_policy.json document
+// PolicyDocument represents a trust_policy.json document
 type PolicyDocument struct {
 	// Version of the policy document
 	Version string `json:"version"`
@@ -21,6 +19,7 @@ type PolicyDocument struct {
 	TrustPolicies []TrustPolicy `json:"trustPolicies"`
 }
 
+// TrustPolicy represents a policy statement in the policy document
 type TrustPolicy struct {
 	// Name of the policy statement
 	Name string `json:"name"`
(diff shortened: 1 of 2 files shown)


SteveLasker (approved):
Great to see the Trust Policy work coming
LGTM

verification/policy.go
```diff
@@ -0,0 +1,153 @@
...
+		return errors.New("Trust Policy document can not have zero statements")
+	}
+	var uniqueStatementNames []string
+	registryScopeCount := make(map[string]int)
+	for _, statement := range policyDoc.TrustPolicies {
+
+		// Verify statement name is valid
+		if statement.Name == "" {
+			return errors.New("Policy statement is missing a name")
+		} else {
```
qmuntal, line 71: This else statement is not necessary. If `Name` is empty the functions return.

verification/policy.go
```diff
@@ -0,0 +1,153 @@
...
+			if !isPresent(statement.Name, uniqueStatementNames) {
+				uniqueStatementNames = append(uniqueStatementNames, statement.Name)
+			}
+		}
+
+		// Verify registry scopes are valid
+		if len(statement.RegistryScopes) == 0 {
+			return errors.New("Policy statement has zero registry scopes")
+		} else {
```
qmuntal, line 80: This else statement is not necessary. If `RegistryScopes` is empty the functions return.

verification/policy.go
```diff
@@ -0,0 +1,153 @@
...
+	jsonFile, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer jsonFile.Close()
+
+	byteValue, _ := ioutil.ReadAll(jsonFile)
+	var policyDoc *PolicyDocument
+
+	err = json.Unmarshal(byteValue, &policyDoc)
```
qmuntal, line 147: You can decode a json object directly from the file by doing `err = json.NewDecoder(jsonFile).Decode(policyDoc)`

verification/policy.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+	}
+
+	// No errors
+	return nil
+}
+
+// LoadPolicyDocument loads a policy document from the given path
+// If successful, returns a pointer to the PolicyDocument. Otherwise, an error
+func LoadPolicyDocument(path string) (*PolicyDocument, error) {
```
qmuntal, line 137: Do we really need to expose this function? It is just a thin wrapper around `os.Open` and `json.Marshal`, which is trivial to implement from outside but increases our API surface, which is bad.

qmuntal (changes requested):
Two general comments:
- First letter in error messages is normally not capitalized\
- use [fmt.Errorf](https://pkg.go.dev/fmt#Errorf) instead of `errors.New` to construct concatenated error messages.

verification/policy.go
```diff
@@ -0,0 +1,153 @@
...
+
+// PolicyDocument represent a trust_policy.json document
+type PolicyDocument struct {
+	// Version of the policy document
+	Version string `json:"version"`
+	// TrustPolicies include each policy statement
+	TrustPolicies []TrustPolicy `json:"trustPolicies"`
+}
+
+type TrustPolicy struct {
```
qmuntal, line 24: Comment missing.

verification/policy.go
```diff
@@ -0,0 +1,153 @@
...
+
+		// strict and permissive verification needs trust store to verify authenticity
+		if statement.SignatureVerification == "strict" || statement.SignatureVerification == "permissive" {
+			if statement.TrustStore == "" {
+				return errors.New("Verification statement with strict or permissive preset is missing a trust store")
+			}
+		}
+
+		// Verify trust store type is valid
+		if statement.TrustStore != "" {
```
qmuntal, line 102: nit: this is statement is not necessary, if `TrustStore` is empty `strings.Index` will return `-1`.

verification/policy.go
```diff
@@ -0,0 +1,153 @@
...
+			i := strings.Index(statement.TrustStore, ":")
+			if i < 0 || !isPresent(statement.TrustStore[:i], supportedTrustStorePrefixes) {
+				return errors.New("Statement '" + statement.Name + "' has a trust store with an unsupported trust store type")
+			}
+		}
+
+		// If there are trusted identitied, verify they are not empty
+		if len(statement.TrustedIdentities) > 0 {
```
qmuntal, line 110: This if statement is not necessary, Go will not loop if `TrustedIdentities` is empty.

qmuntal (changes requested):

verification/policy_test.go
```diff
@@ -0,0 +1,198 @@
...
+func dummyPolicyDocument() (policyDoc PolicyDocument) {
+	policyDoc = PolicyDocument{
+		Version:       "1.0",
+		TrustPolicies: []TrustPolicy{dummyPolicyStatement()},
+	}
+	return
+}
+
+func writeToTempFile(policyDoc PolicyDocument) string {
+	tmpFile, _ := ioutil.TempFile(os.TempDir(), "notation-test-")
```
qmuntal, line 30: nit: If you use [testing.T#TempDir](https://pkg.go.dev/testing#T.TempDir) instead of `os.TempDir` Go will make sure the temp directory is removed, so you won't have to call `os.Remove` later on.

SteveLasker:
thanks @qmuntal for the great feedback

verification/policy.go
rgnote: Agreed. Wasn't aware it's as simple as `os.Open()` and `json.NewDecoder(jsonFile).Decode(policyDoc)`

verification/policy.go
rgnote: Need to keep this one because we need to differentiate between empty trust store and a trust store with an invalid type. If we remove the `if` statement, then statements without a trust store (valid case for `skip` verification), we would throw *unsupported trust store type* error

SteveLasker:
@rgnote, looks like there's a DCO error

verification/policy.go
SteveLasker: @rgnote, can you add that feedback to the comment, so the next person understands the reasoning?

rgnote: Will do

verification/policy.go
```diff
@@ -47,107 +46,85 @@ func isPresent(val string, values []string) bool {
...
-	wildcardScope := "*"
+	wildcard := "*"
 	supportedPolicyVersions := []string{"1.0"}
 	supportedVerificationPresets := []string{"strict", "permissive", "audit", "skip"}
 	supportedTrustStorePrefixes := []string{"ca", "signingservice"}
 
 	// Validate Version
 	if !isPresent(policyDoc.Version, supportedPolicyVersions) {
-		return errors.New("Version '" + policyDoc.Version + "' is not supported")
+		return fmt.Errorf("version '%s' is not supported", policyDoc.Version)
```
qmuntal, line 56: nit: `'%s'` can be replaced by `%q`, which automatically quotes strings.

verification/policy.go
```diff
@@ -47,107 +46,85 @@ func isPresent(val string, values []string) bool {
...
+			return errors.New("wildcard registry scope can not be shared with other registries")
+		}
+		for _, scope := range statement.RegistryScopes {
+			registryScopeCount[scope]++
 		}
 
 		// Verify signature verification preset is valid
 		if !isPresent(statement.SignatureVerification, supportedVerificationPresets) {
-			return errors.New("SignatureVerification '" + statement.SignatureVerification + "' is not supported")
+			return fmt.Errorf("signatureVerification '%s' is not supported", statement.SignatureVerification)
```
qmuntal, line 86: nit: `'%s'` can be replaced by `%q`, which automatically quotes strings.

qmuntal (approved):

verification/policy.go
rgnote: Addressed

verification/policy.go
rgnote: Addressed

verification/policy.go
```diff
@@ -0,0 +1,130 @@
...
+			return errors.New("policy statement is missing a name")
+		}
+		policyStatementNameCount[statement.Name]++
+
+		// Verify registry scopes are valid
+		if len(statement.RegistryScopes) == 0 {
+			return fmt.Errorf("policy statement %q has zero registry scopes", statement.Name)
+		}
+		if len(statement.RegistryScopes) > 1 && isPresent(wildcard, statement.RegistryScopes) {
+			return errors.New("wildcard registry scope can not be shared with other registries")
```
gokarnm, line 78:
```suggestion
                        return fmt.Errorf("trust policy statement %q uses wildcard registry scope '*', a wildcard scope cannot be used in conjunction with other scope values, it must be the only single value for registry scope", statement.Name)
			return errors.New("wildcard registry scope can not be shared with other registries")
```

verification/policy.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+		for _, scope := range statement.RegistryScopes {
+			registryScopeCount[scope]++
+		}
+
+		// Verify signature verification preset is valid
+		if !isPresent(statement.SignatureVerification, supportedVerificationPresets) {
+			return fmt.Errorf("signatureVerification %q is not supported", statement.SignatureVerification)
```
gokarnm, line 86:
```suggestion
			return fmt.Errorf("trust policy statement %q uses unsupported signatureVerification value %q", statement.Name, statement.SignatureVerification)
```

verification/policy.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+
+		// Any signature verification other than "skip" needs a trust store
+		if statement.SignatureVerification != "skip" && (statement.TrustStore == "" || len(statement.TrustedIdentities) == 0) {
+			return fmt.Errorf("%q is either missing a trust store or trusted identities", statement.Name)
```
gokarnm, line 91:
```suggestion
			return fmt.Errorf("trust policy statement %q is either missing a trust store or trusted identities, both must be specified", statement.Name)
```

verification/policy.go
gokarnm: It would be better to combine this and previous validation as
```
if signatureverification is skip
  truststore and trustedentities must be empty
else 
  both should have valid values
```

verification/policy.go
```diff
@@ -0,0 +1,130 @@
...
+			i := strings.Index(statement.TrustStore, ":")
+			if i < 0 || !isPresent(statement.TrustStore[:i], supportedTrustStorePrefixes) {
+				return fmt.Errorf("%q has a trust store with an unsupported trust store type", statement.Name)
+			}
+		}
+
+		// If there are trusted identities, verify they are not empty
+		for _, identity := range statement.TrustedIdentities {
+			if identity == "" {
+				return fmt.Errorf("%q has an empty trusted identity", statement.Name)
```
gokarnm, line 105:
```suggestion
				return fmt.Errorf("trust policy statement %q has an empty trusted identity", statement.Name)
```

verification/policy.go
```diff
@@ -0,0 +1,130 @@
...
+		// Verify trust store type is valid if it is present (trust store is optional for "skip" signature verification)
+		if statement.TrustStore != "" {
+			i := strings.Index(statement.TrustStore, ":")
+			if i < 0 || !isPresent(statement.TrustStore[:i], supportedTrustStorePrefixes) {
+				return fmt.Errorf("%q has a trust store with an unsupported trust store type", statement.Name)
+			}
+		}
+
+		// If there are trusted identities, verify they are not empty
+		for _, identity := range statement.TrustedIdentities {
```
gokarnm, line 103: Would be better to move trusted identities validation into its own method. Validation for trusted identity `x509.subject` and valid and minimum RDNs is missing.

https://github.com/notaryproject/notaryproject/blob/main/trust-store-trust-policy-specification.md#trusted-identities-constraints

verification/policy.go
```diff
@@ -0,0 +1,130 @@
...
+	for key := range policyStatementNameCount {
+		if policyStatementNameCount[key] > 1 {
+			return fmt.Errorf("multiple policy statements with same name : %s", key)
+		}
+	}
+
+	// Verify one policy statement per registry scope
+	for key := range registryScopeCount {
+		if registryScopeCount[key] > 1 {
+			return fmt.Errorf("registry %q is present in multiple policy statements", key)
```
gokarnm, line 124:
```suggestion
			return fmt.Errorf("registry scope %q is present in multiple trust policy statements, every registry scope value can only be associated with one statement", key)
```

verification/policy.go
```diff
@@ -0,0 +1,130 @@
...
+		// If there is a wildcard in trusted identies, there shouldn't be any other identities
+		if len(statement.TrustedIdentities) > 1 && isPresent(wildcard, statement.TrustedIdentities) {
+			return errors.New("wildcard trusted identity can not be shared with other identities")
+		}
+	}
+
+	// Verify unique policy statement names across the policy document
+	for key := range policyStatementNameCount {
+		if policyStatementNameCount[key] > 1 {
+			return fmt.Errorf("multiple policy statements with same name : %s", key)
```
gokarnm, line 117:
```suggestion
			return fmt.Errorf("multiple trust policy statements use the same name %q, statement names must be unique", key)
```

verification/policy.go
```diff
@@ -0,0 +1,130 @@
...
+		// Any signature verification other than "skip" needs a trust store
+		if statement.SignatureVerification != "skip" && (statement.TrustStore == "" || len(statement.TrustedIdentities) == 0) {
+			return fmt.Errorf("%q is either missing a trust store or trusted identities", statement.Name)
+		}
+
+		// Verify trust store type is valid if it is present (trust store is optional for "skip" signature verification)
+		if statement.TrustStore != "" {
+			i := strings.Index(statement.TrustStore, ":")
+			if i < 0 || !isPresent(statement.TrustStore[:i], supportedTrustStorePrefixes) {
+				return fmt.Errorf("%q has a trust store with an unsupported trust store type", statement.Name)
```
gokarnm, line 98:
```suggestion
				return fmt.Errorf("trust policy statement %q uses an unsupported trust store type %q in trust store value %q", statement.Name, statement.TrustStore[:i], statement.TrustStore)
```

verification/policy.go
```diff
@@ -0,0 +1,130 @@
...
+
+		// If there are trusted identities, verify they are not empty
+		for _, identity := range statement.TrustedIdentities {
+			if identity == "" {
+				return fmt.Errorf("%q has an empty trusted identity", statement.Name)
+			}
+		}
+		// If there is a wildcard in trusted identies, there shouldn't be any other identities
+		if len(statement.TrustedIdentities) > 1 && isPresent(wildcard, statement.TrustedIdentities) {
+			return errors.New("wildcard trusted identity can not be shared with other identities")
```
gokarnm, line 110:
```suggestion
                        return fmt.Errorf("trust policy statement %q uses a wildcard trusted identity '*', a wildcard identity cannot be used in conjunction with other values, it must be the only single value for trusted identity", statement.Name)
```

verification/policy.go
gokarnm: Registry scope validation rules could be moved to its own method. Missing the validation on allowed values, either a `*` or fully qualified registry URL without the scheme/protocol. E.g. `registry.acme-rockets.io/software/net-monitor`

gokarnm (changes requested):

verification/policy.go
rgnote: Will add URL validation. Regarding the error message, will use `trust policy statement %q uses wildcard registry scope '*', a wildcard scope cannot be used in conjunction with other scope values", statement.Name`

verification/policy.go
rgnote: The code becomes more complex that way
```
if signatureverification is skip
  if truststore and trustedentities are not empty
    throw error
else
  if truststore is empty
    throw error
  if truststore is not valid
    throw error
  if trustedidentities are zero
    throw error
  if trustedidentities are invalid
    throw error
```

verification/policy.go
rgnote: Will address this in a separate PR

verification/policy.go
rgnote: Will change it to `"trust policy statement %q uses a wildcard trusted identity '*', a wildcard identity cannot be used in conjunction with other values`

verification/policy.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// ValidatePolicyDocument validates a policy document according to it's version's rule set.
+// if any rule is violated, returns an error
+func ValidatePolicyDocument(policyDoc *PolicyDocument) error {
+	// Constants
+	wildcard := "*"
+	supportedPolicyVersions := []string{"1.0"}
+	supportedVerificationPresets := []string{"strict", "permissive", "audit", "skip"}
+	supportedTrustStorePrefixes := []string{"ca"}
```
gokarnm, line 52: Would be nice to change these into types in next PR or when we implement verification logic. The verification logic will refer these values.

See https://github.com/notaryproject/notation-go/blob/main/plugin/errors.go#L9 as an example.

verification/policy.go
gokarnm: Agreed, but it can me made to read easier by moving trustStore, trustIdentity validation as separate methods. The current logic is more difficult to follow, we've missed the case that skip must not specify trustStore and trustedIdentity, it could be kept optional but that is confusing to users. Can be addressed in next PR.

gokarnm (approved):

PR #40, opened 2022-04-27
Implement NewSignerFromCertificate

qmuntal:
This PR moves all the exported functions from [github.com/notaryproject/notation/pkg/signature](https://github.com/notaryproject/notation/blob/main/pkg/signature/jws.go) to `notation-go/signature`. After this PR is merged, that package can be deleted.

It also generalizes the API, so it is easier to test and to use from external plugins.

Open question: I think that with this PR we can delete the [github.com/notaryproject/crypto/cryptoutil](https://github.com/notaryproject/notation-go/tree/main/crypto/cryptoutil) package, as once `notation` starts using the new APIs, it won't depend on `cryptoutil` anymore. Do you know of someone else who is using `cryptoutil`?

@shizhMSFT @SteveLasker


signature/jws/signer.go
```diff
@@ -86,6 +87,26 @@ func NewSignerWithCertificateChain(method jwt.SigningMethod, key crypto.PrivateK
...
+			return nil, err
+		}
+		certs[i] = cert
+	}
+	key := cert.PrivateKey
+	method, err := SigningMethodFromKey(key)
+	if err != nil {
+		return nil, err
+	}
+	return NewSignerWithCertificateChain(method, key, certs)
```
jaredpar, line 107: This call is going to fail if `len(certs) == 0`. This isn't wrong but seems like a lot of work above is wasted in this case. Did you consider verifying `len(certs) > 0` at the start of this method?

signature/jws/signer_test.go
```diff
@@ -176,3 +177,35 @@ func generateKeyCertPair() (*rsa.PrivateKey, *x509.Certificate, error) {
...
+	key, cert, err := generateKeyCertPair()
+	if err != nil {
+		t.Fatalf("generateKeyCertPair() error = %v", err)
+	}
+	tlscert := tls.Certificate{
+		Certificate:                  [][]byte{cert.Raw},
+		PrivateKey:                   key,
+		SupportedSignatureAlgorithms: []tls.SignatureScheme{tls.ECDSAWithP521AndSHA512},
+	}
+	s, err := NewSignerFromCertificate(tlscert)
```
jaredpar, line 191: Consider testing the case where `len(tlscert.Certificate) == 0` to ensure the error case is handled correctly

signature/jws/verifier.go
```diff
@@ -53,6 +54,31 @@ func NewVerifier() *Verifier {
 	return &Verifier{}
 }
 
+// AddCertPEM decodes data as a PEM certificate
+// and adds it to v.VerifyOptions.Roots.
+func (v *Verifier) AddCertPEM(data []byte) error {
```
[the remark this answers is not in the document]
jaredpar: Overall this method looks very similar to `AppendCertsFromPEM`. Is there a reason we can't use that?

jaredpar (approved):

SteveLasker (approved):
LGTM
Thank you, @jaredpar

signature/jws/signer.go
qmuntal: > Did you consider verifying `len(certs) > 0` at the start of this method?

I didn't, but now I will 😄

signature/jws/signer_test.go
qmuntal: Done!

signature/jws/verifier.go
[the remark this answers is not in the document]
qmuntal: Wow, good catch. I didn't know about that function, and I think it can replace our own implementation.

SteveLasker:
@qmuntal, with changes across multiple repos, is there a staging place where we can pull a collection of the binaries and validate end to end?

qmuntal:
I've decided to retract this PR after thinking about the implications of @jaredpar and @rgnote comments.

My initial goal was to factor out `NewSignerFromCertificate` and `Verifier.AddCertPEM` from `notation` to `notation-go`, so it can be reused by other applications. What I didn't saw when doing the PR is that `notation` code could be simplified by reusing functions and methods from the standard library, leaving `NewSignerFromCertificate` and `Verifier.AddCertPEM` without much value.

I'll submit a PR to `notation` with the identified simplifications and submit another PR to `notation-go` adding documentation on tip about how to use our existing API to create signers and verifiers from PEM files. Thanks everyone!

PR #42, opened 2022-05-05
Implement generate signature plugin workflow

qmuntal:
This PR contains the following changes:
- Add support for `describe-key` command
- Expand `generate-signature` request to include `keySpec`, `hashAlgorithm` and `configMap`
- Implement plugin signer with `generate-signature` workflow support
- Unit tests for everything

This PR still doesn't implement:
- Setting the right contract version in all plugin command
- `generate-signature` workflow

@SteveLasker @shizhMSFT @gokarnm @jaredpar


signature/jws/spec.go
```diff
@@ -47,3 +52,62 @@ func packPayload(desc notation.Descriptor, opts notation.SignOptions) *payload {
...
+		if ext == extKeyUsage {
+			hasExtKeyUsage = true
+			break
+		}
+	}
+	if !hasExtKeyUsage {
+		return fmt.Errorf("extKeyUsage must contain be %d", extKeyUsage)
+	}
+	for _, ext := range cert.Extensions {
+		switch ext.Id[3] {
```
jaredpar, line 92: Are we guaranteed there are at least 4 elements here? Could not find that this was guaranteed looking through the docs.

qmuntal: Good catch. Looking at this again I've noticed that I'm missing some more `Id` checks and unit tests. Will fix it.

signature/jws/spec.go
```diff
@@ -47,3 +52,62 @@ func packPayload(desc notation.Descriptor, opts notation.SignOptions) *payload {
...
+		case 15:
+			if !ext.Critical {
+				return errors.New("the keyUsage extension must be marked critical")
+			}
+		}
+	}
+	if cert.BasicConstraintsValid && cert.IsCA {
+		return errors.New("if the basicConstraints extension is present, the cA field MUST be set false")
```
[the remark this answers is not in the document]
qmuntal: It should be `CA`, copy paste typo.

jaredpar (approved):

jaredpar (approved):

qmuntal:
@shizhMSFT @rgnote @gokarnm @SteveLasker this is ready to be reviewed.

@jaredpar thanks for the review!

qmuntal:
Plugin spec updated with latest changes in https://github.com/notaryproject/notaryproject/pull/155

signature/jws/signer.go
```diff
@@ -87,58 +81,56 @@ func NewSignerWithCertificateChain(method jwt.SigningMethod, key crypto.PrivateK
...
+			"cty": signature.MediaTypeJWSEnvelope,
+		},
+		Claims: claims,
+	}
+}
 
-	// generate unprotected header
-	header := unprotectedHeader{
-		CertChain: s.certChain,
+func jwtEnvelop(ctx context.Context, opts notation.SignOptions, compact string, certChain []string) ([]byte, error) {
```
shizhMSFT, line 109:
```suggestion
func jwtEnvelope(ctx context.Context, opts notation.SignOptions, compact string, certChain []string) ([]byte, error) {
```

crypto/jwsutil/envelope.go
```diff
@@ -1,53 +0,0 @@
-package jwsutil
```
shizhMSFT, line 1: May I know why the `jwsutil` package is removed?

shizhMSFT: Note that this package is implemented according to RFC 7515 and it is not tied to the notation specs.

signature/jws/spec.go
```diff
@@ -4,46 +4,97 @@
...
+	return nil
+}
+
+var (
+	oidExtensionKeyUsage = []int{2, 5, 29, 15}
+)
+
+// validateCert checks cert meets the requirements defined in
+// https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#certificate-requirements.
+func verifyCert(cert *x509.Certificate, extKeyUsage x509.ExtKeyUsage) error {
```
shizhMSFT, line 64: The name `verifyCert` is misleading. Will `verifyCertExtKeyUsage` be better?

signature/jws/spec.go
```diff
@@ -4,46 +4,97 @@
...
+	}
+	for _, c := range certChain[1:] {
+		for _, ext := range c.ExtKeyUsage {
+			if ext == x509.ExtKeyUsageTimeStamping {
+				if err := verifyCert(c, x509.ExtKeyUsageTimeStamping); err != nil {
+					return fmt.Errorf("timestamping certificate does not meet the minimum requirements: %w", err)
+				}
+			}
+		}
+	}
```
shizhMSFT, line 54: It seems that we are having a timestamp CA in the cert chain. I'm a bit confused.

spec/v1/signature/jws.go
```diff
@@ -0,0 +1,56 @@
...
+	ContentType string `json:"cty"`
+}
+
+// JWSUnprotectedHeader contains the set of unprotected headers.
+type JWSUnprotectedHeader struct {
+	// RFC3161 time stamp token Base64-encoded.
+	TimeStampToken string `json:"timestamp,omitempty"`
+
+	// List of X.509 certificates, each one Base64-encoded.
+	CertChain []string `json:"x5c"`
```
shizhMSFT, line 40: Should the `CertChain` be `[][]byte`? The JSON encoder can take care of the base64 encoding.

spec/v1/signature/jws.go
```diff
@@ -0,0 +1,56 @@
+package signature
```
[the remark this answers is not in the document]
qmuntal: > It is strange to have spec/v1 versioning inside notation-go as it is already versioned in SemVer2.

Adding a folder for each spec version I'm decoupling the versioning of `notation-go` with the versioning of the spec. With this I want to allow evolving other packages within `notation-go`, i.e. `signature/jws` and `plugin/manager`, without them being constrained by the spec version. In other words, what if we want to redesign `signature/jws` in a way that is not backwards compatible? We would have to bump `notation-go` major version, but we couldn't because the major version would refer to the spec, not to the library.

> Also, what if plugin and signature are versioned differently?

The plugin spec references the signature spec in multiple places, e..g.: supported envelopes, supported key types, supported hash algorithms, certificate requirements, etc. Given this fact, for me it makes more sense to version them together, else we would have to think very carefully how we deal with users using different versions of each spec.

Yet, I'm open to other opinions!

spec/v1/signature/types.go
```diff
@@ -0,0 +1,77 @@
...
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
+// Equal reports whether d and t points to the same content.
+func (d Descriptor) Equal(t Descriptor) bool {
+	return d.MediaType == t.MediaType && d.Digest == t.Digest && d.Size == t.Size
+}
+
+// Key defines a key type and size.
+type Key string
```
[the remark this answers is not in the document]
qmuntal: It does.

notation.go
```diff
@@ -55,7 +33,7 @@ func (opts SignOptions) Validate() error {
 type Signer interface {
 	// Sign signs the artifact described by its descriptor,
 	// and returns the signature.
-	Sign(ctx context.Context, desc Descriptor, opts SignOptions) ([]byte, error)
+	Sign(ctx context.Context, desc signature.Descriptor, opts SignOptions) ([]byte, error)
```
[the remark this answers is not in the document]
qmuntal: Any particular reason? The descriptor is part of the signing spec, so IMO it should be in the spec package.

signature/jws/plugin.go
```diff
@@ -0,0 +1,205 @@
...
+	PluginConfig map[string]string
+}
+
+// Sign signs the artifact described by its descriptor, and returns the signature.
+func (s *PluginSigner) Sign(ctx context.Context, desc signature.Descriptor, opts notation.SignOptions) ([]byte, error) {
+	out, err := s.Runner.Run(ctx, plugin.CommandGetMetadata, nil)
+	if err != nil {
+		return nil, fmt.Errorf("metadata command failed: %w", err)
+	}
+	metadata := out.(*plugin.Metadata)
```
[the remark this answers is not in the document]
qmuntal: I'll fix it.

signature/jws/signer.go
qmuntal: I always mistype this word...

crypto/jwsutil/envelope.go
qmuntal: The only internal consumer was `signature/jws`, and after some simplifications `jwsutil` was no longer imported anywhere.

I've you think it make sense to keep it for external consumers, I can revert the removal.

signature/jws/spec.go
qmuntal: Sure

signature/jws/spec.go
qmuntal: The plugin spec doesn't say there can't be a timestamp CA in the certificate chain. It does says that the chain must be compliant with the [spec certificate requirements](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#certificate-requirements), which specifies the minimum requirements for a timestamp cert.

spec/v1/signature/jws.go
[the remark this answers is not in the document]
qmuntal: In fact, I took the idea from https://github.com/opencontainers/image-spec/tree/main/specs-go.

spec/v1/signature/jws.go
qmuntal: Wow, I didn't know the json package encodes []byte using `base64`. Then yes, `TimeStampToken` and `CertChain` would better be byte slices.

spec/v1/signature/jws.go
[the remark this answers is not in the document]
shizhMSFT: Here I think we pop up an interesting question. Does `notation-go` implementation support multiple specs at the same time?

It makes sense for an implementation to support multiple versions of protocols for compatibility purposes. Does it make sense to support an old signature spec if we have a new one?

In addition, spec changes are considered as breaking changes. Therefore, the major version should be bumped up. In other words, `github.com/notaryproject/notation-go` will become `github.com/notaryproject/notation-go/v2`, and adopters can import both packages at the same time for supporting both versions of specs.

crypto/jwsutil/envelope.go
shizhMSFT: Although it is internal, that package is well designed / implemented / decoupled and can be used later.

signature/jws/spec.go
shizhMSFT: Then I think it is an extra check. The cert requirement for timestamping should only apply to the timestamping signature and related certs.

Instead of having the discussion on the implementation here, we need to revisit the spec itself. @gokarnm Do you have any comments for this?

spec/v1/signature/jws.go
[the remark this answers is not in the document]
qmuntal: I see your point. Then we would be bumping `notation-go` major version when there is a new spec major version or when there is a breaking change in its Go API. So it might be that `notation-go/v4` implements, i.e., the notary spec v3.

Works for me, I'll remove the `v1` folder.

crypto/jwsutil/envelope.go
qmuntal: The I'll add it again.

notation.go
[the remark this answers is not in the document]
shizhMSFT: Then we have the `Signer` interface to be spec specific, and probably we need to put the `Signer` into the `specs/v1` folder. LoL.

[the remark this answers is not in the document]
qmuntal: `Signer` is not part of the spec, it's just an interface defined by `notation-go`, so I would put it in `/spec`. As we have removed the version subfolders from `/spec`, then we have no conflict here anymore.

signature/jws/spec.go
qmuntal: I'll remove this check until the spec is clarified.

spec/v1/signature/jws.go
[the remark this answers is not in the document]
gokarnm: @qmuntal , @shizhMSFT adding some info to help with deciding the best course of action, there are multiple versions to deal with here. 
* Notary core specification include - signature specifications, supported envelopes, supported key types and algorithms, and cert requirements. These can be extended or some constraints may be relaxed, which would be non breaking changes.
* Breaking changes to Notary core specs would require major version bump.
* Trust policy language (`trustPolicy.json`) has it's own versioning, `notation-go` can support multiple major or minor versions of trust policy format, as long as the changes are compatible with Notary core specs.
* Plugin contract has it's own versioning, major and minor versions of plugin contract introduce newer `capabilities` and corresponding newer plugin commands. Check `supported-contract-versions` in [plugin spec](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#user-content-plugin-metadata).  `notation-go` is expected to support at least multiple minor versions of plugin contract (and possibly major version if we want to), existing plugin vendors should not be forced to re-implement and distribute when new minor version of plugin contract spec is published, and then implemented by `notation-go`.

signature/jws/spec.go
gokarnm: Agree with Shiwei about the validation, and also that the spec is unclear. The certificate chain for primary signature (JWS `x5c`) needs to conform to signing certificate requirements. The timestamping cert requirements apply to cert chain in TSA countersignature (JWS `timestamp`). I think there is a gap here also for requirements for intermediate and root cert in the chain. 

@priteshbandi , the [cert requirements](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#certificate-requirements) section needs updates.

gokarnm: Opened notaryproject/notaryproject#157 to track this.

signature/jws/spec.go
```diff
@@ -4,46 +4,78 @@
...
-		Notation: notationClaim{
-			Subject: desc,
-		},
+	return notaryClaim{
 		RegisteredClaims: jwt.RegisteredClaims{
-			IssuedAt:  jwt.NewNumericDate(time.Now()),
 			ExpiresAt: expiresAt,
+			IssuedAt:  jwt.NewNumericDate(time.Now()),
 		},
+		Notary: signature.JWSNotaryClaim{Subject: desc},
```
gokarnm, line 35: `Notary` claim should be renamed to `Subject`, there is typo in [signature spec](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#jws-json-serialization), addressed by this [PR](https://github.com/notaryproject/notaryproject/pull/156/files#diff-ab047165e5c22dc4d7380a85b336bd5ec76f2873057092648e583678a75a2c58R34)

signature/jws/plugin.go
```diff
@@ -0,0 +1,185 @@
...
+		return nil, fmt.Errorf("failed to marshal signing payload: %v", err)
+	}
+
+	// Execute plugin sign command.
+	req := &plugin.GenerateSignatureRequest{
+		ContractVersion: "1",
+		KeyID:           s.KeyID,
+		KeySpec:         key.KeySpec,
+		Hash:            alg.Hash(),
+		Payload:         signing,
```
gokarnm, line 102: Is this base64 encoded a second time? Updated the spec as per your [suggestion](https://github.com/notaryproject/notaryproject/pull/155#discussion_r871991310).

signature/jws/plugin.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	certs, err := parseCertChain(resp.CertificateChain)
+	if err != nil {
+		return nil, err
+	}
+
+	// Verify the hash of the request payload against the response signature
+	// using the public key of the signing certificate.
+	signed64Url := base64.RawURLEncoding.EncodeToString(resp.Signature)
```
gokarnm, line 137: Bit confused here, seems we are converting base64 encoded `resp.Signature`, the base64 encoded raw signature generated by the plugin, to base64URL encoded string required by JWS spec? Can we add a comment here.

gokarnm (changes requested):
Added some comments, will cover the rest of changes in another pass.

signature/jws/plugin.go
```diff
@@ -0,0 +1,185 @@
+package jws
```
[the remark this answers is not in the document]
qmuntal: I'll suggest to keep `PluginSigner` in `jws` until we have another envelope type to support. At that point we will have more information about how to design an envelope-agnostic plugin signer.

signature/jws/plugin.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("metadata command failed: %w", err)
+	}
+	metadata, ok := out.(*plugin.Metadata)
+	if !ok {
+		return nil, fmt.Errorf("plugin runner returned incorrect get-plugin-metadata response type '%T'", out)
+	}
+	if err := metadata.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid plugin metadata: %w", err)
+	}
```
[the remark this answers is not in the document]
qmuntal: Sure!

signature/jws/plugin.go
```diff
@@ -0,0 +1,185 @@
...
+		return s.generateSignature(ctx, desc, opts)
+	} else if metadata.HasCapability(plugin.CapabilityEnvelopeGenerator) {
+		return s.generateSignatureEnvelope(ctx, desc, opts)
+	}
+	return nil, fmt.Errorf("plugin does not have signing capabilities")
+}
+
+func (s *PluginSigner) describeKey(ctx context.Context, config map[string]string) (*plugin.DescribeKeyResponse, error) {
+	req := &plugin.DescribeKeyRequest{
+		ContractVersion: "1",
```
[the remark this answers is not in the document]
qmuntal: Done!

signature/jws/spec.go
qmuntal: Done!

signature/jws/plugin.go
qmuntal: Done!

signature/jws/plugin.go
qmuntal: `resp.Signature` type is `[]byte`, so it is not base64-encoded. The spec says that signature in the response json should be base64-encoded, and `json.[Un]Marshal` takes care of that, but the in-memory representation uses a raw `[]byte` for convencience.

spec/v1/signature/jws.go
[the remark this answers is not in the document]
dtzar: I agree with @shizhMSFT that it would be best to avoid utilizing the `v1` folder format and instead rely on versioning via the normal GitHub SemVer release process.

If we remove the `v1` folder, SemVer on release process for notation-go, and then SemVer release inside the plugin code itself (on separate repos) - will this satisfy the points you're asking @gokarnm?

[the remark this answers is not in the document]
qmuntal: I've removed the `v1` folder. It would make sense if everything is versioned together, but if every spec has its own lifecycle, then we will have to find another strategy for `notation-go` to support multiple spec version.
I don't think GitHub releases are the solution, as `notation-go` implements several independent specs (plugins, signature, trust store, etc) and we can just say notation-go@v1 only works with this and that fixed versions of the other specs. We can't forsee which version combinations will be used in the wild, so we have to provide a flexible interface.

Anyway, this goes out of the scope for this PR.

notation.go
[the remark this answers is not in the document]
shizhMSFT: There are two protocols here. One is at the `Signer` interface level, another is at the spec level. We probably need a design doc here.

shizhMSFT (approved):
LGTM

SteveLasker:
@gokarnm , is your requested change resolved?

notation.go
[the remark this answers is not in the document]
gokarnm: IMO types which are inputs and outputs of Sign/Verify methods in top level package should also be in the same package, `Descriptor` should remain at `notation` level.

I think `spec` package is not required, Notation implements one major version of Notary V2 specifications, which can introduce non breaking changes. Breaking changes can be introduced in a new major version of specification and corresponding new major version of Notation. 
An exception is the plugin contract spec versioning, which is independent of other versioning, but we need not reflect this in package naming, consumers do not directly interact with plugin contract version.
We should also put most types under `internal` package, and selectively move then out as publicly visible types as required by Notation CLI. Are we planning to use `pkg` folder for public library types? I was looking at Go lang [project layout standard](https://github.com/golang-standards/project-layout), but not sure how widely it's adopted.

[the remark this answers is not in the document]
gokarnm: > There are two protocols here. One is at the `Signer` interface level, another is at the spec level. We probably need a design doc here.

Can you clarify @shizhMSFT? I didn't understand why two interfaces are required, are you referring to the low level `Signer` interface to be added in `notation-core-go`.

signature/jws/spec.go
gokarnm: The certificate requirements can include KU, EKU, basicConstraints etc. I'd name it VerifyCertRequirements or similar.

gokarnm: I'd defer any further changes to this method till notaryproject/notaryproject#157 is resolved.

spec/v1/signature/jws.go
gokarnm: Nit, you can add this link for [x5c](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.6) in the a comment and mention that it's base64 DER.

signature/jws/plugin.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// Check keyID is honored.
+	if s.KeyID != resp.KeyID {
+		return nil, fmt.Errorf("keyID mismatch")
+	}
+
+	// Check algorithm is supported.
+	jwsAlg := resp.SigningAlgorithm.JWS()
+	if jwsAlg == "" {
+		return nil, fmt.Errorf("signing algorithm %q not supported", resp.SigningAlgorithm)
```
gokarnm, line 130:
```suggestion
		return nil, fmt.Errorf("signing algorithm %q in generateSignature response is not supported", resp.SigningAlgorithm)
```

signature/jws/plugin.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	// Check certificate chain is not empty.
+	if len(resp.CertificateChain) == 0 {
+		return nil, errors.New("empty certificate chain")
```
gokarnm, line 135:
```suggestion
		return nil, errors.New("generateSignature response has empty certificate chain")
```

signature/jws/plugin.go
```diff
@@ -0,0 +1,195 @@
...
+		if err != nil {
+			return nil, err
+		}
+		certs[i] = cert
+	}
+	return certs, nil
+}
+
+func verifyJWT(sigAlg string, payload string, sig string, signingCert *x509.Certificate) error {
+	// Verify the hash of req.payload against resp.signature using the public key if the leaf certificate.
```
gokarnm, line 192:
```suggestion
	// Verify the hash of req.payload against resp.signature using the public key in the leaf certificate.
```

signature/jws/plugin.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// Generate payload to be signed.
+	payload := packPayload(desc, opts)
+	if err := payload.Valid(); err != nil {
+		return nil, err
+	}
+
+	// Generate signing string.
+	token := jwtToken(alg.JWS(), payload)
+	signing, err := token.SigningString()
```
gokarnm, line 99: nit, rename `signing` to `payloadToSign` makes it easy to understand, spec uses the same terminology.

signature/jws/plugin.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	// Verify the hash of the request payload against the response signature
+	// using the public key of the signing certificate.
+	// At this point, resp.Signature is not base64-encoded,
+	// but verifyJWT expects a base64URL encoded string.
+	signed64Url := base64.RawURLEncoding.EncodeToString(resp.Signature)
+	err = verifyJWT(jwsAlg, signing, signed64Url, certs[0])
+	if err != nil {
+		return nil, fmt.Errorf("verification error: %v", err)
```
gokarnm, line 150:
```suggestion
		return nil, fmt.Errorf("signature returned by generateSignature cannot be verified: %v", err)
```

signature/jws/plugin.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	// Check the the certificate chain conforms to the spec.
+	if err := verifyCertExtKeyUsage(certs[0], x509.ExtKeyUsageCodeSigning); err != nil {
+		return nil, fmt.Errorf("signing certificate does not meet the minimum requirements: %w", err)
```
gokarnm, line 155:
```suggestion
		return nil, fmt.Errorf("signing certificate in generateSignature response.CertificateChain does not meet the minimum requirements: %w", err)
```

signature/jws/plugin.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	// Assemble the JWS signature envelope.
+	return jwtEnvelope(ctx, opts, signing+"."+signed64Url, resp.CertificateChain)
```
gokarnm, line 159: Method name should be `jwsEnvelope` , `converttoJws` or similar.

gokarnm: We should including a timestamp signature similar to [jws signer](https://github.com/notaryproject/notation-go/blob/main/signature/jws/signer.go#L126-L132).

spec/signature/types.go
```diff
@@ -0,0 +1,108 @@
...
+	case EC_384:
+		return ECDSA_SHA_384
+	case EC_512:
+		return ECDSA_SHA_512
+	}
+	return ""
+}
+
+// Hash algorithm associated with the key spec.
+type Hash string
```
gokarnm, line 64: I'd suggest renaming to HashAlgorithm (similar to SignatureAlgorithm), it's confusing when reading references and fields with name hash, which seems to be the hash value rather than hashing algorithm.

spec/signature/types.go
```diff
@@ -0,0 +1,108 @@
...
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
+// Equal reports whether d and t points to the same content.
+func (d Descriptor) Equal(t Descriptor) bool {
+	return d.MediaType == t.MediaType && d.Digest == t.Digest && d.Size == t.Size
+}
+
+// KeyType defines a key type and size.
+type KeyType string
```
gokarnm, line 31: I'd rename to KeySpec to be inline with plugin spec, otherwise we have code like `KeySpec signature.KeyType `json:"keySpec"` where it gets confusing. Key type by itself usually means RSA, EC etc.

spec/v1/signature/jws.go
qmuntal: Done!

spec/plugin/plugin.go
```diff
@@ -0,0 +1,122 @@
...
+	KeyID            string                       `json:"keyId"`
+	Signature        []byte                       `json:"signature"`
+	SigningAlgorithm signature.SignatureAlgorithm `json:"signingAlgorithm"`
+
+	// Ordered list of certificates starting with leaf certificate
+	// and ending with root certificate.
+	CertificateChain [][]byte `json:"certificateChain"`
+}
+
+// GenerateEnvelopeRequest contains the parameters passed in a generate-envelop request.
```
gokarnm, line 91: typo generate-envelop

spec/plugin/plugin.go
```diff
@@ -0,0 +1,122 @@
...
+type GenerateEnvelopeRequest struct {
+	ContractVersion       string            `json:"contractVersion"`
+	KeyID                 string            `json:"keyId"`
+	PayloadType           string            `json:"payloadType"`
+	SignatureEnvelopeType string            `json:"signatureEnvelopeType"`
+	Payload               []byte            `json:"payload"`
+	PluginConfig          map[string]string `json:"pluginConfig,omitempty"`
+}
+
+// GenerateSignatureResponse is the response of a generate-envelop request.
```
gokarnm, line 101: typo generate-envelop

gokarnm (changes requested):
Mostly superficial comments around naming and error messages.

For organizing packages and types, IMO we should not have a `spec` top level package. I'd keep public methods and corresponding input/output types (e.g. Descriptor) in the same package. Any private types should be moved to `internal` package, any public plugin types should move to `plugin` package.

notation.go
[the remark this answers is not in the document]
qmuntal: > I think spec package is not required, Notation implements one major version of Notary V2 specifications, which can introduce non breaking changes. Breaking changes can be introduced in a new major version of specification and corresponding new major version of Notation.

I will not tire of repeating that mixing the `notation-go` module versioning with the Notary versioning is an artificial restriction that will limit `notation-go` capacity to improve its API.  What if in a couple of months, or even years, we find that part of the API is too easy to misuse or has security vulnerabilities? What if we think there is a much better way to implement the Notary spec? We won't be able to react properly if we can't bump the major version of `nottion-go`.

About having the `spec` package or putting everything at the root of the module, I think there is value on having a clear separation for types which are defined in the Notary spec and types that are invented by `notation-go`. I'm not a maintainer of this library, so if you don't see it then I can remove it. 

> An exception is the plugin contract spec versioning, which is independent of other versioning, but we need not reflect this in package naming, consumers do not directly interact with plugin contract version.
We should also put most types under internal package, and selectively move then out as publicly visible types as appropriate for Notation CLI and other consumers

`spec/plugin` package is intended for plugin authors consumption, if we move those types into `internal` they will have to reimplement the whole plugin spec. I can assure you that all types and methods in `spec` are relevant for consumers, as I'm implementing the `notation-azure-kv` plugin and `notation` cli in parallel and I'm just exporting what's necessary.

> Are we planning to use pkg folder for public library types? I was looking at Go lang [project layout standard](https://github.com/golang-standards/project-layout), but not sure how widely it's adopted.

Having a package named `pkg` is not idiomatic, it used to be long time ago, and the current best practice is to place packages at the module root, or stack them with care.

spec/signature/types.go
qmuntal: Makes sense.

signature/jws/plugin.go
qmuntal: `jwtEnvelope` already includes the timestamp signature if the appropriate signing option is set. Is there anything missing?

qmuntal:
Thanks everyone for the feedback. I've finally removed the `spec` package, as there is no consensus on how to manage the versioning of the spec itself and this discussion is out of the scope of this PR.

signature/jws/plugin.go
gokarnm: You are right, the renamed `jwsEnvelope()` will timestamp both in the plugin and non plugin flow.

notation.go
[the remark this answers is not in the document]
gokarnm: > I will not tire of repeating that mixing the `notation-go` module versioning with the Notary versioning is an artificial restriction that will limit `notation-go` capacity to improve its API. What if in a couple of months, or even years, we find that part of the API is too easy to misuse or has security vulnerabilities?

I agree with the concern, I don't think we have thought enough about versioning of the multiple things (plugins, signature, trust policy) and which ones may reflect in package structure. For example, trust policy document has its own version property, so that policy language in `trustPolicy.json` can be independently revised without impacting anything else. We'll be making a round of refactoring to get it in line with intended separation between notation, notation-go and notation-core-go, we'll also wait for the rest of the specs to be completed. I'll track this as one of the items in a refactoring issue, and an independent issue to clarify versioning. Do you have few examples of projects that expose a spec and spec version to consumers as part of package structure, [/image-spec](https://github.com/opencontainers/image-spec/tree/main/specs-go) was one, but that's quite stable.

gokarnm (approved):
LGTM. Thanks @qmuntal for implementing this and improving the spec along the way!

notation.go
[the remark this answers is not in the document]
qmuntal: > Do you have few examples of projects that expose a spec and spec version to consumers as part of package structure, [/image-spec](https://github.com/opencontainers/image-spec/tree/main/specs-go) was one, but that's quite stable.

To name a few big ones:

- [azure-go-sdk](https://github.com/Azure/azure-sdk-for-go/tree/main/services/keyvault) use it all around.
- [aws-sdk-go](https://github.com/aws/aws-sdk-go) is more stable but also uses it.
- [google/go](https://cloud.google.com/go/docs/reference).

SteveLasker (approved):
LGTM
Great work @qmuntal

PR #43, opened 2022-05-12
Add more validations to trust-policy parsing

rgnote:
Changes:
1. Added Registry Scope validations using the regex from [distribution spec](https://github.com/distribution/distribution/blob/main/reference/regexp.go#L31)
2. Added DN validations using [go-ldap](https://github.com/go-ldap/ldap) package
3. Moved the policy document validation logic to sub-functions for better readability
4. Moved the verification utility functions to helpers.go

Testing:
Added more unit tests and extensively tested the logic with local dummy policy files


go.mod
```diff
@@ -6,3 +6,8 @@ require (
 	github.com/golang-jwt/jwt/v4 v4.4.1
 	github.com/opencontainers/go-digest v1.0.0
 )
+
+require (
+	github.com/go-ldap/ldap v3.0.3+incompatible // indirect
```
qmuntal, line 11: How is that `ldap` is marked as indirect when it is directly referenced in our code?

rgnote: I should have run `go mod tidy`. Also, I realized that I should use go-ldap v3 as mentioned [here](https://github.com/go-ldap/ldap#go-modules). Thanks for pointing this out. Will raise a new revision for this PR.

go.mod
```diff
@@ -3,6 +3,13 @@ module github.com/notaryproject/notation-go
 go 1.17
 
 require (
+	github.com/go-ldap/ldap/v3 v3.4.3
```
gokarnm, line 6: `go-ldap` has a dependency on `github.com/Azure/go-ntlmssp`, but the functionality we need, [DN validation](github.com/Azure/go-ntlmssp) does not. Is there a way to avoid this transitive dependency? Is the transitive dependency dropped during build as we do not use any types from `github.com/Azure/go-ntlmssp`?

verification/helpers.go
```diff
@@ -0,0 +1,74 @@
...
+	if domain == "" || repository == "" || !domainRegexp.MatchString(domain) || !repositoryRegexp.MatchString(repository) {
+		return fmt.Errorf(errorMessage, scope)
+	}
+
+	// No errors
+	return nil
+}
+
+// validateDistinguishedName validates if a DN name parsable and follows Notary V2 rules
+func validateDistinguishedName(name string) error {
```
gokarnm, line 44: IMO, this method should move into `notation-core-go`. All X509 related parsing, validation, cert chain validation logic should be pushed there.

gokarnm: Should this method and `validateRegistryScopeFormat` be part of policy.go, they are only used in the context of validating a trust policy statement.

go.mod
rgnote: AFAIK, final build will only include the specific methods and their dependencies. Although I could not confirm this with official Go docs, I found multiple online forums confirming this. Seems like https://github.com/jondot/goweight could help us see what's included in the final build. I added it to my TODO list to revisit later.

verification/helpers.go
rgnote: These helpers are for policy validation but I wanted to move them to a helper file so that policy.go is cleaner and deals with higher level functionality like policy verification. There is no side effect of having these methods in a helper class as Go package is a closure of everything under the package directory.

go.mod
qmuntal: unpopular opinion: Pulling in `github.com/go-ldap/ldap`, which is a library to talk with LDAP servers, just for validating a DN, seems like using a sledgehammer to crack a nut.

I would argue it's better just to copy `ldap.ParseDN`, and related tests, into `notation-go` and remove that dependency.

rgnote: @qmuntal If we consume the whole package in our builds (like in java), then I agree with what you said. Thanks to Go static linking, our build will only include `dn.ParseDN` and it's dependencies. I prefer to keep a dependency on `go-ldap` to ensure we get the updates and security fixes. Also, `go-ldap` is the most popular library for ldap in Go, I'm not too worried about having a dependency on it.

rgnote: Besides, I want to give them the credit rather than simply copying their code :)

qmuntal: I'm not that concerned about binary size, as Go is really good at prunning unnecessary code. I'm worried about having a sane dependency tree. Keeping it to the barr minimum is always a plus, even if it means duplicating code.

That said, this particular dpeendency is not a red line for me.

rgnote: Tracking this in https://github.com/notaryproject/notation-go/issues/45

verification/helpers.go
```diff
@@ -0,0 +1,74 @@
...
+			rDnCount[attribute.Type]++
+		}
+	}
+
+	// Verify there are no duplicate RDNs (multi-valdued RDNs are not supported)
+	for key := range rDnCount {
+		if rDnCount[key] > 1 {
+			return fmt.Errorf("distinguished name (DN) %q has duplicate RDNs, DN can only have unique RDNs", name)
+		}
+	}
```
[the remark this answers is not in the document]
rgnote: `go-ldap` does support multi-valued RDNs, hence notation-go lib can also support them. However, we should reject duplicate RDNs (`OU=Sales,DC=example,DC=net`) as there is no usecase for them in notary V2.

rgnote:
@qmuntal @gokarnm Addressed both your comments

qmuntal (approved):
LGTM!

verification/helpers.go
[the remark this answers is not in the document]
gokarnm: We want notation-go lib to not support multi-valued RDNs as they are not commonly used, and confusing for end customers.

gokarnm (approved):

NiazFK (approved):
LGTM

PR #44, opened 2022-05-17
Add support for trust store parsing

rgnote:
Logic to support reading named trust stores from a directory of PEM files. DER files will be supported in a separate PR. Added unit tests for different failure cases.


verification/store.go
```diff
@@ -0,0 +1,66 @@
...
+	fileInfo, err := os.Lstat(path)
+	if err != nil {
+		return nil, err
+	}
+	mode := fileInfo.Mode()
+	if !mode.IsDir() || mode&fs.ModeSymlink != 0 {
+		return nil, fmt.Errorf("%q is not a regular directory (symlinks are not supported)", path)
+	}
+
+	files, err := ioutil.ReadDir(path)
```
qmuntal, line 39: `ioutil` is de-facto deprecated. Better use `os.ReadDir`.

rgnote: @qmuntal done

qmuntal (approved):

verification/store.go
```diff
@@ -0,0 +1,71 @@
...
+	files, err := os.ReadDir(path)
+	if err != nil {
+		return nil, err
+	}
+
+	var trustStore X509TrustStore
+	for _, file := range files {
+		joinedPath := filepath.Join(path, file.Name())
+		if file.IsDir() || file.Type()&fs.ModeSymlink != 0 {
+			return nil, fmt.Errorf("%q is not a regular file (directories or symlinks are not supported)", joinedPath)
```
gokarnm, line 47:
```suggestion
			return nil, fmt.Errorf("%q is not a regular file (subdirectories or symlinks are not supported)", joinedPath)
```

gokarnm (changes requested):

verification/store.go
rgnote: We are saying "directories" are not allowed in the context of `joinedPath`. "subdirectories" makes sense if the error is in the context of the named trust store path.

verification/store.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+		certs, err := corex509.ReadCertificateFile(joinedPath)
+		if err != nil {
+			return nil, fmt.Errorf("Error while reading certificates from %q. Error : %q", joinedPath, err)
```
rgnote, line 51: @gokarnm I feel this code is better than what is suggested in https://blog.boot.dev/golang/wrapping-errors-in-go-how-to-handle-nested-errors/
LMK if you have any concerns with this.

PR #46, opened 2022-05-19
Implement the built-in JWS signer in terms of an in-process plugin

qmuntal:
This PR refactors the JWS signer so that the previous `Signer` is implemented in terms of `PluginSigner`, and renames `PluginSigner` to `Signer`.

The trick is the create an in-process plugin which implement the `generate-signature` workflow using the provided private key and certificates. 

With this change, we consolidate both implementations into the same code, removing duplicated logic and opening the door to implementing other in-process plugins, which are at least useful for testing purposes.

Code currently using `PluginSigner{...}` will have to switch to `NewSignerPlugin(...)`.


dtzar:
@qmuntal - can you please explain/reference which code that you've implemented or are aware of which is utilizing the `PluginSigner`?

signature/jws/signer.go
```diff
@@ -68,31 +55,99 @@ func NewSignerWithCertificateChain(method jwt.SigningMethod, key crypto.PrivateK
...
+// the provided key and certificates.
+type builtinPlugin struct {
+	keySpec notation.KeySpec
+
+	// method is the method to sign artifacts.
+	method jwt.SigningMethod
+
+	// key is the signing key used to sign artifacts.
+	// The signing key can be either remote or local.
+	key crypto.PrivateKey
```
gokarnm, line 88: Not sure why we say this can be a remote or local key, it would always be a local key on disk.

signature/jws/signer.go
```diff
@@ -68,31 +55,99 @@ func NewSignerWithCertificateChain(method jwt.SigningMethod, key crypto.PrivateK
...
+
+func (builtinPlugin) metadata() *plugin.Metadata {
+	// The only properties that are really relevant
+	// are the supported contract version and the capabilities.
+	// All other are just filled with meaningful data.
+	return &plugin.Metadata{
+		SupportedContractVersions: []string{plugin.ContractVersion},
+		Capabilities:              []plugin.Capability{plugin.CapabilitySignatureGenerator},
+		Name:                      "built-in",
+		Description:               "Built-in signer",
```
gokarnm, line 103:
```suggestion
		Name:                      "built-in",
		Description:               "Notation built-in signer",
```

signature/jws/signer.go
```diff
@@ -105,6 +160,27 @@ func jwtToken(alg string, claims jwt.Claims) *jwt.Token {
 	}
 }
 
+func keySpecFromJWS(alg string) (notation.KeySpec, error) {
+	var keySpec notation.KeySpec
+	switch alg {
+	case jwt.SigningMethodES256.Name:
+		keySpec = notation.EC_256
```
gokarnm, line 167: Would it be better to create a `keySpecFromKey(key crypto.PrivateKey)` method instead, currently we get a JWS alg from privatekey, and then keyspec from alg.

signature/jws/signer.go
```diff
@@ -68,31 +55,99 @@ func NewSignerWithCertificateChain(method jwt.SigningMethod, key crypto.PrivateK
...
+		// the hash provided in req1.Hash and a Sign method
+		// which does not hash data itself.
+		signed, err := r.method.Sign(string(req1.Payload), r.key)
+		if err != nil {
+			return nil, plugin.RequestError{
+				Code: plugin.ErrorCodeGeneric,
+				Err:  err,
+			}
+		}
+		signedDecoded, err := base64.RawURLEncoding.DecodeString(signed)
```
gokarnm, line 133: I'd add a comment about why we need to decode the signature before sending it in response.

gokarnm (changes requested):

signature/jws/signer.go
qmuntal: Done

signature/jws/signer.go
qmuntal: Done

signature/jws/signer.go
qmuntal: `NewSignerWithCertificateChain` allows to pass custom keys other than `rsa.PrivateKey` and `ecdsa.PrivateKey`. If we derive the keySpec from the key instead of the algorithm, we will be forcing callers to pass one of these two concrete types, as we the only way to implement your suggestion is to use a type switch for known types, similar to how `SigningMethodFromKey` is implemented.

qmuntal: Hmm, if we don't allow custom signers, which is the value of `NewSignerWithCertificateChain`? AFAIU `NewSigner` covers all use cases.

Anyway, I'll implement your suggestion.

signature/jws/algorithm.go
```diff
@@ -47,3 +48,37 @@ func SigningMethodFromKey(key interface{}) (jwt.SigningMethod, error) {
...
+	case *rsa.PublicKey:
+		switch key.Size() {
+		case 256:
+			return notation.RSA_2048, nil
+		case 384:
+			return notation.RSA_3072, nil
+		case 512:
+			return notation.RSA_4096, nil
+		default:
+			return notation.RSA_2048, nil
```
gokarnm, line 69: We should error out on unsupported key sizes.

```suggestion
		default:
			return "", fmt.Errorf("RSA key of size %q bits is not supported", key.Size()*8)
```

signature/jws/algorithm.go
```diff
@@ -47,3 +48,37 @@ func SigningMethodFromKey(key interface{}) (jwt.SigningMethod, error) {
...
+	case *ecdsa.PublicKey:
+		switch key.Curve.Params().BitSize {
+		case jwt.SigningMethodES256.CurveBits:
+			return notation.EC_256, nil
+		case jwt.SigningMethodES384.CurveBits:
+			return notation.EC_384, nil
+		case jwt.SigningMethodES512.CurveBits:
+			return notation.EC_512, nil
+		default:
+			return "", errors.New("ecdsa key not recognized")
```
gokarnm, line 80:
```suggestion
			return "", return fmt.Errorf("EC key %q of size %q bits is not supported", key.Curve.Params().Name, key.Curve.BitSize)
```

signature/jws/algorithm.go
```diff
@@ -47,3 +48,37 @@ func SigningMethodFromKey(key interface{}) (jwt.SigningMethod, error) {
...
+		}
+	}
+	return "", errors.New("key not recognized")
```
gokarnm, line 83:
```suggestion
	return "", errors.New("unsupported key type, only RSA and EC keys are supported")
```

gokarnm (changes requested):
The same changes should apply to [SignMethodFromKey](https://github.com/notaryproject/notation-go/blob/6f3abd71b7a029bb28ac1f80cb4638e73ab3a4fe/signature/jws/algorithm.go#L15), and we should remove [EdDSA](https://github.com/notaryproject/notation-go/blob/6f3abd71b7a029bb28ac1f80cb4638e73ab3a4fe/signature/jws/algorithm.go#L45-L47) as it's not a NotaryV2 supported alg.

signature/jws/signer.go
gokarnm: I think you are right `NewSignerWithCertificateChain` need not be a public method.

gokarnm:
@shizhMSFT this PR needs one more review, could you review?

qmuntal:
> The same changes should apply to [SignMethodFromKey](https://github.com/notaryproject/notation-go/blob/6f3abd71b7a029bb28ac1f80cb4638e73ab3a4fe/signature/jws/algorithm.go#L15), and we should remove [EdDSA](https://github.com/notaryproject/notation-go/blob/6f3abd71b7a029bb28ac1f80cb4638e73ab3a4fe/signature/jws/algorithm.go#L45-L47) as it's not a NotaryV2 supported alg.

Done!

shizhMSFT (commented):
I have a general comment on the design.

This PR proposes that the build-in JWS signer is modelled as a special plugin. That is,

```
         /- Build-in Plugin
*Signer <-- Plugin 1
         \- Plugin 2
```

Since we are already treating `*Signer` as a generic interface, which implements `notation.Signer`, why not unexport `Signer` and `PluginSigner` and return `notation.Signer` for `NewSignerPlugin` and `NewSigner`?

For example,

```golang
func NewSignerPlugin(runner plugin.Runner, keyID string, pluginConfig map[string]string) (notation.Signer, error) {
```

```
                 /- *jws.signer
notation.Signer <                      /- Plugin 1
                 \- *jws.pluginSigner <
                                       \- Plugin 2
```

signature/jws/algorithm.go
```diff
@@ -23,27 +29,25 @@ func SigningMethodFromKey(key interface{}) (jwt.SigningMethod, error) {
 	case *rsa.PublicKey:
 		switch key.Size() {
```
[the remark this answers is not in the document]
qmuntal: Done!

signature/jws/algorithm.go
```diff
@@ -23,27 +29,25 @@ func SigningMethodFromKey(key interface{}) (jwt.SigningMethod, error) {
...
-		case jwt.SigningMethodES384.CurveBits:
-			return jwt.SigningMethodES384, nil
-		case jwt.SigningMethodES512.CurveBits:
-			return jwt.SigningMethodES512, nil
+		case 256:
+			return notation.EC_256, jwt.SigningMethodES256, nil
+		case 384:
+			return notation.EC_384, jwt.SigningMethodES384, nil
+		case 521:
+			return notation.EC_512, jwt.SigningMethodES512, nil
```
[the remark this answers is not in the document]
qmuntal: Good point. When defining `EC_512` is just copied the values from https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#algorithm-selection. Is that table wrong?

signature/jws/signer.go
qmuntal: If we unexport `NewSignerWithCertificateChain` then exporting `SigningMethodFromKey` neither makes sense, as it only makes sense in conjunction with `NewSignerWithCertificateChain`.

Anyway, these changes go out of the scope of this PR.

qmuntal:
> Since we are already treating *Signer as a generic interface, which implements notation.Signer, why not unexport Signer and PluginSigner and return notation.Signer for NewSignerPlugin and NewSigner?

Agree and done 👍

signature/jws/signer_test.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 func TestSignerInterface(t *testing.T) {
-	if _, ok := interface{}(&Signer{}).(notation.Signer); !ok {
+	if _, ok := interface{}(&pluginSigner{}).(notation.Signer); !ok {
 		t.Error("&Signer{} does not conform notation.Signer")
 	}
 }
```
shizhMSFT, line 23: Since `NewXXX` returns `notation.Signer`, this kind of test cases are redundant.

signature/jws/algorithm.go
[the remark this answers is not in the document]
shizhMSFT: That table is wrong :(
/cc @priteshbandi

Reference: [FIPS 186-4 6.1.1](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf)

shizhMSFT (approved):

shizhMSFT:
LGTM

signature/jws/signer_test.go
qmuntal: Removed.

qmuntal:
@shizhMSFT I've pushed two changes after your last review:
- Calculate EC key size using *n* bit length instead of the length of the underlying size, as mandated by [FIPS 186-4 6.1.1](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf).
- Add signing test for all supported key specs.

signature/jws/signer.go
qmuntal: At the end I've removed `NewSignerWithCertificateChain` and unexported `SigningMethodFromKey`.

signature/jws/signer_test.go
```diff
@@ -50,6 +46,48 @@ func TestSignWithCertChain(t *testing.T) {
...
+func TestSignWithCertChain(t *testing.T) {
+	// sign with key
+	tests := []struct {
+		name string
+		fn   func() (crypto.PrivateKey, error)
+	}{
+		{
+			name: string(notation.RSA_2048),
+			fn:   func() (crypto.PrivateKey, error) { return rsa.GenerateKey(rand.Reader, 2048) },
+		},
```
gokarnm, line 58: We should add some negative tests for RSA and EC with Notary unsupported key lengths where applicable, and a negative test for non RSA or EC key. Can be addressed as a separate PR.

gokarnm (approved):

NiazFK (approved):
LGTM

PR #55, opened 2022-06-13
Add SignatureVerificationLevel type

rgnote:
Adding types for VerificationType, VerificationAction, SignatureVerificationLevel as spec'd in https://github.com/notaryproject/notaryproject/blob/main/trust-store-trust-policy-specification.md#signature-verification-details

### DIFF
--- a/verification/signature_verification_levels.go
+++ b/verification/signature_verification_levels.go
@@ -1,0 +1,105 @@
+package verification
+
+import "fmt"
+
+// VerificationType is an enum for signature verification types such as Integrity, Authenticity, etc.
+type VerificationType string
+
+// VerificationAction is an enum for signature verification actions such as Enforced, Logged, Skipped.
+type VerificationAction string
+
+// SignatureVerificationLevel encapsulates the signature verification preset and it's actions for each verification type
+type SignatureVerificationLevel struct {
+	Name            string
+	VerificationMap map[VerificationType]VerificationAction
+}
+
+const (
+	Integrity             VerificationType = "Integrity"
+	Authenticity          VerificationType = "Authenticity"
+	TrustedTimestampCheck VerificationType = "TrustedTimestampCheck"
+	ExpiryCheck           VerificationType = "ExpiryCheck"
+	RevocationCheck       VerificationType = "RevocationCheck"
+
+	Enforced VerificationAction = "Enforced"
+	Logged   VerificationAction = "Logged"
+	Skipped  VerificationAction = "Skipped"
+)
+
+var (
+	Strict = SignatureVerificationLevel{
+		"strict",
+		map[VerificationType]VerificationAction{
+			Integrity:             Enforced,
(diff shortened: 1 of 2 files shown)


verification/signature_verification_levels.go
```diff
@@ -0,0 +1,105 @@
...
+		Audit,
+		Skip,
+	}
+)
+
+// FindSignatureVerificationLevel finds if the given string corresponds to a supported SignatureVerificationLevel, otherwise throws an error
+func FindSignatureVerificationLevel(s string) (*SignatureVerificationLevel, error) {
+
+	for _, level := range SignatureVerificationLevels {
+		if level.Name == s {
```
priteshbandi, line 100: Should we use `strings.EqualFold` for case insensitive comparison?

rgnote: IMO, we should not allow inconsistent SignatureVerificationLevel names. Let's stick to lower case only.

priteshbandi (approved):
LGTM

verification/signature_verification_levels.go
```diff
@@ -0,0 +1,105 @@
...
+	Name            string
+	VerificationMap map[VerificationType]VerificationAction
+}
+
+const (
+	Integrity             VerificationType = "Integrity"
+	Authenticity          VerificationType = "Authenticity"
+	TrustedTimestampCheck VerificationType = "TrustedTimestampCheck"
+	ExpiryCheck           VerificationType = "ExpiryCheck"
+	RevocationCheck       VerificationType = "RevocationCheck"
```
gokarnm, line 22:
```suggestion
	TrustedTimestamp VerificationType = "TrustedTimestamp"
	Unexpired           VerificationType = "Unexpired"
	RevocationCheck       VerificationType = "RevocationCheck"
```

verification/signature_verification_levels.go
```diff
@@ -0,0 +1,105 @@
...
+
+	Enforced VerificationAction = "Enforced"
+	Logged   VerificationAction = "Logged"
+	Skipped  VerificationAction = "Skipped"
+)
+
+var (
+	Strict = SignatureVerificationLevel{
```
shizhMSFT, line 30: Should we have pointers instead of instances

```suggestion
	Strict = &SignatureVerificationLevel{
```

verification/signature_verification_levels.go
rgnote: Discussed offline and changing to 

```			Integrity,
			Authenticity,
			AuthenticTimestamp,
			Expiry,
			Revocation
```

verification/signature_verification_levels.go
rgnote: Don't see why we can't do that

shizhMSFT (approved):
LGTM. It would be better with more documentation

gokarnm (approved):
LGTM!

PR #56, opened 2022-06-14
Add CodeQL Security Scanning

Wwwsylvia:
Signed-off-by: Sylvia Lei <lixia_lei@outlook.com>

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,36 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: main
+  pull_request:
+    branches: main
+  schedule:
+    - cron: '29 2 * * 5'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


gokarnm (approved):
LGTM!

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,36 @@
...
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+    strategy:
+      matrix:
+        go-version: [1.17, 1.18]
```
shizhMSFT, line 21: @gokarnm There's a general question on the go version. The notation-go library is quite different from other libraries as it is a security library. We should encourage developers to compile the library with the latest go since the latest version contains the latest security patches.

shizhMSFT (approved):
LGTM

.github/workflows/codeql.yml
gokarnm: Agreed, though I don't have a good sense of what that translates to in practice for this library. What are your suggestions on what we should support currently, and say when 1.19 is released. Seems Go will release security fixes to two most recent minor releases.

https://go.dev/security/policy
https://github.com/golang/go/wiki/MinorReleases

PR #58, opened 2022-06-15
Fix trust store tests issue on Windows

rgnote:

### DIFF
--- a/verification/store.go
+++ b/verification/store.go
@@ -64,6 +64,10 @@ func LoadX509TrustStore(path string) (*X509TrustStore, error) {
 		trustStore.Certificates = append(trustStore.Certificates, certs...)
 	}
 
+	if len(trustStore.Certificates) < 1 {
+		return nil, fmt.Errorf("trust store %q has no x509 certificates", path)
+	}
+
 	trustStore.Name = filepath.Base(path)
 	trustStore.Path = path
 
--- a/verification/store_test.go
+++ b/verification/store_test.go
@@ -19,7 +19,7 @@ func TestLoadValidTrustStore(t *testing.T) {
 }
 
 func TestLoadSymlinkTrustStore(t *testing.T) {
-	// TODO run symlink tests on Windows
+	// TODO run symlink tests on Windows. See https://github.com/notaryproject/notation-go/issues/59
 	if runtime.GOOS == "windows" {
 		t.Skip("skipping the symlink test on Windows")
 	}
--- a/verification/store_test.go
+++ b/verification/store_test.go
@@ -1,12 +1,15 @@
 package verification
 
 import (
+	"fmt"
+	"path/filepath"
+	"runtime"
 	"testing"
 )
 
 // TestLoadTrustStore tests a valid trust store
 func TestLoadValidTrustStore(t *testing.T) {
-	trustStore, err := LoadX509TrustStore("testdata/trust-store/valid-trust-store")
+	trustStore, err := LoadX509TrustStore(filepath.FromSlash("testdata/trust-store/valid-trust-store"))
 	if err != nil {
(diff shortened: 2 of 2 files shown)


verification/store_test.go
```diff
@@ -16,36 +19,61 @@ func TestLoadValidTrustStore(t *testing.T) {
 }
 
 func TestLoadSymlinkTrustStore(t *testing.T) {
-	_, err := LoadX509TrustStore("testdata/trust-store/valid-trust-store_SYMLINK")
-	if err == nil || err.Error() != "\"testdata/trust-store/valid-trust-store_SYMLINK\" is not a regular directory (symlinks are not supported)" {
+	// TODO run symlink tests on Windows
+	if runtime.GOOS == "windows" {
```
rgnote, line 23: Skipping symlink tests on Windows as it requires Admin permissions to create symlinks. [source](https://github.com/golang/go/issues/22874)

verification/store_test.go
```diff
@@ -16,36 +19,61 @@ func TestLoadValidTrustStore(t *testing.T) {
...
+		t.Skip("skipping the symlink test on Windows")
+	}
+	path := filepath.FromSlash("testdata/trust-store/valid-trust-store_SYMLINK")
+	_, err := LoadX509TrustStore(path)
+	fmt.Println(err)
```
Wwwsylvia, line 28: Do we need to print out the error here?

verification/store_test.go
```diff
@@ -16,36 +19,61 @@ func TestLoadValidTrustStore(t *testing.T) {
...
+	if err == nil || err.Error() != fmt.Sprintf("%q is not a regular directory (symlinks are not supported)", path) {
 		t.Fatalf("symlinks should return error : %q", err)
```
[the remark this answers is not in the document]
rgnote: This is expected. Since the if condition is checking whether the err is nil or the error message is not matching, it is good to know to which is the case by printing the err.

verification/store_test.go
rgnote: Will remove it

verification/store.go
```diff
@@ -64,6 +64,10 @@ func LoadX509TrustStore(path string) (*X509TrustStore, error) {
 		trustStore.Certificates = append(trustStore.Certificates, certs...)
 	}
 
+	if len(trustStore.Certificates) < 1 {
+		return nil, fmt.Errorf("trust store %q has no x509 certificates", path)
+	}
+
```
gokarnm, line 70: I'm debating if this function should throw an error or the caller should account for this condition. The caller has more context such as value of `trustpolicy.truststore` value, and in the case where we extend policy to support specifying multiple trust stores, an empty store is a valid condition.

rgnote: > in the case where we extend policy to support specifying multiple trust stores, an empty store is a valid condition.

Why is it a valid case?

rgnote: Discussed offline. Keeping it as is for now. Will remove the validation when needed.

gokarnm (approved):

verification/store_test.go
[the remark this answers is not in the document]
Wwwsylvia: We can print both the expected error and the returned error. Like what we do at https://github.com/notaryproject/notation-go/blob/main/plugin/metadata_test.go#L25.

priteshbandi (approved):
LGTM

verification/store_test.go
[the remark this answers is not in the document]
rgnote: Wouldn't that still print `%!q(<nil>)` as the returned err is still a nil? The link you sent is printing booleans, not errors.

gokarnm:
@shizhMSFT merging this PR with two approvals.

PR #60, opened 2022-06-17
Run unit tests in Github workflow

Wwwsylvia:
Signed-off-by: Lixia (Sylvia) Lei <lixia_lei@outlook.com>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,16 +9,17 @@ on:
 jobs:
   build:
     name: Continuous Testing
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [1.17]
+        go-version: [1.17, 1.18]
       fail-fast: true
     steps:
       - name: Set up Go ${{ matrix.go-version }}
         uses: actions/setup-go@v3
         with:
           go-version: ${{ matrix.go-version }}
+          check-latest: true
       - name: Check out code
         uses: actions/checkout@v3
       - name: Cache Go modules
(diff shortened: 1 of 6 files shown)


gokarnm (commented):
Can we change `test.yml` to `build.yml` to be consistent with other repos.

.github/workflows/test.yml
```diff
@@ -9,16 +9,17 @@ on:
 jobs:
   build:
     name: Continuous Testing
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
rgnote, line 12: Same comment as in the other PR, can we build/test in Mac and Windows?

dtzar: @rgnote - How about this be a separate 'feature' request?  It's not common IMO to build a golang library on multiple platforms - theoretically it should just work because it's golang.

.github/workflows/test.yml
```diff
@@ -9,16 +9,17 @@ on:
...
     strategy:
       matrix:
-        go-version: [1.17]
+        go-version: [1.17, 1.18]
```
dtzar, line 15: We should standardize on 1.18 IMO.  Notation is on this version.  notation-core-go should be upgraded to the same.

dtzar (approved):

.github/workflows/test.yml
gokarnm: We can track this as separate request. We do need to test on multiple platforms because there are differences in implementation (E.g. [directory structure spec](https://github.com/notaryproject/notation/blob/main/specs/directory.md)) that need to be tested in each platform.

.github/workflows/test.yml
Wwwsylvia: `notation-go` uses go 1.17, and this pipeline test for both 1.17 and 1.18.

.github/workflows/test.yml
Wwwsylvia: Make sense. But we may want to do this in a seperate PR, as currently we use `make test` to generate coverage report, which will not work on Windows.

shizhMSFT (approved):
LGTM

PR #67, opened 2022-06-24
[WIP] feat: add new registry interface and renam Link

binbin-li:
### What?

1. Add a new interface to support get signature manifests given the digest. We need to retrieve both digests and annotations to get them verified.
2. Rename `Link` to `CreateSignatureManifest` to better reflect its functionality.


registry/repository.go
```diff
@@ -24,6 +24,11 @@ type RepositoryClient struct {
 	remote.Repository
 }
 
+type SignatureManifest struct {
+	Blobs       []artifactspec.Descriptor
```
binbin-li, line 28: Looks like there is only one descriptor in the blobs, do we want to keep descriptor instead of a slice?

registry/interface.go
```diff
@@ -9,6 +9,9 @@ import (
 
 // SignatureRepository provides a storage for signatures
 type SignatureRepository interface {
+	// GetCoreSignatureManifest returns signature manifests for verification filterd by given subject digest
+	GetCoreSignatureManifest(ctx context.Context, subjectDigest digest.Digest) ([]CoreSignatureManifest, error)
+
 	// Lookup finds all signatures for the specified manifest
 	Lookup(ctx context.Context, manifestDigest digest.Digest) ([]digest.Digest, error)
```
shizhMSFT, line 16: How about `List` to replace `GetCoreSignatureManifest` and `Lookup`, listing all signatures for the specified manifest? Will that sound better?

```go
import artifactspec "github.com/oras-project/artifacts-spec/specs-go/v1"
```

```go
List(ctx context.Context, manifestDigest digest.Digest) ([]SignatureManifest, error)
```

registry/repository.go
shizhMSFT: One blob is better than a slice.

registry/interface.go
```diff
@@ -18,6 +21,6 @@ type SignatureRepository interface {
 	// Put uploads the signature to the registry
 	Put(ctx context.Context, signature []byte) (notation.Descriptor, error)
 
-	// Link creates an signature artifact linking the manifest and the signature
-	Link(ctx context.Context, manifest, signature notation.Descriptor) (notation.Descriptor, error)
+	// CreateSignatureManifest creates an signature artifact linking the manifest and the signature
+	CreateSignatureManifest(ctx context.Context, manifest, signature notation.Descriptor) (notation.Descriptor, error)
```
shizhMSFT, line 25: How about combining `Put` and `Link` into one?
```go
Put(ctx context.Context, signature []byte, annotations map[string]string) (notation.Descriptor, SignatureManifest, error)
```

where the returned descriptor is the descriptor of the signature manifest?

binbin-li: Since `CreateSignatureManifest` is also doing something similar to `put`, it's reasonable to me to have them combined.

registry/interface.go
binbin-li: The renaming makes sense to me. My only concern is that the returned `SignatureManifest` which only includes `blob` and `annotations` fileds. It's like a partial `SignatureManifest`. What if we need an api that returns a complete `SignatureManifest` in the future? Do we want to fill all fields in this struct right now or give it a different name?

registry/repository.go
binbin-li: thanks! I'll update it.

shizhMSFT:
We also need an interface to resolve manifest descriptors from tag or digest.

We may need

```go
type Repository interface{
    SignatureRepository

    Resolve(ctx context.Context, reference string) (notation.Descriptor, error)
}
```

registry/interface.go
binbin-li: Done

registry/interface.go
binbin-li: Synced offline. Created `SignatureManifest` struct including a `blob` and `annotations`. And refactored `List` to return a slice of `SignatureManifest`.

registry/interface.go
```diff
@@ -9,15 +9,20 @@ import (
 
 // SignatureRepository provides a storage for signatures
 type SignatureRepository interface {
-	// Lookup finds all signatures for the specified manifest
-	Lookup(ctx context.Context, manifestDigest digest.Digest) ([]digest.Digest, error)
+	// List returns all signature manifests given the manifest digest
+	List(ctx context.Context, manifestDigest digest.Digest) ([]SignatureManifest, error)
```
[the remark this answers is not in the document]
binbin-li: Good call. I'll rename it to `ListSignatureManifests`.

registry/repository.go
```diff
@@ -35,18 +40,18 @@ func NewRepositoryClient(client remote.Client, ref registry.Reference, plainHTTP
 	}
 }
 
-// GetManifestDescriptor returns signature manifest information by tag or digest.
-func (c *RepositoryClient) GetManifestDescriptor(ctx context.Context, ref string) (notation.Descriptor, error) {
-	desc, err := c.Repository.Resolve(ctx, ref)
+// Resolve returns signature manifest information by tag or digest.
```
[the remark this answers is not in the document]
binbin-li: I'll get the comment updated.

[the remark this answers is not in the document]
binbin-li: Done!

registry/interface.go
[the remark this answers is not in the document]
binbin-li: Done!

rgnote (approved):
LGTM

registry/interface.go
```diff
@@ -9,15 +9,20 @@ import (
 
 // SignatureRepository provides a storage for signatures
 type SignatureRepository interface {
-	// Lookup finds all signatures for the specified manifest
-	Lookup(ctx context.Context, manifestDigest digest.Digest) ([]digest.Digest, error)
+	// ListSignatureManifests returns all signature manifests given the manifest digest
+	ListSignatureManifests(ctx context.Context, manifestDigest digest.Digest) ([]SignatureManifest, error)
 
 	// Get downloads the signature by the specified digest
 	Get(ctx context.Context, signatureDigest digest.Digest) ([]byte, error)
```
rgnote, line 16: nit: I would reword the comment and the method arguments since this can be used to retrieve any content, not just the signature blob.

registry/repository.go
```diff
@@ -24,6 +24,11 @@ type RepositoryClient struct {
 	remote.Repository
 }
 
+type SignatureManifest struct {
+	Blob        artifactspec.Descriptor
```
rgnote, line 28: Can we keep the blob be `notation.Descriptor` rather than `artifactSpec.Descriptor`?

rgnote (approved):

shizhMSFT (approved):
LGTM

PR #72, opened 2022-06-30
Verification helpers

rgnote:
Changes:

1. Changed `trustPolicy.TrustStore` to `trustPolicy.TrustStores`
2. Added `Type` field to X509TrustStore to be able to filter trust stores based on their type
3. Added verification workflow errors and helper methods to load policies and trust stores
4. Added `VerificationResult` and `SignatureVerificationOutsome` types to encapsulate verification results.
5. Associated unit tests for above changes


verification/helpers.go
```diff
@@ -32,6 +70,19 @@ func getArtifactPathFromUri(artifactUri string) (string, error) {
 	return artifactPath, nil
 }
 
+func getArtifactDigestFromUri(artifactUri string) (string, error) {
+	i := strings.LastIndex(artifactUri, ":")
+	if i < 0 {
+		return "", fmt.Errorf("artifact URI %q could not be parsed, make sure it is the fully qualified OCI artifact URI without the scheme/protocol. e.g domain.com:80/my/repository:digest", artifactUri)
+	}
+
+	artifactDigest := artifactUri[i+1:]
```
binbin-li, line 79: Is it possible that `i+1` is out of range causing panic

verification/types.go
```diff
@@ -1,13 +1,34 @@
...
 type VerificationType string
 
 // VerificationAction is an enum for signature verification actions such as Enforced, Logged, Skipped.
 type VerificationAction string
 
+// VerificationResult encapsulates the verification result (passed or failed) for a verification type, including the
+// desired verification action as specified in the trust policy
+type VerificationResult struct {
+	FailedToVerify bool
+	Success        bool
```
binbin-li, line 18: just out of curiosity, what's the difference between `Success` and `FailedToVerify`

verification/helpers.go
```diff
@@ -1,13 +1,51 @@
...
+	var result = make(map[string]*X509TrustStore)
+	for _, trustPolicy := range policyDocument.TrustPolicies {
+		for _, trustStore := range trustPolicy.TrustStores {
+			if result[trustStore] != nil {
+				// we loaded this trust store already
+				continue
+			}
+			i := strings.Index(trustStore, ":")
+			prefix := trustStore[:i]
+			name := trustStore[i+1:]
```
shizhMSFT, line 38: try to use [strings.Cut](https://pkg.go.dev/strings#Cut) as `i` might be `-1`.

verification/helpers.go
```diff
@@ -32,6 +70,19 @@ func getArtifactPathFromUri(artifactUri string) (string, error) {
...
+	if i < 0 {
+		return "", fmt.Errorf("artifact URI %q could not be parsed, make sure it is the fully qualified OCI artifact URI without the scheme/protocol. e.g domain.com:80/my/repository:digest", artifactUri)
+	}
+
+	artifactDigest := artifactUri[i+1:]
+	if artifactDigest == "" {
+		return "", fmt.Errorf("artifact URI %q has an invalid digest %q, make sure the URI is the fully qualified OCI artifact URI without the scheme/protocol. e.g domain.com:80/my/repository:digest", artifactUri, artifactDigest)
+	}
+	return artifactDigest, nil
+}
```
shizhMSFT, line 84: It seems this function is not used anywhere. Similarly, other functions like `loadX509TrustStores` are not referenced anywhere. If those function are just for tests, they should be `_test.go` files instead of regular go files.

verification/helpers.go
```diff
@@ -32,6 +70,19 @@ func getArtifactPathFromUri(artifactUri string) (string, error) {
 	return artifactPath, nil
 }
 
+func getArtifactDigestFromUri(artifactUri string) (string, error) {
+	i := strings.LastIndex(artifactUri, ":")
+	if i < 0 {
+		return "", fmt.Errorf("artifact URI %q could not be parsed, make sure it is the fully qualified OCI artifact URI without the scheme/protocol. e.g domain.com:80/my/repository:digest", artifactUri)
```
shizhMSFT, line 76: `digest` is digest like `sha256:xxx` or a tag like `v1`?

verification/verifier_helpers.go
```diff
@@ -0,0 +1,53 @@
...
+	if isPresent(wildcard, trustPolicy.TrustedIdentities) {
+		return nil
+	}
+
+	var trustedX509Identities []map[string]string
+	for _, identity := range trustPolicy.TrustedIdentities {
+		i := strings.Index(identity, ":")
+
+		identityPrefix := identity[:i]
+		identityValue := identity[i+1:]
```
shizhMSFT, line 25: `strings.Cut()` might be used instead.

notation.go
```diff
@@ -33,6 +34,11 @@ func (d Descriptor) Equal(t Descriptor) bool {
 	return d.MediaType == t.MediaType && d.Digest == t.Digest && d.Size == t.Size
 }
 
+// Payload describes the content that gets signed.
+type Payload struct {
+	TargetPayload artifactspec.Descriptor `json:"targetArtifact"`
```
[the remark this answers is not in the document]
rgnote: Good catch. Should be `TargetArtifact`. Will update

verification/policy.go
```diff
@@ -32,8 +32,8 @@ type TrustPolicy struct {
 	RegistryScopes []string `json:"registryScopes"`
 	// SignatureVerification setting for this policy statement
 	SignatureVerification string `json:"signatureVerification"`
-	// TrustStore this policy statement uses
-	TrustStore string `json:"trustStore,omitempty"`
+	// TrustStores this policy statement uses
+	TrustStores []string `json:"trustStores,omitempty"`
```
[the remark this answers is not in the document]
rgnote: We discussed this change in yesterday's call. Made the code update as it is minor. If the spec deviates, I can come back and make those changes.

verification/helpers.go
rgnote: It can't be `-1` because we are validating it at https://github.com/notaryproject/notation-go/blob/main/verification/policy.go#L129

verification/helpers.go
rgnote: These functions are not for testing. `Verifier` logic will use these. https://github.com/notaryproject/notation-go/blob/main/verification/verifier.go#L21

verification/helpers.go
rgnote: Spec is not supporting tags yet. We can come back and update this code if we need to support tags.

verification/verifier_helpers.go
rgnote: This surely makes the code look pretty. Will address in my next PR

rgnote: Looks like strings.Cut() was added in 1.18. notation-go is still on 1.17
We can address it when we upgrade to 1.18

verification/errors.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (e ErrorNoApplicableTrustPolicy) Error() string {
+	if e.msg != "" {
+		return e.msg
+	}
+	return "there is no applicable trust policy for the given artifact"
+}
+
+// ErrorSignatureRetrievalFailed is used when notation is unable to retrieve the digital signature/s for the given artifact
+type ErrorSignatureRetrievalFailed struct {
```
gokarnm, line 28: Should we differentiate this condition from when there are no signatures associated with an artifact (unsigned)? That condition also seems different than the one where we evaluated some signatures and failed.

verification/helpers.go
```diff
@@ -1,13 +1,51 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return policyDocument, nil
+}
+
+func loadX509TrustStores(policyDocument *PolicyDocument, trustStoreBasePath string) (map[string]*X509TrustStore, error) {
+	var result = make(map[string]*X509TrustStore)
+	for _, trustPolicy := range policyDocument.TrustPolicies {
+		for _, trustStore := range trustPolicy.TrustStores {
```
gokarnm, line 31: It seems this method loads all the trust stores references by any of the trust policies, why do we need to load all of then in the context of a single signature verification.

verification/policy.go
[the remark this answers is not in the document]
gokarnm: @shizhMSFT I'll be submitting a spec PR for this, we discussed the change and use case in Thursday call . Rakesh has been working on verification workflow logic and it was simpler to update it in a single pass.

verification/types.go
```diff
@@ -1,13 +1,42 @@
...
 type VerificationType string
 
 // VerificationAction is an enum for signature verification actions such as Enforced, Logged, Skipped.
 type VerificationAction string
 
+// VerificationResult encapsulates the verification result (passed or failed) for a verification type, including the
+// desired verification action as specified in the trust policy
+type VerificationResult struct {
+	// FailedToVerify is set to true if the verification is inconclusive due to a run time error
+	FailedToVerify bool
```
gokarnm, line 18: Can we capture this condition as an error, this seems like an internal error condition in notation itself. You can set success to false, and capture details in error.

verification/verifier_helpers.go
```diff
@@ -0,0 +1,53 @@
...
+	"strings"
+)
+
+// isFailureResult checks whether a VerificationResult fails the entire signature verification workflow.
+// signature verification workflow is considered failed if there is a VerificationResult with "Enforced" as the action but the result was inconclusive (failed to verify) or unsuccessful
+func isFailureResult(result VerificationResult) bool {
+	return result.Action == Enforced && (result.FailedToVerify || !result.Success)
+}
+
+func verifyX509TrustedIdentities(certs []*x509.Certificate, trustPolicy *TrustPolicy) error {
```
gokarnm, line 15: The trusted identity (anchor) is checked against the signing certificate, can we pass only the signing cert instead of the chain.

verification/verifier_helpers.go
```diff
@@ -0,0 +1,53 @@
...
+	if isPresent(wildcard, trustPolicy.TrustedIdentities) {
+		return nil
+	}
+
+	var trustedX509Identities []map[string]string
```
gokarnm, line 20: IMO it's better to have a strongly types DN struct which contains the RDN map, than using nested map/array of more than two level.

gokarnm (changes requested):

verification/errors.go
```diff
@@ -0,0 +1,49 @@
+package verification
+
+// ErrorVerificationSkipped is used when signature verification was not performed due to "SignatureVerification" being set to "skip" in the trust policy
+type ErrorVerificationSkipped struct {
+	msg string
+}
```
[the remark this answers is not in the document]
rgnote: Correct, this does not perform any verification. This error is only to convey callers that the verification was skipped. The alternative would be populating `SignatureVerificationOutcome` object for each of the verification types and return. That did not seam elegant. If there is a better alternative, I can consider.

verification/errors.go
rgnote: See `ErrorSignatureRetrievalFailed` for that purpose

verification/helpers.go
rgnote: This is to compute hashes on the trust store certificates so that we can early early during verification

From https://github.com/notaryproject/notaryproject/blob/main/signing-and-verification-workflow.md#verification-steps
> Calculate the SHA-256 fingerprint of all the trusted certificates and compare them against the list of SHA-256 certificate fingerprints present in io.cncf.notary.x509certs.fingerprint.sha256 annotation of artifact manifest.

verification/types.go
rgnote: If we do that, we won't be able to differentiate a bad signature from a good signature with a runtime error. Wouldn't the error message that is shown to the CLI users depend on this differentiation?

verification/verifier_helpers.go
rgnote: We are already using the leaf for this purpose - https://github.com/notaryproject/notation-go/blob/main/verification/verifier.go#L71

Since we are using pointers, there is no impact on performance either ways. I can still do it if required.

verification/types.go
binbin-li: Gotcha, thanks for the explanation!

verification/verifier_helpers.go
gokarnm: Agreed, the suggestion is mainly to keep the method signature idiomatic, passing the whole chain allows a future change to incorrectly access other elements of the cert chain, whereas we want to restrict to only the leaf.

verification/policy.go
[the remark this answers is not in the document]
gokarnm: Related spec update [PR](https://github.com/notaryproject/notaryproject/pull/164).

notation.go
```diff
@@ -33,6 +34,11 @@ func (d Descriptor) Equal(t Descriptor) bool {
...
+}
+
```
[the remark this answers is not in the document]
shizhMSFT: We should use the descriptor type defined on line 18. The original design for `notation.Descriptor` is that the descriptor in the payload should not depend on other specs.

verification/types.go
rgnote: Discussed offline. Removed `FailedToVerify`

notation.go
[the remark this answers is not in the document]
rgnote: Changed it to `notation.Descriptor`

gokarnm (approved):
LGTM!

binbin-li:
lgtm

verification/types.go
```diff
@@ -1,13 +1,40 @@
...
 // VerificationType is an enum for signature verification types such as Integrity, Authenticity, etc.
 type VerificationType string
 
 // VerificationAction is an enum for signature verification actions such as Enforced, Logged, Skipped.
 type VerificationAction string
 
+// VerificationResult encapsulates the verification result (passed or failed) for a verification type, including the
+// desired verification action as specified in the trust policy
+type VerificationResult struct {
+	// 	Success is set to true if the verification was successful
```
shizhMSFT, line 17:
```suggestion
	// Success is set to true if the verification was successful
```

verification/helpers.go
shizhMSFT: `domain.com:80/my/repository:digest` this reference string makes me confused. If it is for digest, it should be in the form of `domain.com:80/my/repository@digest`. That is, `@` instead of `:`.

verification/helpers_test.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+func TestGetArtifactDigestFromUri(t *testing.T) {
+
+	tests := []struct {
+		artifactUri string
+		digest      string
+		wantErr     bool
+	}{
+		{"domain.com:80/repository:digest", "digest", false},
```
shizhMSFT, line 18: We should set `wantErr: true` for this case.

verification/policy.go
```diff
@@ -125,9 +125,11 @@ func validateTrustedIdentities(statement TrustPolicy) error {
...
 	supportedTrustStorePrefixes := []string{"ca"}
 
-	i := strings.Index(statement.TrustStore, ":")
-	if i < 0 || !isPresent(statement.TrustStore[:i], supportedTrustStorePrefixes) {
-		return fmt.Errorf("trust policy statement %q uses an unsupported trust store type %q in trust store value %q", statement.Name, statement.TrustStore[:i], statement.TrustStore)
+	for _, trustStore := range statement.TrustStores {
+		i := strings.Index(trustStore, ":")
+		if i < 0 || !isPresent(trustStore[:i], supportedTrustStorePrefixes) {
+			return fmt.Errorf("trust policy statement %q uses an unsupported trust store type %q in trust store value %q", statement.Name, trustStore[:i], trustStore)
+		}
```
shizhMSFT, line 132: It seems this block of code does not support case like `ca:acme-rockets,tsa:acme-tsa`.

Reference: https://github.com/notaryproject/notaryproject/pull/164/files#r914176243
/cc @gokarnm

verification/verifier_helpers.go
shizhMSFT: Please create an issue for the future update. Meanwhile, the index `i` needs to be checked to see whether `i >= 0`.

verification/policy.go
rgnote: Yes, I'm aware of this. Not parsing TSA trust stores yet as TSA verification is out of scope for RC1. Will create an issue for tracking

rgnote: https://github.com/notaryproject/notation-go/issues/74

verification/verifier_helpers.go
rgnote: We are already verifying that trusted identities are in the right format [here](https://github.com/notaryproject/notation-go/blob/main/verification/policy.go#L97)
Upgrade issue : https://github.com/notaryproject/notation-go/issues/75

verification/helpers_test.go
rgnote: Updated the code to accept `domain.com:80/repository@<alg>:<digest>`

My understanding is that <alg> is not included when querying manifests from the registry, is that correct? Or the entire thing `<alg>:<digest>` is considered a digest and we need to use that when [resolving](https://github.com/notaryproject/notation-go/blob/main/registry/interface.go#L27)?

rgnote: NVM. Alg is needed to query with a digest. I addressed it in the latest revision

verification/helpers.go
```diff
@@ -32,6 +68,21 @@ func getArtifactPathFromUri(artifactUri string) (string, error) {
...
+		return "", invalidUriErr
+	}
+
+	j := strings.LastIndex(artifactUri[i+1:], ":")
+	if j < 0 || j+1 == len(artifactUri[i+1:]) {
+		return "", invalidUriErr
+	}
+
+	return artifactUri[i+1:], nil
+}
```
shizhMSFT, line 84: This block of code is workable but fragile. Try [registry.ParseReference()](https://pkg.go.dev/oras.land/oras-go/v2@v2.0.0-rc.1/registry#ParseReference).

shizhMSFT (approved):
LGTM

PR #87, opened 2022-07-22
feat: Use PackArtifact to push signature manifest

binbin-li:
### What?
This PR refactors to use the new method `PackArtifact` to pack and push the signature manifest as mentioned in https://github.com/oras-project/oras-go/issues/175#issuecomment-1168158460

### DIFF
--- a/registry/repository.go
+++ b/registry/repository.go
@@ -92,20 +92,20 @@ func (c *RepositoryClient) Get(ctx context.Context, digest digest.Digest) ([]byt
 }
 
 // PutSignatureManifest creates and uploads an signature artifact linking the manifest and the signature
-func (c *RepositoryClient) PutSignatureManifest(ctx context.Context, signature []byte, manifest notation.Descriptor, annotaions map[string]string) (notation.Descriptor, SignatureManifest, error) {
+func (c *RepositoryClient) PutSignatureManifest(ctx context.Context, signature []byte, manifest notation.Descriptor, annotations map[string]string) (notation.Descriptor, SignatureManifest, error) {
 	signatureDesc, err := c.uploadSignature(ctx, signature)
 	if err != nil {
 		return notation.Descriptor{}, SignatureManifest{}, err
 	}
 
-	manifestDesc, err := c.uploadSignatureManifest(ctx, artifactDescriptorFromNotation(manifest), signatureDesc, annotaions)
+	manifestDesc, err := c.uploadSignatureManifest(ctx, artifactDescriptorFromNotation(manifest), signatureDesc, annotations)
 	if err != nil {
 		return notation.Descriptor{}, SignatureManifest{}, err
 	}
 
 	signatureManifest := SignatureManifest{
 		Blob:        notationDescriptorFromArtifact(signatureDesc),
-		Annotations: annotaions,
+		Annotations: annotations,
 	}
 	return notationDescriptorFromOCI(manifestDesc), signatureManifest, nil
 }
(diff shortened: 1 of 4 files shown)


registry/repository.go
```diff
@@ -55,7 +56,7 @@ func (c *RepositoryClient) ListSignatureManifests(ctx context.Context, manifestD
 	// TODO(shizhMSFT): filter artifact type at the server side
 	if err := c.Repository.Referrers(ctx, ocispec.Descriptor{
 		Digest: manifestDigest,
-	}, func(referrers []artifactspec.Descriptor) error {
+	}, "", func(referrers []artifactspec.Descriptor) error {
```
shizhMSFT, line 59: Instead of empty artifact type, you may specify the artifact type when invoking `Referrers()`.

binbin-li: Gotcha, updated to `ArtifactTypeNotation`

registry/repository.go
```diff
@@ -55,7 +56,7 @@ func (c *RepositoryClient) ListSignatureManifests(ctx context.Context, manifestD
 	// TODO(shizhMSFT): filter artifact type at the server side
 	if err := c.Repository.Referrers(ctx, ocispec.Descriptor{
 		Digest: manifestDigest,
-	}, func(referrers []artifactspec.Descriptor) error {
+	}, ArtifactTypeNotation, func(referrers []artifactspec.Descriptor) error {
```
chloeyin, line 59: Just a question. Since we have already provided `ArtifactTypeNotation` to the refer api. do we still need to manually filter it again in line 61?

chloeyin: https://github.com/oras-project/oras-go/blob/97a9c43c52f9d89ecf5475bc59bd1f96c8cc61f6/registry/remote/repository.go#L455-L481
I think api will filter it if we pass the artifact type.

binbin-li: Good call, it can be removed safely now.

chloeyin:
LGTM

shizhMSFT (approved):
LGTM

registry/repository.go
gokarnm: I think client side filtering should still be retained as `artifactType` filtering is an optional server feature as per ORAS [referrers spec](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md#filtering-results). I'm not sure though that the response contains any indication from the server that `artifactType` filtering was not applied.

@shizhMSFT any feedback?

shizhMSFT: @gokarnm The client-side filtering is ensured by the `oras-go` SDK.

registry/repository.go
```diff
@@ -98,25 +98,16 @@ func (c *RepositoryClient) PutSignatureManifest(ctx context.Context, signature [
...
 	}
 
-	// generate artifact manifest
-	artifactManifest := artifactspec.Manifest{
-		MediaType:    artifactspec.MediaTypeArtifactManifest,
-		ArtifactType: ArtifactTypeNotation,
-		Blobs:        []artifactspec.Descriptor{signatureDesc},
-		Subject:      artifactDescriptorFromNotation(manifest),
-		Annotations:  annotaions,
+	manifestDesc, err := c.uploadSignatureManifest(ctx, artifactDescriptorFromNotation(manifest), signatureDesc, annotaions)
```
gokarnm, line 101: nit, `annotaions`.

binbin-li: oops, thanks for catching!

gokarnm: Can we rename `manifest` to `subjectManifest` to make it clearer, we are dealing with signature blob, subject and signature manifest descriptors here. Same for line 153.

gokarnm (approved):
LGTM! Would be great if we can address this [renaming suggestion](https://github.com/notaryproject/notation-go/pull/87#discussion_r930690524) as well.

registry/repository.go
binbin-li: that makes sense, already updated.

shizhMSFT (approved):
LGTM

PR #103, opened 2022-08-10
refactor: refactor signer to support cose signature

chloeyin:

# What
Refactor notation-go to support cose signature

# How to review

1. plugin/plugin.go: Add some fields to the plugin contract struct to better support new interface. Need to update when new spec is ready.
2. signature/algorithm.go: Add `tostring` method for keyspec and hash function. Need to update when new spec is ready. If we don't need hash and keyspec in signing request, it can be deprecated.
3. signature/provider.go: A combination of original signature provider and runner. It implements runner and signer interface. For built-in provider, it use `LocalSigner` from `notation-core-go`. For external provider, it implements `signer` interface by using runner to interact with external plugin (sending`describle key` and `generate signature` request)
5. signature/plugin.go: replace runner with provider
6. signature/signer.go: construct a built-in provider with certs and key.
7. ignore any test file changes

### DIFF
--- a/signature/provider.go
+++ b/signature/provider.go
@@ -1,0 +1,186 @@
+package signature
+
+import (
+	"context"
+	"crypto"
+	"crypto/x509"
+	"fmt"
+
+	"github.com/notaryproject/notation-core-go/signature"
+	"github.com/notaryproject/notation-go/plugin"
+)
+
+// provider wraps a runner and a signature.Signer
+type provider interface {
+	plugin.Runner
+	signature.Signer
+	SetConfig(map[string]string)
+}
+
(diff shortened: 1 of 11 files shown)


signature/provider.go
```diff
@@ -0,0 +1,177 @@
...
+	"context"
+	"crypto"
+	"crypto/x509"
+	"fmt"
+
+	"github.com/notaryproject/notation-core-go/signature"
+	"github.com/notaryproject/notation-go/plugin"
+)
+
+type provider interface {
```
binbin-li, line 13: open question, not sure if `signer` or `signerProvider` better fits here.

binbin-li (approved):
lgtm overall

PR #104, opened 2022-08-10
fix: dir package optimize

JeyJeyGao:
1. GetPath logic: if the given path doesn't exist, return the first
   possible path in the union directories, so when create a new file the
   UnionDirFS will select user directory which was provided as the first
   directory when create the UnionDirFS
2. change the constant name

### DIFF
--- a/dir/fs.go
+++ b/dir/fs.go
@@ -76,12 +76,12 @@ func (u unionDirFS) Open(name string) (fs.File, error) {
 // GetPath returns the path of the named file or directory under the FS
 //
 // if path exists, it returns the first existing path in union directories (dirs)
-// if path doesn't exist, it returns fs.ErrNotExist error
+//
+// if path doesn't exist, it returns the first possible path in the union directories
+// for creating new file and a fs.ErrNotExist error.
 func (u unionDirFS) GetPath(elem ...string) (string, error) {
-	var targetPath string
 	pathSuffix := path.Join(elem...)
 	for _, dir := range u.Dirs {
-		targetPath = filepath.Join(dir.Root(), pathSuffix)
 		_, err := fs.Stat(dir, pathSuffix)
 		if err != nil {
 			if errors.Is(err, fs.ErrNotExist) {
@@ -92,10 +92,19 @@ func (u unionDirFS) GetPath(elem ...string) (string, error) {
 		}
 		// got the first existing file path and break
 		// return the path with current OS separator
-		return targetPath, nil
+		return filepath.Join(dir.Root(), pathSuffix), nil
 	}
-	// return the last possible path for creating new file
-	return targetPath, &fs.PathError{
+	// if the given path does not exist, return the first possible path in the
+	// union directories for creating new file.
+	var fallbackPath string
+	if len(u.Dirs) == 0 {
+		return "", &fs.PathError{
+			Op:  "getpath",
+			Err: errors.New("the union directory is empty"),
+		}
+	}
+	fallbackPath = filepath.Join(u.Dirs[0].Root(), pathSuffix)
+	return fallbackPath, &fs.PathError{
 		Op:   "getpath",
 		Err:  fs.ErrNotExist,
 		Path: pathSuffix,
--- a/dir/path.go
+++ b/dir/path.go
@@ -2,21 +2,20 @@ package dir
 
 import (
 	"errors"
-	"fmt"
 	"io/fs"
 
 	"github.com/opencontainers/go-digest"
 )
 
 const (
-	// CertificateExtension defines the extension of the certificate files
-	CertificateExtension = ".crt"
-
 	// ConfigFile is the name of config file
 	ConfigFile = "config.json"
 
-	// KeyExtension defines the extension of the key files
-	KeyExtension = ".key"
+	// LocalCertificateExtension defines the extension of the certificate files
+	LocalCertificateExtension = ".crt"
+
+	// LocalKeyExtension defines the extension of the key files
+	LocalKeyExtension = ".key"
 
 	// LocalKeysDir is the directory name for local key store
 	LocalKeysDir = "localkeys"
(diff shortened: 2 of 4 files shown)


dir/fs.go
```diff
@@ -92,10 +90,16 @@ func (u unionDirFS) GetPath(elem ...string) (string, error) {
 		}
 		// got the first existing file path and break
 		// return the path with current OS separator
-		return targetPath, nil
+		return filepath.Join(dir.Root(), pathSuffix), nil
 	}
-	// return the last possible path for creating new file
-	return targetPath, &fs.PathError{
+	// if the given path does not exist, return the first possible path in the
+	// union directories for creating new file.
```
shizhMSFT, line 96: It still returns an error when falling back. Is that expected?

JeyJeyGao: Yes, it is expected. It returns fs.ErrNotExist, so the user know the situation and decide whether report an error or create a new file.

shizhMSFT (approved):
LGTM

dir/path.go
```diff
@@ -65,7 +65,7 @@ func (p *PathManager) Config() string {
 //
 // extension: support .crt|.key
 func (p *PathManager) Localkey(name string, extension string) string {
```
rgnote, line 67: I don't think any caller would be interested in just retrieving cert or key path. Can we return both paths instead of one at a time?

dir/fs.go
```diff
@@ -78,10 +78,8 @@ func (u unionDirFS) Open(name string) (fs.File, error) {
 // if path exists, it returns the first existing path in union directories (dirs)
 // if path doesn't exist, it returns fs.ErrNotExist error
```
rgnote, line 79: Update the comment

dir/fs.go
```diff
@@ -92,10 +90,16 @@ func (u unionDirFS) GetPath(elem ...string) (string, error) {
...
 		// return the path with current OS separator
-		return targetPath, nil
+		return filepath.Join(dir.Root(), pathSuffix), nil
 	}
-	// return the last possible path for creating new file
-	return targetPath, &fs.PathError{
+	// if the given path does not exist, return the first possible path in the
+	// union directories for creating new file.
+	var fallbackPath string
+	if len(u.Dirs) > 0 {
```
rgnote, line 98: Can we throw an error if `u.Dirs` is empty?

dir/path.go
```diff
@@ -9,14 +9,14 @@ import (
 )
 
 const (
-	// CertificateExtension defines the extension of the certificate files
-	CertificateExtension = ".crt"
-
 	// ConfigFile is the name of config file
 	ConfigFile = "config.json"
 
+	// CertificateExtension defines the extension of the certificate files
```
rgnote, line 15: Update the comment

dir/path.go
```diff
@@ -9,14 +9,14 @@ import (
...
+	LocalCertificateExtension = ".crt"
+
 	// KeyExtension defines the extension of the key files
```
rgnote, line 18: Update the comment

dir/path.go
JeyJeyGao: Good idea! Updated.

dir/fs.go
JeyJeyGao: Updated.

dir/fs.go
JeyJeyGao: Yes, added an error for empty `u.Dirs`.

dir/path.go
```diff
@@ -60,17 +59,14 @@ func (p *PathManager) Config() string {
 	return path
 }
 
-// LocalKey returns path of the local private keys or certificate
+// LocalKey returns path of the local private keys and certificate
```
rgnote, line 62:
```suggestion
// LocalKey returns path of the local private key and it's certificate
```

dir/path_test.go
```diff
@@ -137,24 +137,13 @@ func TestPathManager_LocalKey(t *testing.T) {
...
 	}
-}
-
-func TestPathManager_LocalKeyFailed(t *testing.T) {
-	path := &PathManager{
-		UserConfigFS: NewUnionDirFS(
-			NewRootedFS("/home/exampleuser/.config/notation/", nil),
-		),
+	if certPath != "/home/exampleuser/.config/notation/localkeys/key1"+LocalCertificateExtension {
+		t.Fatal("get Localkey() failed.")
```
rgnote, line 145: Why the returned paths is not having extensions included?

JeyJeyGao: if certPath != "/home/exampleuser/.config/notation/localkeys/key1"+`LocalCertificateExtension` {
extension is here.

dir/path.go
JeyJeyGao: Updated.

dir/path_test.go
rgnote: In path.go, we are retrieving the key path with extension included, right?
```
	keyPath, err := p.UserConfigFS.GetPath(LocalKeysDir, name+LocalKeyExtension)
```

If that is the case, then the returned path should be `/home/exampleuser/.config/notation/localkeys/key1.key` (rather than `/home/exampleuser/.config/notation/localkeys/key1`). Isn't it?

JeyJeyGao: Yes, it is. So, what is your question? The unit test checks the path with extension.

> 	keyPath, certPath := path.Localkey("key1")
	if keyPath != "/home/exampleuser/.config/notation/localkeys/key1"+LocalKeyExtension {
		t.Fatal("get Localkey() failed.")
	}
	if certPath != "/home/exampleuser/.config/notation/localkeys/key1"+LocalCertificateExtension {
		t.Fatal("get Localkey() failed.")
	}

dtzar:
@rgnote - it seems your feedback has been addressed.  Can you please take a look again and approve if all is good?

rgnote (approved):

shizhMSFT (approved):
LGTM

PR #111, opened 2022-08-12
Create version.go

vaninrao10:
Package Version shows the current notation-go version.


shizhMSFT:
Should the version of `notation-go` library be marked as the git tag and recorded in `go.mod`?

shizhMSFT (changes requested):
The `version` package should be only applied to CLI.

priteshbandi:
We should need version only for cli(notation), for libraries(notation-go and notaiton-core-go) we can rely on git tags.

internal/Version/version.go
```diff
@@ -0,0 +1,17 @@
+package version
+
+var (
+	// Version shows the current notation-go version, optionally with pre-release.
+	Version = "v1.0.0-alpha.3"
```
dtzar, line 5: I believe @yizha1 is targeting v0.10.0-alpha.3 per a different PR.  We need to align versions also with the roadmap release notes PR @iamsamirzon

iamsamirzon: @dtzar  - Agree on aligning on the right versions to use. I had intentionally recommended updating the major revision to go from 0 to 1 because the new Notation client and even the specs have been significantly changes from the last release.

I suggest we use the version numbering I had recommend in https://github.com/notaryproject/roadmap/pull/64

iamsamirzon:
> Should the version of `notation-go` library be marked as the git tag and recorded in `go.mod`?

Either way we need to update the release process in here to document our decision. https://github.com/notaryproject/notation/blob/main/RELEASE_CHECKLIST.md

dtzar:
Per the conversation on the community call today, we don't want to have the version number in the go library code because the GitHub release is all which is needed/useful to determine the version.

PR #113, opened 2022-08-17
refactor:refactor notation plugin runner to support cose

chloeyin:
1. Move key and cert to the member fields of external provider.
2. Add `GuessSignatureEnvelopeFormat` to insepct signature format in a brute force manner.
3. Add validate to signer to ensure the envelope format is supported.
4. Bug fix and doc.

### DIFF
--- a/signature/plugin.go
+++ b/signature/plugin.go
@@ -16,7 +16,7 @@ import (
 // pluginSigner signs artifacts and generates signatures.
 type pluginSigner struct {
 	sigProvider       provider
-	mediaTypeEnvelope string
+	envelopeMediaType string
 	keyID             string
 	pluginConfig      map[string]string
 }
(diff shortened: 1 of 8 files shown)


signature/plugin.go
```diff
@@ -25,22 +25,38 @@ type pluginSigner struct {
...
-		sigProvider:       newExternalProvider(runner, keyID),
-		mediaTypeEnvelope: mediaTypeEnvelope,
+	signer := &pluginSigner{
+		sigProvider:       extProvider,
+		envelopeMediaType: envelopeMediaType,
 		keyID:             keyID,
 		pluginConfig:      pluginConfig,
-	}, nil
+	}
+	if err := signer.validate(); err != nil {
```
binbin-li, line 46: I feel we can validate the mediaType before creating the signer.

PR #119, opened 2022-08-23
test:fix plugin tests

chloeyin:
1. Rewrite plugin test

### DIFF
--- a/signature/plugin_test.go
+++ b/signature/plugin_test.go
@@ -2,79 +2,280 @@ package signature
 
 import (
 	"context"
-	"crypto/rand"
-	"crypto/rsa"
 	"crypto/x509"
-	"encoding/base64"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"reflect"
 	"strings"
 	"testing"
 	"time"
 
-	"github.com/golang-jwt/jwt/v4"
-	"github.com/notaryproject/notation-core-go/signer"
+	"github.com/notaryproject/notation-core-go/signature"
+	"github.com/notaryproject/notation-core-go/signature/cose"
+	"github.com/notaryproject/notation-core-go/signature/jws"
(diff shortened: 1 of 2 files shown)


signature/plugin_test.go
```diff
@@ -2,425 +2,836 @@ package signature
...
+	validMetaDataWithSigningCapabilityFunc = func(ctx context.Context, req plugin.Request) (interface{}, error) {
+		return &validMetadata, nil
+	}
+	validMetaDataWithEnvelopeGeneratorCapabilityFunc = func(ctx context.Context, req plugin.Request) (interface{}, error) {
+		metaData := validMetadata
+		metaData.Capabilities = []plugin.Capability{plugin.CapabilityEnvelopeGenerator}
+		return &metaData, nil
+	}
+	// TODO: plugin spec may change
+	// validDescribeKeyWFunc = func(ctx context.Context, req plugin.Request) (interface{}, error) {
```
binbin-li, line 51: do we need to remove it?

signature/plugin_test.go
```diff
@@ -2,425 +2,836 @@ package signature
...
+	signFunc
+	certChainFunc
+	keySpecFunc
+	runner *mockRunner
 }
 
-func (r *mockRunner) Run(_ context.Context, _ plugin.Request) (interface{}, error) {
-	defer func() { r.n++ }()
-	return r.resp[r.n], r.err[r.n]
+type option func(*options)
```
binbin-li, line 63: nit: I feel `optionFunc` might make more sense

signature/plugin_test.go
chloeyin: > do we need to remove it?

Removed

signature/plugin_test.go
chloeyin: Updated

PR #121, opened 2022-08-26
refactor:support cose and new plugin spec

chloeyin:
1. Support new plugin spec(mainly new algorithm and hash name)
2. Sign with payload instead of hash
3. Update sign request and signer interface
4. A better mock plugin provider for test
5. Add more unit test

```
go test ./signature -cover -count=1
ok      github.com/notaryproject/notation-go/signature  11.247s coverage: 88.5% of statements
```

### DIFF
--- a/signature/provider.go
+++ b/signature/provider.go
@@ -19,16 +19,17 @@ var builtInPluginMetaData = plugin.Metadata{
 	URL:                       "https://github.com/notaryproject/notation-go",
 }
 
-// provider wraps a plugin.Runner and a signature.Signer
+// provider wraps a plugin.Runner and a signature.Signer.
 type provider interface {
 	plugin.Runner
 	signature.Signer
 	SetConfig(map[string]string)
 }
 
-// builtinPlugin is a builtin provider implementation
-// It only supports describe key and metadata command
-// It wraps signature.Signature to support builtin signing method
+// builtinPlugin is a builtin provider implementation.
+//
+// It only supports describe key and metadata command.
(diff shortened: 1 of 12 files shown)


signature/provider.go
```diff
@@ -126,69 +117,58 @@ func (p *externalProvider) describeKey(ctx context.Context) (*plugin.DescribeKey
...
-
-// CertificateChain returns cert chain of a keyID
-func (p *externalProvider) CertificateChain() ([]*x509.Certificate, error) {
-	return p.certs, nil
-}
-
-// KeySpec return s keyspec of a keyID
-func (p *externalProvider) KeySpec() (signature.KeySpec, error) {
-	return p.keySpec, nil
+	keySpec := ParseKeySpecFromName(keyResp.KeySpec)
```
binbin-li, line 171: seems like `ParseKeySpecFromName` will return empty keySpec when `keyResp.KeySpec` is unsupported. Do you think we should return err instead of empty keySpec in this case?

chloeyin: How about still returning an empty keySpec. If `ParseKeySpecFromName ` returns an empty keySpec, `KeySpec` will call `describleKey`, and we can change keySpec to a struct type instead of a pointer.
```
if p.keySpec != signature.KeySpec{} {
      return p.keySpec, nil
}
keyResp, err := p.describleKey(context.Background())
// ...
p.keySpec = ParseKeySpecFromName(keyResp.KeySpec)
```

PR #134, opened 2022-09-09
feat: dir package allows user to define user/system write level

JeyJeyGao:
1. config add `EnvelopeType` argument
2. dir package optimize: allow user define `WriteLevel` = UserLevel | SystemLevel
3. add ConfigForWrite(), TrustStoreForWrite(), TrustPolicyForWrite() func to get writable path
4. refector `config` pacakge: receive a `path` argument when load and save file to allow user to overwrite the default directory structure

### DIFF
--- a/dir/path.go
+++ b/dir/path.go
@@ -36,13 +36,28 @@ const (
 	TrustStoreDir = "truststore"
 )
 
+// WriteLevel defines the write directory level supporting UserLevel or SystemLevel.
+type WriteLevel int
+
+const (
+	// SystemLevel is the label to specify write directory to system level
+	SystemLevel WriteLevel = 0
+	// UserLevel is the label to specify write directory to user level
+	UserLevel WriteLevel = 1
+)
+
 // PathManager contains the union directory file system and methods
 // to access paths of notation
 type PathManager struct {
-	ConfigFS     UnionDirFS
-	CacheFS      UnionDirFS
-	LibexecFS    UnionDirFS
-	UserConfigFS UnionDirFS
+	ConfigFS  UnionDirFS
+	CacheFS   UnionDirFS
+	LibexecFS UnionDirFS
+
+	UserConfigFS   UnionDirFS
+	SystemConfigFS UnionDirFS
+
+	UserLibexecFS   UnionDirFS
+	SystemLibexecFS UnionDirFS
 }
 
 func checkError(err error) {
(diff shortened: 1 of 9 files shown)


dir/path.go
```diff
@@ -120,3 +153,18 @@ func (p *PathManager) CachedSignatureStoreDirPath() string {
 	checkError(err)
 	return path
 }
+
+func getPathForWrite(user UnionDirFS, system UnionDirFS, item ...string) string {
```
patrickzheng200, line 157: since 'item' could be more than one,  'items' might be better.

dir/path.go
```diff
@@ -120,3 +153,18 @@ func (p *PathManager) CachedSignatureStoreDirPath() string {
...
+	var (
+		path string
+		err  error
+	)
+	if WriteLevel == SystemLevel {
+		path, err = system.GetPath(item...)
+	} else {
+		path, err = user.GetPath(item...)
+	}
```
patrickzheng200, line 166: When system == p.SystemConfigFS and user == p.UserConfigFS, then this logic does not cover the case of writing/updating/removing from both levels. Therefore, to work on both system and user levels (e.g. add a file to both levels), one needs to explicitly change the WriteLevel var outside this func by calling 'WriteLevel = 0', and the caller has to remember the current WriteLevel value to avoid miss operations on the wrong level.

So it looks to me that WriteLevel should be an argument of getPathForWrite. Functions calling getPathForWrite shall pass in WriteLevel, which is decided by the caller.

dir/path.go
```diff
@@ -52,14 +69,19 @@ func checkError(err error) {
...
+// Config returns the ready-only path of config.json
 func (p *PathManager) Config() string {
 	path, err := p.ConfigFS.GetPath(ConfigFile)
 	checkError(err)
 	return path
 }
 
-// LocalKey returns path of the local private key and it's certificate
+// ConfigForWrite returns the writable path of config.json
+func (p *PathManager) ConfigForWrite() string {
```
patrickzheng200, line 80: WriteLevel is better to be an argument to this func instead of a var in the package. See below.

dir/path.go
```diff
@@ -69,27 +91,38 @@ func (p *PathManager) Localkey(name string) (keyPath, certPath string) {
...
+// TrustPolicy returns the ready-only path of trustpolicy.json file
 func (p *PathManager) TrustPolicy() string {
 	path, err := p.ConfigFS.GetPath(TrustPolicyFile)
 	checkError(err)
 	return path
 }
 
-// X509TrustStore returns the path of x509 trust store certificate
+// TrustPolicyForWrite returns the writable path of trustpolicy.json file
+func (p *PathManager) TrustPolicyForWrite() string {
```
patrickzheng200, line 109: Same here. See below.

dir/path.go
```diff
@@ -69,27 +91,38 @@ func (p *PathManager) Localkey(name string) (keyPath, certPath string) {
...
+
+// X509TrustStore returns the read-only path of x509 trust store certificate
 func (p *PathManager) X509TrustStore(prefix, namedStore string) string {
 	path, err := p.ConfigFS.GetPath(TrustStoreDir, "x509", prefix, namedStore)
 	checkError(err)
 	return path
 }
 
+// X509TrustStoreForWrite returns the writable path of x509 trust store certificate
+func (p *PathManager) X509TrustStoreForWrite(prefix, namedStore string) string {
```
patrickzheng200, line 121: Same here. See below.

dir/path.go
JeyJeyGao: Done.

dir/path.go
JeyJeyGao: Thanks. Added writeLevel argument.

dir/path.go
JeyJeyGao: Done.

dir/path.go
JeyJeyGao: Done.

dir/path.go
JeyJeyGao: Done.

patrickzheng200 (approved):
LGTM

PR #137, opened 2022-09-13
fix: dir package isolates system/user directory

### LINKED ISSUE #144: directory structure needs to isolate `read` and `write` path
As current design, when system level directory has the config file while user level directory doesn't have the config file, it uses the system level config file for read and write. 

It is necessary to allow the user to know that whether the write path is user level or system level, so we need to:
1. isolate `read` /`write` path
2. provide the ability to select `user`/`system` level directory to write.
Linked to [System vs User file/directory precedence#203](https://github.com/notaryproject/notation/issues/203)

JeyJeyGao:
1. config add EnvelopeType argument.
2. dir package optimize: allow user define DirLevel = UserLevel | SystemLevel | UnionLevel.
4. refector config pacakge: receive a path argument when load and save file to allow user to overwrite the default directory structure.

Test: added unit test.

Resolves #144 

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -45,7 +45,7 @@ func NewConfig() *Config {
 func (c *Config) Save(path string) error {
 	// set default path
 	if path == "" {
-		path = dir.Path.ConfigForWrite(dir.UserLevel)
+		path = dir.Path.Config(dir.UserLevel)
 	}
 	return save(path, c)
 }
@@ -57,7 +57,7 @@ func (c *Config) Save(path string) error {
 func LoadConfig(path string) (*Config, error) {
 	// set default path
 	if path == "" {
-		path = dir.Path.Config()
+		path = dir.Path.Config(dir.UnionLevel)
 	}
 
 	// load config
--- a/config/config.go
+++ b/config/config.go
@@ -41,10 +41,6 @@ func NewConfig() *Config {
 //
 // if the `path` is an empty string, it uses built-in user level config directory.
 func (c *Config) Save(path string) error {
-	// set default path
-	if path == "" {
-		path = dir.Path.Config(dir.UserLevel)
-	}
 	return save(path, c)
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -25,6 +25,8 @@ type Config struct {
 	InsecureRegistries       []string                 `json:"insecureRegistries"`
 	CredentialsStore         string                   `json:"credsStore,omitempty"`
 	CredentialHelpers        map[string]string        `json:"credHelpers,omitempty"`
+	// EnvelopeType defines the envelope type for signing
+	EnvelopeType string `json:"envelopeType,omitempty"`
 }
 
 // VerificationCertificates is a collection of public certs used for verification.
--- a/config/config.go
+++ b/config/config.go
@@ -43,7 +43,7 @@ func NewConfig() *Config {
 
 // Save stores the config to file.
 //
-// if the `path` is an empty string, it uses build-in user level config directory.
+// if the `path` is an empty string, it uses built-in user level config directory.
 func (c *Config) Save(path string) error {
 	// set default path
 	if path == "" {
@@ -54,7 +54,7 @@ func (c *Config) Save(path string) error {
 
 // LoadConfig reads the config from file or return a default config if not found.
 //
-// if `path` is an empty string, it uses build-in config.json directory, including
+// if `path` is an empty string, it uses built-in config.json directory, including
 // user level and system level.
 func LoadConfig(path string) (*Config, error) {
 	// set default path
--- a/config/config.go
+++ b/config/config.go
@@ -43,28 +43,28 @@ func NewConfig() *Config {
 
 // Save stores the config to file.
 //
-// if the `path` is not set, it uses build-in user level config directory.
-func (c *Config) Save(path ...string) error {
-	if len(path) > 0 {
-		return save(path[0], c)
+// if the `path` is an empty string, it uses build-in user level config directory.
+func (c *Config) Save(path string) error {
+	// set default path
+	if path == "" {
+		path = dir.Path.ConfigForWrite(dir.UserLevel)
 	}
-	return save(dir.Path.ConfigForWrite(dir.UserLevel), c)
+	return save(path, c)
 }
 
 // LoadConfig reads the config from file or return a default config if not found.
 //
-// if `path` is not set, it uses build-in config.json directory, including
+// if `path` is an empty string, it uses build-in config.json directory, including
 // user level and system level.
-func LoadConfig(path ...string) (*Config, error) {
-	var (
-		err    error
-		config Config
-	)
-	if len(path) > 0 {
-		err = load(path[0], &config)
-	} else {
-		err = load(dir.Path.Config(), &config)
+func LoadConfig(path string) (*Config, error) {
+	// set default path
+	if path == "" {
+		path = dir.Path.Config()
 	}
+
+	// load config
+	var config Config
+	err := load(path, &config)
 	if err != nil {
 		if errors.Is(err, fs.ErrNotExist) {
 			return NewConfig(), nil
--- a/config/config.go
+++ b/config/config.go
@@ -3,6 +3,8 @@ package config
 import (
 	"errors"
 	"io/fs"
+
+	"github.com/notaryproject/notation-go/dir"
 )
 
 // CertificateReference is a named file path.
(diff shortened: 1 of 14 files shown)


config/config.go
```diff
@@ -37,15 +41,30 @@ func NewConfig() *Config {
 	}
 }
 
-// Save stores the config to file
-func (c *Config) Save() error {
-	return save(ConfigPath, c)
+// Save stores the config to file.
+//
+// if the `path` is an empty string, it uses build-in user level config directory.
```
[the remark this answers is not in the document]
JeyJeyGao: Thanks!

config/config.go
```diff
@@ -37,15 +41,30 @@ func NewConfig() *Config {
...
 }
 
 // LoadConfig reads the config from file or return a default config if not found.
-func LoadConfig() (*Config, error) {
+//
+// if `path` is an empty string, it uses build-in config.json directory, including
+// user level and system level.
+func LoadConfig(path string) (*Config, error) {
+	// set default path
+	if path == "" {
```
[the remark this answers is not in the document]
JeyJeyGao: If `path` is empty, it uses the default value. It is not elegant to provide a default value by using an empty `path` string, but Golang doesn't support the default function parameter. If I support more invalid `path` string, the code will be much not elegant. In order to use the default value, different user will call it by different invalid `path` sring.

patrickzheng200:
Looks like we can move the target of this PR to the dev-rc.1 branch for now?

JeyJeyGao:
>

JeyJeyGao:
> Looks like we can move the target of this PR to the dev-rc.1 branch for now?

This PR has no dependent on COSE refactor, so it can be merged to main branch.

dir/path.go
```diff
@@ -36,13 +36,28 @@ const (
...
+// WriteLevel defines the write directory level supporting UserLevel or SystemLevel.
+type WriteLevel int
+
+const (
+	// SystemLevel is the label to specify write directory to system level
+	SystemLevel WriteLevel = 0
+	// UserLevel is the label to specify write directory to user level
+	UserLevel WriteLevel = 1
+)
+
```
priteshbandi, line 48: Why levels are specific to write? can we make them generic so they be used for read too?

config/config.go
```diff
@@ -23,6 +25,8 @@ type Config struct {
 	InsecureRegistries       []string                 `json:"insecureRegistries"`
 	CredentialsStore         string                   `json:"credsStore,omitempty"`
 	CredentialHelpers        map[string]string        `json:"credHelpers,omitempty"`
+	// EnvelopeType defines the envelope type for signing
+	EnvelopeType string `json:"envelopeType,omitempty"`
```
priteshbandi, line 29: This change is out of scope for https://github.com/notaryproject/notation/issues/203, please remove it from this PR.

dir/base.go
```diff
@@ -93,12 +93,21 @@ func loadPath() {
 		UserConfigFS: NewUnionDirFS(
 			NewRootedFS(userConfig, nil),
 		),
+		SystemConfigFS: NewUnionDirFS(
+			NewRootedFS(userConfig, nil),
+		),
```
priteshbandi, line 98: Why do we need `SystemConfigFS` and `UserConfigFS` when we already have `ConfigFS`? I am assuming because `ConfigFS` tries to tries to read the user dir first then the system dir? If so, it raises the question why do we need `fs.go`? Can we get rid of it and the complications that come with it and rewrite it as simple utility methods which return paths based on the config we define(in spec)

dir/base.go
```diff
@@ -93,12 +93,21 @@ func loadPath() {
...
 		LibexecFS: NewUnionDirFS(
 			NewRootedFS(userLibexec, nil),
 			NewRootedFS(systemLibexec, nil),
 		),
+		UserLibexecFS: NewUnionDirFS(
+			NewRootedFS(userLibexec, nil),
+		),
+		SystemLibexecFS: NewUnionDirFS(
+			NewRootedFS(systemLibexec, nil),
+		),
```
priteshbandi, line 111: ditto

dir/path.go
```diff
@@ -69,27 +89,38 @@ func (p *PathManager) Localkey(name string) (keyPath, certPath string) {
...
 	return path
 }
 
-// TrustPolicy returns the path of trustpolicy.json file
+// TrustPolicy returns the ready-only path of trustpolicy.json file
 func (p *PathManager) TrustPolicy() string {
 	path, err := p.ConfigFS.GetPath(TrustPolicyFile)
 	checkError(err)
 	return path
 }
```
priteshbandi, line 104: Instead of creating a separate function for reading and writing, can we write a function that takes UserLevel/SystemLevel as a param and determines the path internally?

dir/path.go
```diff
@@ -69,27 +89,38 @@ func (p *PathManager) Localkey(name string) (keyPath, certPath string) {
...
 func (p *PathManager) SigningKeyConfig() string {
 	path, err := p.UserConfigFS.GetPath(SigningKeysFile)
 	checkError(err)
 	return path
 }
 
-// TrustPolicy returns the path of trustpolicy.json file
+// TrustPolicy returns the ready-only path of trustpolicy.json file
 func (p *PathManager) TrustPolicy() string {
 	path, err := p.ConfigFS.GetPath(TrustPolicyFile)
```
priteshbandi, line 101: The path of trystpolicy file is know(based on system or user level) then why are we searching that in directory using `GetPath`?

config/config.go
```diff
@@ -37,15 +41,30 @@ func NewConfig() *Config {
...
-func (c *Config) Save() error {
-	return save(ConfigPath, c)
+// Save stores the config to file.
+//
+// if the `path` is an empty string, it uses built-in user level config directory.
+func (c *Config) Save(path string) error {
+	// set default path
+	if path == "" {
+		path = dir.Path.ConfigForWrite(dir.UserLevel)
+	}
```
priteshbandi, line 51: Why do we need this if can we take this path from caller? 
Applicable for all load and save functions in config.go and keys.go.

dir/path.go
```diff
@@ -52,14 +67,19 @@ func checkError(err error) {
...
+// Config returns the ready-only path of config.json
 func (p *PathManager) Config() string {
 	path, err := p.ConfigFS.GetPath(ConfigFile)
 	checkError(err)
 	return path
 }
 
-// LocalKey returns path of the local private key and it's certificate
+// ConfigForWrite returns the writable path of config.json
+func (p *PathManager) ConfigForWrite(writeLevel WriteLevel) string {
```
priteshbandi, line 78: Instead of creating a separate function for reading and writing, can we write a function that takes UserLevel/SystemLevel as a param and determines the path internally?

Also, applicable to other similar functions.

rgnote (commented):
Is there a spec update for adding `EnvelopeType`? If yes, can you provide that link?

priteshbandi (commented):
Left some feedback on overall structure of config module. It looks like the `UnionDirFS` abstraction is making it complex to add system level and user level read/write permutations .

-------

Also, depending on decision we make in https://github.com/notaryproject/notation/issues/203, we might need to make significant changes in this PR.

config/config.go
JeyJeyGao: removed.

dir/base.go
JeyJeyGao: The reason that I uses io/fs package is `./plugin/manager/manager.go` uses io/fs to WalkDir the directory to list all available plugins. According to https://github.com/notaryproject/notation/blob/main/specs/directory.md, if both system and user level have plugins, it should walk through both of them to find plugins. The UnionDirFS in fs.go provides a fs.FS to read multiple directories, so the [manager.go](https://github.com/notaryproject/notation-go/blob/main/plugin/manager/manager.go#L91) can use it. 

The UnionDirFS can read multiple directories with priority, however, it can not distinguish user or system directory, so I add 3 FS to cover different use cases.
```
                // need to walk through both user and system level
		ConfigFS: NewUnionDirFS(
			NewRootedFS(userConfig, nil),
			NewRootedFS(systemConfig, nil),
		),
                // write to user level
		UserConfigFS: NewUnionDirFS(
			NewRootedFS(userConfig, nil),
		),
                // write to system level
		SystemConfigFS: NewUnionDirFS(
			NewRootedFS(userConfig, nil),
		),
```

dir/path.go
JeyJeyGao: GetPath will search both `user` and `system` level, and user level has higher priority, so the GetPath helps search and decide the priority.

config/config.go
JeyJeyGao: Because we have directory spec and it should be the default case (an empty path), and it is convenient for user to not specify the directory.

dir/path.go
JeyJeyGao: Sure, updated.
Also, renamed `WriteLevel` to `DirLevel`, and added `UnionLevel`

dir/path.go
JeyJeyGao: Updated.

dir/base.go
JeyJeyGao: Removed `UserConfigFS` and `SystemConfigFS`, please review it again.

dir/base.go
JeyJeyGao: Removed. Please review it again.

dir/path.go
JeyJeyGao: Renamed `WriteLevel` to `DirLevel`

JeyJeyGao:
> Added some feedback on overall structure of config module. It looks like the `UnionDirFS` abstraction is making it complex to add system level and user level read/write permutations, we should rethink about it .
> 
> Also, depending on decision we make in [notaryproject/notation#203](https://github.com/notaryproject/notation/issues/203), we might need to make significant changes in this PR.

Refactor it again. Now, `UnionDirFS` is only used to GetPath() in both system and user directory. If user specify the System/User level directory, return the path directly.

config/config.go
JeyJeyGao: Sure. Deleted.

priteshbandi:
We can close/update this PR and update/open new one for only supporting user level config

https://github.com/notaryproject/notation/issues/203#issuecomment-1281663886

PR #141, opened 2022-09-14
fix: optimize config package

JeyJeyGao:
change the config load & save function's `path` parameter to be a single string value instead of an array of string.


patrickzheng200 (approved):
LGTM

PR #145, opened 2022-09-21
fix: only check root cert is CA cert

binbin-li:
### What?

For a valid cert chain, we require that the root cert should be a CA cert. But the current implementation require all certs should satisfy with it. Update it to validate the root cert only.

### Test
Updated unit tests.

### DIFF
--- a/verification/store.go
+++ b/verification/store.go
@@ -56,10 +56,9 @@ func LoadX509TrustStore(path string) (*X509TrustStore, error) {
 		if len(certs) < 1 {
 			return nil, fmt.Errorf("could not parse a certificate from %q, every file in a trust store must have a PEM or DER certificate in it", joinedPath)
 		}
-		for _, cert := range certs {
-			if !cert.IsCA {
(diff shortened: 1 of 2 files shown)


patrickzheng200 (approved):
LGTM

verification/store.go
```diff
@@ -56,10 +56,9 @@ func LoadX509TrustStore(path string) (*X509TrustStore, error) {
 		if len(certs) < 1 {
 			return nil, fmt.Errorf("could not parse a certificate from %q, every file in a trust store must have a PEM or DER certificate in it", joinedPath)
 		}
-		for _, cert := range certs {
-			if !cert.IsCA {
-				return nil, fmt.Errorf("certificate with subject %q from file %q is not a CA certificate, only CA certificates (BasicConstraint CA=True) are allowed", cert.Subject, joinedPath)
-			}
+		rootCert := certs[len(certs) - 1]
```
patrickzheng200, line 59: per discussion, we need also check intermediate certs to be CAs.

PR #147, opened 2022-09-22
fix: fix the certs validation in trust store

binbin-li:
### What?

Change the certificates validation in a trust store.

As described in spec: https://github.com/notaryproject/notaryproject/blob/main/trust-store-trust-policy-specification.md#trust-store, certs in trust store can be either root certs or intermediate certs though the intermediate certs are not recommended. And it's possible that users put a self-signed root cert in the trust store as well.

The current implementation checks if all certs under a trust store directory are CA certs which misses the case that a self-signed root cert in the trust store.

### Test
Updated and added unit tests.


verification/store.go
```diff
@@ -75,3 +70,33 @@ func LoadX509TrustStore(path string) (*X509TrustStore, error) {
...
+
+func validateCerts(certs []*x509.Certificate, path string) error {
+	// to prevent any trust store misconfigurations, ensure there is at least
+	// one certificate from each file.
+	if len(certs) < 1 {
+		return fmt.Errorf("could not parse a certificate from %q, every file in a trust store must have a PEM or DER certificate in it", path)
+	}
+
+	if len(certs) == 1 {
+		// if there is only one certificate, it must be a self-signed cert or
```
priteshbandi, line 82: Single certificate in trust store doesn't means that its a self signed signing certificate. Trust store can have multiple certificate and one out of them can be a self signed signing certificate.

Logic:
```go
func validateCerts(certs[] * x509.Certificate, path string) error {
    // to prevent any trust store misconfigurations, ensure there is at least
    // one certificate from each file.
    if len(certs) < 1 {
        return fmt.Errorf("could not parse a certificate from %q, every file in a trust store must have a PEM or DER certificate in it", path)
    }
    for _, cert: = range certs {
        if !cert.IsCA {
                if err: = cert.CheckSignature(cert.SignatureAlgorithm, cert.RawTBSCertificate, cert.Signature);
                err != nil {
                    fmt.Errorf("certificate with subject %q from file %q is not a CA certificate or self-signed signing certificate", cert.Subject, joinedPath)
                }
            }
        }
    }
}

```

binbin-li: thanks for the comment! Updated.

priteshbandi (approved):
LGTM

verification/store.go
binbin-li: Actually, I found https://pkg.go.dev/crypto/x509#Certificate.CheckSignatureFrom might better fit this case, and it uses `CheckSignature` under the hood. Let me know if you have any concerns on it. thanks! @priteshbandi

shizhMSFT (approved):
LGTM

verification/store.go
priteshbandi: We cannot use [CheckSignatureFrom](https://pkg.go.dev/crypto/x509#Certificate.CheckSignatureFrom) because it imposes some extra constraints that are not valid. Instead we should use CheckSignature

PR #148, opened 2022-09-22
fix: fix the certs validation in trust store

binbin-li:
### What?

Change the certificates validation in a trust store.

As described in spec: https://github.com/notaryproject/notaryproject/blob/main/trust-store-trust-policy-specification.md#trust-store, certs in trust store can be either root certs or intermediate certs though the intermediate certs are not recommended. And it's possible that users put a self-signed non-CA root cert in the trust store as well.

The current implementation checks if all certs under a trust store directory are CA certs which misses the case that a self-signed non-CA root cert in the trust store.

### Test
Updated and added unit tests.


verification/store.go
```diff
@@ -75,3 +70,33 @@ func LoadX509TrustStore(path string) (*X509TrustStore, error) {
...
+	// to prevent any trust store misconfigurations, ensure there is at least
+	// one certificate from each file.
+	if len(certs) < 1 {
+		return fmt.Errorf("could not parse a certificate from %q, every file in a trust store must have a PEM or DER certificate in it", path)
+	}
+
+	if len(certs) == 1 {
+		// if there is only one certificate, it must be a self-signed cert or
+		// CA cert.
+		if !isSelfSigned(certs[0]) && !certs[0].IsCA {
```
JeyJeyGao, line 84: Root cert must be self-signed, so if it is not self signed, it is an invalid certificate chain, right? And if it is self signed root cert, we don't care whether it is CA.

```suggestion
		if !isSelfSigned(certs[0]) {
		    return fmt.Error()
		}
```

patrickzheng200 (approved):
LGTM

verification/store.go
binbin-li: Already updated the logic as Pritesh called out in https://github.com/notaryproject/notation-go/pull/147#discussion_r978098557

verification/store.go
```diff
@@ -75,3 +69,25 @@ func LoadX509TrustStore(path string) (*X509TrustStore, error) {
...
+func validateCerts(certs []*x509.Certificate, path string) error {
+	// to prevent any trust store misconfigurations, ensure there is at least
+	// one certificate from each file.
+	if len(certs) < 1 {
+		return fmt.Errorf("could not parse a certificate from %q, every file in a trust store must have a PEM or DER certificate in it", path)
+	}
+
+	for _, cert := range certs {
+		if !cert.IsCA {
+			if err := cert.CheckSignature(cert.SignatureAlgorithm, cert.RawTBSCertificate, cert.Signature); err != nil {
```
patrickzheng200, line 82: cert.CheckSignature does not guarantee self-signed right? If the cert is not a CA, then we need to make sure it's the only certificate in this certs slice, and it's self-signed.

binbin-li: Updated to use `CheckSignatureFrom` instead.

PR #158, opened 2022-09-29
fix: trustpolicy identity malformed check

JeyJeyGao:
Fix the issue:
previously the identity parameter format was not checked. The following identity config causes a runtime error.
```
trustpolicy.json
{...
 "trustedIdentities": [
              "CN=wabbit-networks.io,O=Notary,L=Seattle,ST=WA,C=US"
            ]
...
}
```

Test:
add unit test

### DIFF
--- a/verification/verifier_helpers.go
+++ b/verification/verifier_helpers.go
@@ -216,7 +216,7 @@ func verifyX509TrustedIdentities(certs []*x509.Certificate, trustPolicy *TrustPo
 	for _, identity := range trustPolicy.TrustedIdentities {
 		parts := strings.Split(identity, ":")
 		if len(parts) != 2 {
-			return fmt.Errorf("trustpolicy identity `%s` malformed.", identity)
+			return fmt.Errorf("trustpolicy identity %q is malformed.", identity)
 		}
 
 		identityPrefix := parts[0]
--- a/verification/verifier_helpers.go
+++ b/verification/verifier_helpers.go
@@ -214,10 +214,13 @@ func verifyX509TrustedIdentities(certs []*x509.Certificate, trustPolicy *TrustPo
 
 	var trustedX509Identities []map[string]string
 	for _, identity := range trustPolicy.TrustedIdentities {
-		i := strings.Index(identity, ":")
+		parts := strings.Split(identity, ":")
+		if len(parts) != 2 {
+			return fmt.Errorf("trustpolicy identity `%s` malformed.", identity)
+		}
 
-		identityPrefix := identity[:i]
-		identityValue := identity[i+1:]
+		identityPrefix := parts[0]
+		identityValue := parts[1]
 
 		if identityPrefix == x509Subject {
 			parsedSubject, err := parseDistinguishedName(identityValue)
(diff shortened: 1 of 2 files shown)


binbin-li (approved):
lgtm

verification/verifier_helpers.go
```diff
@@ -214,10 +214,13 @@ func verifyX509TrustedIdentities(certs []*x509.Certificate, trustPolicy *TrustPo
 
 	var trustedX509Identities []map[string]string
 	for _, identity := range trustPolicy.TrustedIdentities {
-		i := strings.Index(identity, ":")
+		parts := strings.Split(identity, ":")
+		if len(parts) != 2 {
+			return fmt.Errorf("trustpolicy identity `%s` malformed.", identity)
+		}
```
patrickzheng200, line 220: what if the identity is '*'?

chloeyin: > what if the identity is '*'?

I don't think [spec](https://github.com/notaryproject/notaryproject/blob/main/trust-store-trust-policy-specification.md#trusted-identities-constraints) supports `*` identities
```
The minimal possible value is x509.subject: C=${country}, ST=${state}, O={organization},
```

chloeyin (approved):

verification/verifier_helpers.go
patrickzheng200: > > what if the identity is '*'?
> 
> I don't think [spec](https://github.com/notaryproject/notaryproject/blob/main/trust-store-trust-policy-specification.md#trusted-identities-constraints) supports `*` identities
> 
> ```
> The minimal possible value is x509.subject: C=${country}, ST=${state}, O={organization},
> ```
According to the spec:
The trustedIdentities list items MUST support a full and partial list of all the attribute types present in [subject DN](https://www.rfc-editor.org/rfc/rfc5280.html#section-4.1.2.6) of x509 certificate. Alternatively, it supports a single item with value *, to indicate that any certificate that chains to the associated trust store (trustStore) is allowed.

patrickzheng200: `if isPresent(wildcard, trustPolicy.TrustedIdentities) {
		return nil
}`
does the '*' check at the top of this func, it's not in the change of this PR so I missed it. LGTM now.

verification/verifier_helpers.go
```diff
@@ -214,10 +214,13 @@ func verifyX509TrustedIdentities(certs []*x509.Certificate, trustPolicy *TrustPo
 
 	var trustedX509Identities []map[string]string
 	for _, identity := range trustPolicy.TrustedIdentities {
-		i := strings.Index(identity, ":")
+		parts := strings.Split(identity, ":")
+		if len(parts) != 2 {
+			return fmt.Errorf("trustpolicy identity `%s` malformed.", identity)
```
patrickzheng200, line 219:
```suggestion
			return fmt.Errorf("trustpolicy identity %q is malformed.", identity)
```

JeyJeyGao: Updated.

patrickzheng200 (approved):
LGTM

PR #159, opened 2022-09-30
feat: add envelope type config

JeyJeyGao:
Add `EnvelopeType` parameter in config.json file

Resolves: https://github.com/notaryproject/notation/issues/325

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -23,6 +23,8 @@ type Config struct {
 	InsecureRegistries       []string                 `json:"insecureRegistries"`
 	CredentialsStore         string                   `json:"credsStore,omitempty"`
 	CredentialHelpers        map[string]string        `json:"credHelpers,omitempty"`
+	// EnvelopeType defines the envelope type for signing
+	EnvelopeType string `json:"envelopeType,omitempty"`
 }
 
 // VerificationCertificates is a collection of public certs used for verification.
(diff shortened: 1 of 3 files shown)


shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

gokarnm:
Is `envelopeType` a required config property? What is the behavior if it's not specified?

If this is a required property, expecting end users to choose the envelope format is not the best user experience. In most cases user may not know the details of envelope format, or how they should choose one. Could we instead have a spec update defining a default format?

JeyJeyGao:
> Is `envelopeType` a required config property? What is the behavior if it's not specified?
> 
> If this is a required property, expecting end users to choose the envelope format is not the best user experience. In most cases user may not know the details of envelope format, or how they should choose one. Could we instead have a spec update defining a default format?

We should have a spec to expain the configuration details. `envelopeType` is not a required parameter. If it is not set, the default value is `jws`. Those logic will be added in notation-cli.

JeyJeyGao:
> Is `envelopeType` a required config property? What is the behavior if it's not specified?
> 
> If this is a required property, expecting end users to choose the envelope format is not the best user experience. In most cases user may not know the details of envelope format, or how they should choose one. Could we instead have a spec update defining a default format?

config description will be added in notation/specs/notation-config.md

PR #166, opened 2022-10-11
added alias for notation-core-go

patrickzheng200:
This is to remove dependency of notation-core-go in notation CLI. notation CLI will instead directly call this alias from notation-go.


core/testhelper/certificatetest.go
```diff
@@ -0,0 +1,19 @@
+package testhelper
+
+import (
+	"crypto/rsa"
+
+	"github.com/notaryproject/notation-core-go/testhelper"
+)
+
+type RSACertTuple = testhelper.RSACertTuple
```
binbin-li, line 9: comment on it?

binbin-li (approved):
minor comment, lgtm

PR #168, opened 2022-10-11
update: added alias for notation-core-go api

patrickzheng200:
This is to move notation-core-go to indirect dependency of notation CLI. Notation CLI will instead call this alias.


JeyJeyGao:
LGTM

core/signature/envelope.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+// RegisteredEnvelopeTypes lists registered envelope media types.
+func RegisteredEnvelopeTypes() []string {
+	return signature.RegisteredEnvelopeTypes()
+}
+
+// ParseEnvelope generates an envelope for given envelope bytes with specified
+// media type.
+func ParseEnvelope(mediaType string, envelopeBytes []byte) (signature.Envelope, error) {
```
shizhMSFT, line 14: You might also want
```go
type Envelope = signature.Envelope

func ParseEnvelope(mediaType string, envelopeBytes []byte) (Envelope, error) {
```

patrickzheng200: code updated

shizhMSFT (approved):
LGTM

priteshbandi (commented):
Why do we need this aliases? Can we modify our notation-go functions to not accept or return notation-core-go types? 

For test certificate we can have a new method in notation-go which will generate self signed cert for testing.

This will provide better abstraction.

patrickzheng200:
> Why do we need this aliases? Can we modify our notation-go functions to not accept or return notation-core-go types?
> 
> For test certificate we can have a new method in notation-go which will generate self signed cert for testing.
> 
> This will provide better abstraction.

notation-core-go is notation-go's direct dependency by design. The alias in this PR is only for notation CLI. Instead of directly calling notation-core-go, notation CLI will call this alias in the future. In this way, notation-core-go will become an indirect dependency of notation CLI, and it avoids duplicate implementations from notation-core-go.

priteshbandi:
> > Why do we need this aliases? Can we modify our notation-go functions to not accept or return notation-core-go types?
> > For test certificate we can have a new method in notation-go which will generate self signed cert for testing.
> > This will provide better abstraction.
> 
> notation-core-go is notation-go's direct dependency by design. The alias in this PR is only for notation CLI. Instead of directly calling notation-core-go, notation CLI will call this alias in the future. In this way, notation-core-go will become an indirect dependency of notation CLI, and it avoids duplicate implementations from notation-core-go.

Instead of creating, managing aliases, and possibly adding more in the future. Can we make APIs of notation-go, notation-core-go independent? I'm making an assumption here that an alias is only required because notation-go APIs expect parameters of types defined in notation-core-go

patrickzheng200:
> > > Why do we need this aliases? Can we modify our notation-go functions to not accept or return notation-core-go types?
> > > For test certificate we can have a new method in notation-go which will generate self signed cert for testing.
> > > This will provide better abstraction.
> > 
> > 
> > notation-core-go is notation-go's direct dependency by design. The alias in this PR is only for notation CLI. Instead of directly calling notation-core-go, notation CLI will call this alias in the future. In this way, notation-core-go will become an indirect dependency of notation CLI, and it avoids duplicate implementations from notation-core-go.
> 
> Instead of creating, managing aliases, and possibly adding more in the future. Can we make APIs of notation-go, notation-core-go independent? I'm making an assumption here that an alias is only required because notation-go APIs expect parameters of types defined in notation-core-go

In this case, then there will be duplicate implementations in notation-go and notation-core-go? For example, ReadCertificateFile() and generating self-signed certificates. And in the future, whenever there's a new method introduced in notation-core-go, and if notation CLI wants to use it, we need to add a copied version inside notation-go. This alias's only purpose is for [notation CLI](https://github.com/notaryproject/notation) to still use notation-core-go API, but in an indirect way. @shizhMSFT Any further suggestions/comments on this one? Thanks.

shizhMSFT:
@priteshbandi @patrickzheng200 The fundamental problem we have here is that we have to choose either the following approaches:
1. `notation` and other applications will import `notation-go` and `notation-core-go` for the corresponding functionalities.
    - Pros: Codes and purposes of two repos are cleaner, clearer, and decoupled.
    - Cons: Versions of `notation-go` and `notation-core-go` are tracked separately. It is possible to have mix-matched versions. However, I don't think it is a concern since the application should always update the version to the latest.
2. `notation` and other applications will import `notation-go` only. If they want to import anything from `notation-core-go`, they need to import through `notation-go`. This PR is for this approach.
    - Pros: `notation-core-go` is hidden from any application. Only a few APIs are exposed to the public / application via `notation-go`. The application only needs to track the version of `notation-go` and can safely leave `notation-core-go` as an indirect dependency.
    - Cons: Duplicated code or API forwarders introduced this PR.

I tend to the 1st approach, which is what we have today. This PR shows what will the libraries look like if we take the 2nd approach for discussions. Ideally, we will close this PR if we remain on the 1st approach after discussion.

/cc @gokarnm

rgnote:
+1 to @shizhMSFT's comment on avoiding API forwarders

patrickzheng200:
Closing this PR as we decided to keep what we have now, i.e. Notation CLI would still require notation-core-go as a direct dependency along with notation-go.

PR #170, opened 2022-10-17
feat: add system level `Harden` setting

JeyJeyGao:
1. isolate system/user level directory
2. add system level harden setting


priteshbandi:
We can close this PR because as per discussion in 10/17 Notary v2 meeting, for RC1 we will only support user level config and punt system level config discussion for RC2.

https://github.com/notaryproject/notation/issues/203#issuecomment-1281663886

PR #175, opened 2022-10-27
fix: rename envelopeType to signatureFormat

JeyJeyGao:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -23,8 +23,8 @@ type Config struct {
 	InsecureRegistries       []string                 `json:"insecureRegistries"`
 	CredentialsStore         string                   `json:"credsStore,omitempty"`
 	CredentialHelpers        map[string]string        `json:"credHelpers,omitempty"`
-	// EnvelopeType defines the envelope type for signing
-	EnvelopeType string `json:"envelopeType,omitempty"`
+	// SignatureFormat defines the signature envelope type for signing
+	SignatureFormat string `json:"signatureFormat,omitempty"`
 }
 
 // VerificationCertificates is a collection of public certs used for verification.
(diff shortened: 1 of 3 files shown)


rgnote:
The targeted users of Notation CLI may not know what is an envelope. SignatureFormat is more reasonable for a dummy user.

config/config.go
```diff
@@ -23,8 +23,8 @@ type Config struct {
 	InsecureRegistries       []string                 `json:"insecureRegistries"`
 	CredentialsStore         string                   `json:"credsStore,omitempty"`
 	CredentialHelpers        map[string]string        `json:"credHelpers,omitempty"`
-	// EnvelopeType defines the envelope type for signing
-	EnvelopeType string `json:"envelopeType,omitempty"`
+	// SignatureFormat defines the envelope type for signing
```
patrickzheng200, line 26:
```suggestion
	// SignatureFormat defines the signature envelope type for signing
```

config/config.go
```diff
@@ -23,8 +23,8 @@ type Config struct {
 	InsecureRegistries       []string                 `json:"insecureRegistries"`
 	CredentialsStore         string                   `json:"credsStore,omitempty"`
 	CredentialHelpers        map[string]string        `json:"credHelpers,omitempty"`
```
priteshbandi, line 25: Do we need `CredentialsStore`, `VerificationCertificates` and `CredentialHelpers` ?

priteshbandi (approved):
LGTM, Added one comment for future cleanup.

rgnote (approved):

config/config.go
JeyJeyGao: CredentialsStore and CredentialHelpers are using for login/logout command.
VerificationCertificates will be no longer used after TS/TP.

config/config.go
JeyJeyGao: Done.

shizhMSFT (approved):
LGTM

PR #177, opened 2022-10-31
update: bump up notation-core-go

patrickzheng200:
This PR bumps up notation-core-go. Since extended attribute key type is changed to interface{}, notation-go code needs to be updated accordingly.

### DIFF
--- a/verification/helpers.go
+++ b/verification/helpers.go
@@ -69,6 +69,15 @@ func isPresent(val string, values []string) bool {
 	return false
 }
 
+func isPresentAny(val interface{}, values []interface{}) bool {
+	for _, v := range values {
+		if v == val {
+			return true
+		}
+	}
+	return false
+}
+
 func getArtifactPathFromUri(artifactUri string) (string, error) {
 	// TODO support more types of URI like "domain.com/repository", "domain.com/repository:tag"
 	i := strings.LastIndex(artifactUri, "@")
--- a/verification/verifier_helpers.go
+++ b/verification/verifier_helpers.go
@@ -253,8 +253,8 @@ func (v *Verifier) executePlugin(ctx context.Context, trustPolicy *TrustPolicy, 
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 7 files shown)


verification/helpers.go
```diff
@@ -69,6 +69,15 @@ func isPresent(val string, values []string) bool {
 	return false
 }
 
+func isPresentAny(val interface{}, values []interface{}) bool {
```
priteshbandi, line 72: qq: can we get rid of function defined at line 63?

priteshbandi (commented):
Can you please update CR with latest changes from notation-core-go

verification/helpers.go
patrickzheng200: > qq: can we get rid of function defined at line 63?

the function at line 63 is used in other places such as policy.go. For example, `isPresent(wildcard, statement.RegistryScopes)`.

priteshbandi (approved):
LGTM

verification/helpers.go
shizhMSFT: Can we use generics like 
```go
func isPresent[T comparable](val T, values []T) bool {
	for _, v := range values {
		if v == val {
			return true
		}
	}
	return false
}
```

shizhMSFT (approved):
LGTM

verification/helpers.go
shizhMSFT: LoL.
```
any does not implement comparable
```

verification/verifier_helpers.go
```diff
@@ -253,8 +253,8 @@ func (v *Verifier) executePlugin(ctx context.Context, trustPolicy *TrustPolicy,
 	if err != nil {
 		return nil, err
 	}
-	var attributesToProcess []string
-	extendedAttributes := make(map[string]interface{})
+	var attributesToProcess []interface{}
+	extendedAttributes := make(map[interface{}]interface{})
```
rgnote, line 257: This can't be marshaled using `encoding/json`

Issue #80, opened 2022-07-11
Tracks multiple TODOs in signature package


priteshbandi:
- [x] include external signing plugin's name and version in signing agent.
- [x] Verify plugin didnot add any additional top level payload attributes.
- [x] Add tests for this method. NewSignerFromFiles

https://github.com/notaryproject/notation-go/pull/77

The associated Linked PullRequest - https://github.com/notaryproject/notation-go/pull/178

PR #178, opened 2022-10-31
Add additional header validation for payload

jondonas:
Addresses TODO's tracked in [Issue 80](https://github.com/notaryproject/notation-go/issues/80)

- Note that new tests for NewSignerFromFiles were already added back [in September](https://github.com/notaryproject/notation-go/commit/2bcfd343f974938e8eeb76e233bc2526e0c36061)

- Validations around no top-level attributes being added are from [5.c.c](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#signing-workflow-using-plugin-1) in the spec

Signed-off-by: Jonathan Donas <jddonas@amazon.com>

### DIFF
--- a/signer/plugin.go
+++ b/signer/plugin.go
@@ -59,14 +59,14 @@ func (s *pluginSigner) Sign(ctx context.Context, desc ocispec.Descriptor, opts n
 
 	logger.Debugf("Using plugin %v with capabilities %v to sign artifact %v in signature media type %v", metadata.Name, metadata.Capabilities, desc.Digest, opts.SignatureMediaType)
 	if metadata.HasCapability(proto.CapabilitySignatureGenerator) {
-		return s.generateSignature(ctx, desc, opts)
+		return s.generateSignature(ctx, desc, opts, metadata)
 	} else if metadata.HasCapability(proto.CapabilityEnvelopeGenerator) {
 		return s.generateSignatureEnvelope(ctx, desc, opts)
 	}
 	return nil, nil, fmt.Errorf("plugin does not have signing capabilities")
 }
 
-func (s *pluginSigner) generateSignature(ctx context.Context, desc ocispec.Descriptor, opts notation.SignOptions) ([]byte, *signature.SignerInfo, error) {
+func (s *pluginSigner) generateSignature(ctx context.Context, desc ocispec.Descriptor, opts notation.SignOptions, metadata *proto.GetMetadataResponse) ([]byte, *signature.SignerInfo, error) {
 	logger := log.GetLogger(ctx)
 	logger.Debug("Generating signature by plugin")
 	config := s.mergeConfig(opts.PluginConfig)
@@ -94,6 +94,8 @@ func (s *pluginSigner) generateSignature(ctx context.Context, desc ocispec.Descr
 			keySpec:      ks,
 		},
 	}
+
+	opts.SigningAgent = fmt.Sprintf("%s %s/%s", signingAgent, metadata.Name, metadata.Version)
 	return genericSigner.Sign(ctx, desc, opts)
 }
 
@@ -141,15 +143,18 @@ func (s *pluginSigner) generateSignatureEnvelope(ctx context.Context, desc ocisp
 		return nil, nil, err
 	}
 
+	content := envContent.Payload.Content
 	var signedPayload envelope.Payload
-	if err = json.Unmarshal(envContent.Payload.Content, &signedPayload); err != nil {
+	if err = json.Unmarshal(content, &signedPayload); err != nil {
 		return nil, nil, fmt.Errorf("signed envelope payload can't be unmarshalled: %w", err)
 	}
 
-	// TODO: Verify plugin did not add any additional top level payload
-	// attributes. https://github.com/notaryproject/notation-go/issues/80
-	if !isDescriptorSubset(desc, signedPayload.TargetArtifact) {
+	if !isPayloadDescriptorValid(desc, signedPayload.TargetArtifact) {
 		return nil, nil, fmt.Errorf("during signing descriptor subject has changed from %+v to %+v", desc, signedPayload.TargetArtifact)
+	}
+
+	if unknownAttributes := areUnknownAttributesAdded(content); len(unknownAttributes) != 0 {
+		return nil, nil, fmt.Errorf("during signing, following unknown attributes were added to subject descriptor: %+q", unknownAttributes)
 	}
 
 	return resp.SignatureEnvelope, &envContent.SignerInfo, nil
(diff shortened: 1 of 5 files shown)


signature/plugin.go
```diff
@@ -123,6 +123,7 @@ func (s *pluginSigner) generateSignature(ctx context.Context, desc notation.Desc
 		}
 		extProvider.prepareSigning(config, ks)
 	}
+	signingAgentId := fmt.Sprintf("Signing Agent: %s - Signing Plugin Name: %s - Signing Plugin Version: %s", notation.SigningAgent, metadata.Name, metadata.Version)
```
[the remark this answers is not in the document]
priteshbandi: Please add unit-test for this change.

signature/plugin.go
```diff
@@ -220,25 +221,42 @@ func (s *pluginSigner) generateSignatureEnvelope(ctx context.Context, desc notat
...
+func isPayloadDescriptorValid(originalContent []byte, originalDesc, newDesc notation.Descriptor) bool {
+	return originalDesc.Equal(newDesc) &&
+		!isDescriptorAttributeAdded(originalContent) &&
+		descriptorAnnotationsPartialEqual(originalDesc, newDesc)
+}
+
+// isDescriptorAttributeAdded checks that the unmarshalled json content does not have any unexpected keys added
+func isDescriptorAttributeAdded(content []byte) bool {
+	var targetArtifactMap map[string]interface{}
+	_ = json.Unmarshal(content, &targetArtifactMap)
```
priteshbandi, line 246: We shouldn't ignore error.

signature/plugin.go
```diff
@@ -220,25 +221,42 @@ func (s *pluginSigner) generateSignatureEnvelope(ctx context.Context, desc notat
...
+	delete(descriptor, "mediaType")
+	delete(descriptor, "digest")
+	delete(descriptor, "size")
+	delete(descriptor, "annotations")
+
+	return len(targetArtifactMap) != 1 || len(descriptor) != 0
+}
+
+// descriptorAnnotationsPartialEqual checks that newDesc hasn't replaced or overridden existing annotations.
+func descriptorAnnotationsPartialEqual(original, newDesc notation.Descriptor) bool {
```
priteshbandi, line 259: Should we move this method to https://github.com/notaryproject/notation-go/blob/7ad4eca1a568c4dc7bc612b589b11ef0010301b4/notation.go#L19-L36

```
// Equal reports whether d and t points to the same content.
func (d Descriptor) isSubset(t Descriptor) bool {
	return d.equals*(t) .....
}
```

signature/plugin.go
jondonas: Error is not possible because we already successfully unmarshalled at this point

signature/plugin.go
```diff
@@ -220,25 +221,42 @@ func (s *pluginSigner) generateSignatureEnvelope(ctx context.Context, desc notat
...
-		return false
-	}
+func isPayloadDescriptorValid(originalContent []byte, originalDesc, newDesc notation.Descriptor) bool {
+	return originalDesc.Equal(newDesc) &&
+		!isDescriptorAttributeAdded(originalContent) &&
+		descriptorAnnotationsPartialEqual(originalDesc, newDesc)
+}
+
+// isDescriptorAttributeAdded checks that the unmarshalled json content does not have any unexpected keys added
+func isDescriptorAttributeAdded(content []byte) bool {
```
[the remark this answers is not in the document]
jondonas: Added an issue to track a unit test for this: https://github.com/notaryproject/notation-go/issues/194

signature/plugin.go
priteshbandi: Can you please add that comment to the code?

vaninrao10:
@priteshbandi can review to close the issue tied to" [Tracks multiple TODOs in signature package#80](https://github.com/notaryproject/notation-go/issues/80)"

signature/plugin.go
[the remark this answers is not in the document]
rgnote: Can you clarify what code refactoring is needed to be able to write unit tests?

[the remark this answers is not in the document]
jondonas: Hard to say. I went over this with Pritesh and we couldn't think of a way to finish the unit tests so he also suggested to cut an issue to address later so this doesn't block the PR

yizha1:
@jondonas could you rebase this PR first?

signer/plugin.go
```diff
@@ -87,6 +87,7 @@ func (s *pluginSigner) generateSignature(ctx context.Context, desc ocispec.Descr
 		},
 	}
 
+	opts.SigningAgent = fmt.Sprintf("%s;%s/%s", signingAgent, metadata.Name, metadata.Version)
```
shizhMSFT, line 90: You may consider using the same format as [User Agent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent)

jondonas: Not sure that format directly applies to the info we're wanting to convey here. Unless you feel strongly about it, would prefer to keep as-is

shizhMSFT: @yizha1 @FeynmanZhou @priteshbandi for comments

priteshbandi: Yeah, lets create an issue (https://github.com/notaryproject/notation-go/issues/225) to discuss about using browser like user agent for rc2. Since PR is improvement over the existing hard coded user agent, can we please go ahead and merge it?

priteshbandi: @jondonas updated the format to abide by [rfc9110.html#field.user-agent](https://httpwg.org/specs/rfc9110.html#field.user-agent)

priteshbandi (approved):
LGTM

shizhMSFT (approved):
LGTM

signer/plugin.go
```diff
@@ -141,17 +143,20 @@ func (s *pluginSigner) generateSignatureEnvelope(ctx context.Context, desc ocisp
...
 
-	// TODO: Verify plugin did not add any additional top level payload
-	// attributes. https://github.com/notaryproject/notation-go/issues/80
-	if !isDescriptorSubset(desc, signedPayload.TargetArtifact) {
+	if !isPayloadDescriptorValid(desc, signedPayload.TargetArtifact) {
 		return nil, nil, fmt.Errorf("during signing descriptor subject has changed from %+v to %+v", desc, signedPayload.TargetArtifact)
 	}
 
+	if unknownAttributes := areUnknownAttributesAdded(content); len(unknownAttributes) != 0 {
+		return nil, nil, fmt.Errorf("during signing descriptor subject added unknown attributes %+q", unknownAttributes)
```
priteshbandi, line 157:
```suggestion
		return nil, nil, fmt.Errorf("during signing, following unknown attributes were added to subject descriptor, %+q", unknownAttributes)
```

PR #179, opened 2022-11-01
refactor: dir package [1]

JeyJeyGao:
```go
// The relative path to {NOTATION_CONFIG}
const (
    PathConfigFile  = "config.json"
    PathSigningKeys = "signingkeys.json"
    PathTrustPolicy = "trustpolicy.json"
)

func LocalKeyPath(name string) (keyPath, certPath string)
func X509TrustStoreDir(items... string) string

var (
    UserConfigDir    string // Aboslute path of user level {NOTATION_CONFIG}
    UserLibexecDir   string // Aboslute path of user level {NOTATION_LIBEXEC}
)

// ConfigFS and PluginFS can be an union fs
func ConfigFS() SysFS
func PluginFS() SysFS

type SysFS interface {
    fs.FS

    SysPath(items ...string) (string, error)
}
    func NewSysFS(root string) SysFS
```

Please review this PR before others.
Signed-off-by: Junjie Gao <junjiegao@microsoft.com>

PR #182, opened 2022-11-02
refactor: config package

JeyJeyGao:
```go
type Config struct
    func LoadConfig() (*Config, error)
    func NewConfig() *Config
    func (c *Config) Save() error

type SigningKeys struct
    func LoadSigningKeys() (*SigningKeys, error)
    func NewSigningKeys() *SigningKeys
    func (s *SigningKeys) Save() error
type KeySuite
    func (k KeySuite) Is(name string) bool
type X509KeyPair
type ExternalKey
```
Please review this PR firstly: https://github.com/notaryproject/notation-go/pull/179


shizhMSFT (approved):
LGTM

shizhMSFT (commented):
It will be better if we add a package documentation in `config.go`.

config/config.go
```diff
@@ -1,3 +1,16 @@
...
+// Example:
+// - load config.json
+//   cfg, err := LoadConfig()
+//
+// - save config.json
+//   cfg.Save()
+//
+// Currently, Only loading from and saving to the built-in directory
+// are supported.
+
```
shizhMSFT, line 13: There is no empty line between the documentation and the line `package config`.

config/config.go
```diff
@@ -1,3 +1,16 @@
+// package config provides the ability to load and save config.json and
+// signingkeys.json files.
+//
+// Example:
+// - load config.json
+//   cfg, err := LoadConfig()
+//
+// - save config.json
+//   cfg.Save()
```
shizhMSFT, line 9: We can remove the examples for now as those examples should be in `example_test.go` later.

config/config.go
```diff
@@ -1,3 +1,16 @@
+// package config provides the ability to load and save config.json and
```
shizhMSFT, line 1: `P` should be capitalized as per golang doc convension.
```suggestion
// Package config provides the ability to load and save config.json and
```

config/config.go
```diff
@@ -1,3 +1,16 @@
...
+//
+// Example:
+// - load config.json
+//   cfg, err := LoadConfig()
+//
+// - save config.json
+//   cfg.Save()
+//
+// Currently, Only loading from and saving to the built-in directory
+// are supported.
```
shizhMSFT, line 12: nit: We don't need to mention this here. Suggest to remove

shizhMSFT (approved):
LGTM

PR #186, opened 2022-11-07
update: Package verification refactoring

patrickzheng200:
PR review order: This PR -> #200.

This PR refactors the verifier package. Files intended to be reviewed in this PR:
1. Files in verifier folder except verifier/trustpolicy and verifier/truststore (they should already been reviewed in #192).
2. internal/envelope
3. notation.go, this is for logic change on only returning successful signature verification outcomes.

(This PR incorporates the refactored plugin and plugin/proto into the verifier package.)

This PR removes internal/mock_origin, internal/notation.go, and internal/registry, as they've been refactored in the previous PRs. internal/plugin is there for compile reason. After the refactoring of Sign (#200 ), internal/plugin shall be removed.

This PR does not change code related to Sign, so please skip the signature package for now. (Refactoring of signature package is inside #200)

Edit:
A brief summary after our multiple rounds of discussions:
1. If verification succeeded, one and only one success outcome will be returned.
2. If verification failed,
         a. if the total number of signatures associated with the artifact is larger than the user given MaxSignatureAttempts, errExceededMaxVerificationLimit will be returned.
         b. Otherwise, ErrorVerificationFailed will be returned.


shizhMSFT:
This PR will be split into several smaller PRs to be reviewed.

notation/errors.go
```diff
@@ -1,49 +1,49 @@
-package verification
+package notation
```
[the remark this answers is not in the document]
patrickzheng200: This is due to import cycle. Both notation and verifier packages import errors.go. verifier package also imports the notation pakcage, which means errors.go cannot be in verifier package. Will update the doc to reflect this change.

notation/notation.go
```diff
@@ -0,0 +1,295 @@
...
+	SignedAnnotations map[string]string
+	// Error that caused the verification to fail (if it fails)
+	Error error
+}
+
+// Verifier is a generic interface for verifying an artifact.
+type Verifier interface {
+	// Verify verifies the signature blob and returns the verified descriptor
+	// upon successful verification.
+	Verify(ctx context.Context, signature []byte, opts VerifyOptions, outcome *VerificationOutcome) (Descriptor, error)
```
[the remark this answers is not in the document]
patrickzheng200: 1. Code updated.
2. Verifier.Verify only validates a single signature by design. This is more convenient for the Verify function, as it will list all the signatures associated with the artifact.

notation/notation.go
```diff
@@ -0,0 +1,295 @@
...
+
+	// TrustPolicyDocument gets the validated trust policy document.
+	TrustPolicyDocument() (*trustpolicy.Document, error)
```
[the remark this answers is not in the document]
patrickzheng200: This is because Verifier is an interface, it does not contain field trust policy. TrustPolicyDocument() is needed in notation.Verify function, as it takes the Verifier interface as input instead of an implementation.

verifier/helpers.go
```diff
@@ -0,0 +1,87 @@
...
+func isPresentAny(val interface{}, values []interface{}) bool {
+	for _, v := range values {
+		if v == val {
+			return true
+		}
+	}
+	return false
+}
+
+func getArtifactDigestFromReference(artifactReference string) (string, error) {
```
priteshbandi, line 74: is it being only used in test?

verifier/helpers.go
```diff
@@ -0,0 +1,87 @@
...
+		if err != nil {
+			return nil, err
+		}
+		certificates = append(certificates, certs...)
+		namedStoreSet[trustStore] = struct{}{}
+	}
+	return certificates, nil
+}
+
+func isPresentAny(val interface{}, values []interface{}) bool {
```
priteshbandi, line 65: Where are we using this?

verifier/helpers.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	defer jsonFile.Close()
+	err = json.NewDecoder(jsonFile).Decode(policyDocument)
+	if err != nil {
+		return nil, err
+	}
+	return policyDocument, nil
+}
+
+func loadX509TrustStores(ctx context.Context, scheme signature.SigningScheme, policy *trustpolicy.TrustPolicy) ([]*x509.Certificate, error) {
```
priteshbandi, line 30: Do we still need this method when we have `GetCertificates` in truststore.go?

verifier/verifier.go
```diff
@@ -0,0 +1,265 @@
...
+		return nil, err
+	}
+	if err = policyDocument.Validate(); err != nil {
+		return nil, err
+	}
+
+	return &verifier{
+		TrustPolicyDoc: policyDocument,
+		PluginManager:  manager.New(dir.PluginFS()),
+	}, nil
```
priteshbandi, line 43:
```suggestion
	return NewVerifier(policyDocument, manager.New(dir.PluginFS()))
```

verifier/helpers.go
```diff
@@ -0,0 +1,87 @@
...
+		typeToLoad = truststore.TypeCA
+	} else if scheme == signature.SigningSchemeX509SigningAuthority {
+		typeToLoad = truststore.TypeSigningAuthority
+	} else {
+		return nil, fmt.Errorf("unrecognized signing scheme %q", scheme)
+	}
+
+	var namedStoreSet = make(map[string]struct{})
+	var certificates []*x509.Certificate
+	x509TrustStore := truststore.NewX509TrustStore(dir.ConfigFS())
```
priteshbandi, line 42: We don't need to crawl directory to find correct truststore dirctory.

verifier/helpers.go
```diff
@@ -0,0 +1,87 @@
...
+	var typeToLoad truststore.Type
+	if scheme == signature.SigningSchemeX509 {
+		typeToLoad = truststore.TypeCA
+	} else if scheme == signature.SigningSchemeX509SigningAuthority {
+		typeToLoad = truststore.TypeSigningAuthority
+	} else {
+		return nil, fmt.Errorf("unrecognized signing scheme %q", scheme)
+	}
+
+	var namedStoreSet = make(map[string]struct{})
```
priteshbandi, line 40:
```suggestion
	var processedStoreSet = make(map[string]struct{})
```

verifier/verifier.go
```diff
@@ -0,0 +1,265 @@
...
+		TrustPolicyDoc: trustPolicy,
+		PluginManager:  pluginManager,
+	}, nil
+}
+
+// TrustPolicyDocument gets the validated trust policy document.
+func (v *verifier) TrustPolicyDocument() (*trustpolicy.Document, error) {
+	if err := v.TrustPolicyDoc.Validate(); err != nil {
+		return nil, err
+	}
```
priteshbandi, line 61: `Validate()` is being called when `New` or `NewVerifier`. Do we still need to call `Validate()` here?

internal/common/common.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+	return false
+}
+
+// Internal type to hold raw and parsed Distinguished Names
+type ParsedDN struct {
+	RawString string
+	ParsedMap map[string]string
+}
```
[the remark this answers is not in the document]
patrickzheng200: Code updated. This is moved into trustpolicy package.

verifier/helpers.go
```diff
@@ -0,0 +1,87 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/notaryproject/notation-core-go/signature"
+	"github.com/notaryproject/notation-go/dir"
+	"github.com/notaryproject/notation-go/verifier/trustpolicy"
+	"github.com/notaryproject/notation-go/verifier/truststore"
+)
+
+func loadPolicyDocument() (*trustpolicy.Document, error) {
```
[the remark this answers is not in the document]
patrickzheng200: Logically, when this method is called, we don't have a trustpolicy.Document yet. This method should load and return one to us.

verifier/helpers.go
patrickzheng200: It is used in processPluginResponse to verify all extended critical attributes are processed by the plugin. We cannot use slice.Contains here because interface{} does not implement comparable.

verifier/helpers.go
patrickzheng200: loadX509TrustStores is an unexported helper function to parse the truststores inside the trust policy statement. GetCertificates takes store type and named store to get the certificates in each named store. So we still need loadX509TrustStores to parse the trust policy statement.

verifier/verifier.go
```diff
@@ -0,0 +1,265 @@
...
+
+	payload := &envelope.Payload{}
+	err = json.Unmarshal(outcome.EnvelopeContent.Payload.Content, payload)
+	if err != nil {
+		outcome.Error = err
+		return notation.Descriptor{}, outcome, err
+	}
+	outcome.SignedAnnotations = payload.TargetArtifact.Annotations
+
+	return payload.TargetArtifact, outcome, nil
```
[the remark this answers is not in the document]
patrickzheng200: Had a discussion earlier with @shizhMSFT on this one. For verifier.Verify(), the upstream caller should verify if the payload.TargetArtifact matches the actual target artifact being verified. notation.Verify() function contains this logic.
For users using verifier.Verify() only, this last step should be verified outside notation-go.

verifier/helpers.go
patrickzheng200: This is used inside x509TrustStore.GetCertificates to fetch the certs. To find correct truststore directory we will just parse the trust policy statement as shown in the code below.

verifier/verifier.go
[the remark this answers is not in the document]
priteshbandi: I am worried that user of librart might miss to do that and consider verification success as overall success.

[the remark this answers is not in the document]
patrickzheng200: > I am worried that user of library might miss to do that and consider verification success as overall success.

verifier.Verify() only verifies the signature by design. I added a comment in verifier.Verify()'s func doc to address this.
(Note: verifier.Verify() does not have repo related information. In other words, we are not able to resolve artifact reference to its descriptor inside this function.)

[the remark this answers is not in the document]
rgnote: I agree with @priteshbandi here. This is assuming best intentions from the library callers. Can we take the artifact descriptor as an input for this method?

signature/plugin.go
```diff
@@ -259,3 +268,10 @@ func parseCertChain(certChain [][]byte) ([]*x509.Certificate, error) {
 	}
 	return certs, nil
 }
+
+// Equal reports whether d and t points to the same content.
+func equal(d *ocispec.Descriptor, t *ocispec.Descriptor) bool {
+	return d.MediaType == t.MediaType &&
+		d.Digest == t.Digest &&
+		d.Size == t.Size
+}
```
shizhMSFT, line 277: Replace with `content.Equal()` from `oras-go`?

verifier/helpers.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer jsonFile.Close()
+	err = json.NewDecoder(jsonFile).Decode(policyDocument)
+	if err != nil {
+		return nil, err
+	}
+	return policyDocument, nil
+}
```
shizhMSFT, line 28: Should this method be promoted to the package `trustpolicy` as `LoadDocument() (*Document, error)`?

verifier/helpers.go
```diff
@@ -0,0 +1,72 @@
...
+	"strings"
+
+	"github.com/notaryproject/notation-core-go/signature"
+	"github.com/notaryproject/notation-go/dir"
+	"github.com/notaryproject/notation-go/verifier/trustpolicy"
+	"github.com/notaryproject/notation-go/verifier/truststore"
+)
+
+func loadPolicyDocument() (*trustpolicy.Document, error) {
+	policyDocument := &trustpolicy.Document{}
```
shizhMSFT, line 17: nit: code style: this line should be between line 22 and 23.

verifier/helpers.go
```diff
@@ -0,0 +1,72 @@
...
+
+func loadX509TrustStores(ctx context.Context, scheme signature.SigningScheme, policy *trustpolicy.TrustPolicy) ([]*x509.Certificate, error) {
+	var typeToLoad truststore.Type
+	if scheme == signature.SigningSchemeX509 {
+		typeToLoad = truststore.TypeCA
+	} else if scheme == signature.SigningSchemeX509SigningAuthority {
+		typeToLoad = truststore.TypeSigningAuthority
+	} else {
+		return nil, fmt.Errorf("unrecognized signing scheme %q", scheme)
+	}
```
shizhMSFT, line 38: Will a `switch` be better?

verifier/helpers.go
```diff
@@ -0,0 +1,72 @@
...
+		if err != nil {
+			return nil, err
+		}
+		certificates = append(certificates, certs...)
+		processedStoreSet[trustStore] = struct{}{}
+	}
+	return certificates, nil
+}
+
+func isPresentAny(val interface{}, values []interface{}) bool {
```
shizhMSFT, line 65: Can we move this function to `internal/slice` as
```golang
func ContainsAny(s []any, v any) bool
```

We are not able to reuse `Contains` since `any` (i.e. `interface{}`) is not a `comparable`.

verifier/verifier.go
```diff
@@ -0,0 +1,248 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/notaryproject/notation-go"
+	"github.com/notaryproject/notation-go/dir"
+	"github.com/notaryproject/notation-go/internal/envelope"
+	"github.com/notaryproject/notation-go/internal/plugin"
+	"github.com/notaryproject/notation-go/internal/plugin/manager"
```
shizhMSFT, line 13: Can you update the code to use the refactored `plugin` package?
See https://github.com/notaryproject/notation-go/tree/main/plugin

verifier/verifier.go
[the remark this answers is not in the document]
shizhMSFT: I agree with @priteshbandi and @rgnote. Since library users are not security experts, they may think that the artifact is verified after a successful call to the `Verifier.Verify()` even though they did not pass the artifact to the `Verify()` function.
> [Murphy's law](https://en.wikipedia.org/wiki/Murphy's_law): Anything that can go wrong will go wrong.

How about 
```golang
type Verifier interface {
    Verify(ctx context.Context, desc ocispec.Descriptor, signature []byte, opts notation.VerifyOptions) (*notation.VerificationOutcome, error)
}
```

signature/plugin.go
patrickzheng200: Updated the code. But note this PR is mainly about the verification refactoring. Please skip the signature package for now.

verifier/verifier.go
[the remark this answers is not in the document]
patrickzheng200: Updated the interface.

internal/slice/slice.go
```diff
@@ -9,3 +9,13 @@ func Contains[E comparable](s []E, v E) bool {
 	}
 	return false
 }
+
+// ContainsAny reports whether val is present in values
+func ContainsAny(values []interface{}, val interface{}) bool {
```
shizhMSFT, line 14: nit: We can use `any` to replace `interface{}` since go `1.18`.

shizhMSFT, line 14: nit: code style: we might want to make the variable name consistant with `Contains()`.
```suggestion
func ContainsAny(s []any, v any) bool {
```

notation.go
```diff
@@ -139,7 +138,7 @@ type Verifier interface {
 	// descriptor upon successful verification.
 	// If nil signature is present and the verification level is not 'skip',
 	// an error will be returned.
```
shizhMSFT, line 140: You may also want to update this doc

verifier/trustpolicy/trustpolicy.go
```diff
@@ -248,6 +250,20 @@ func (trustPolicyDoc *Document) GetApplicableTrustPolicy(artifactReference strin
 	}
 }
 
+func LoadPolicyDocument() (*Document, error) {
```
shizhMSFT, line 253: Missing documentation

verifier/verifier.go
```diff
@@ -72,17 +73,23 @@ func (v *verifier) Verify(ctx context.Context, signature []byte, opts notation.V
...
 	}
 
-	return payload.TargetArtifact, outcome, nil
+	if !content.Equal(payload.TargetArtifact, desc) {
+		outcome.Error = errors.New("content descriptor mismatch")
+		err = outcome.Error
+		return outcome, err
+	}
+
+	return outcome, nil
```
shizhMSFT, line 92: It can be simplified as
```suggestion
	if !content.Equal(payload.TargetArtifact, desc) {
		outcome.Error = errors.New("content descriptor mismatch")
	}

	return outcome, outcome.Error
```

shizhMSFT (commented):
In `verifier_helpers.go`, not all receiver functions (i.e. methods associated with structs) are indeed receiver functions. We can make it a "static function".

Meanwhile, I suggest moving all receiver functions of the `verifier` struct to `verifier.go` and all others to `helper.go`, and cleans up `verifier_helpers.go`.

verifier/verifier.go
```diff
@@ -0,0 +1,255 @@
...
+	pluginManager  pluginManager
+}
+
+// pluginManager is for mocking in unit tests
+type pluginManager interface {
+	Get(ctx context.Context, name string) (*manager.Plugin, error)
+	Runner(name string) (plugin.Runner, error)
+}
+
+// New a verifier based on local file system
```
rgnote, line 33:
```suggestion
// New returns a verifier based on local file system
```

verifier/verifier.go
```diff
@@ -0,0 +1,255 @@
...
+	pluginConfig := opts.PluginConfig
+
+	trustPolicy, err := v.trustPolicyDoc.GetApplicableTrustPolicy(artifactRef)
+	if err != nil {
+		return nil, notation.ErrorNoApplicableTrustPolicy{Msg: err.Error()}
+	}
+	// ignore the error since we already validated the policy document
+	verificationLevel, _ := trustPolicy.SignatureVerification.GetVerificationLevel()
+
+	outcome := &notation.VerificationOutcome{
```
rgnote, line 68: Can we return here if verification level is skip?

rgnote (commented):
Rest looks good to me

verifier/verifier.go
patrickzheng200: Nice catch!

verifier/verifier.go
patrickzheng200: Code updated. verifier package now incorporates the refactored plugin and plugin/proto packages. 
Since PR #199 by @JeyJeyGao introduces the refactored PluginMock in internal/mocks.go, tests in the verifier package are utilizing it directly. After PR #199 is merged, this PR should rebase and work smoothly with the PluginMock.

patrickzheng200:
> In `verifier_helpers.go`, not all receiver functions (i.e. methods associated with structs) are indeed receiver functions. We can make it a "static function".
> 
> Meanwhile, I suggest moving all receiver functions of the `verifier` struct to `verifier.go` and all others to `helper.go`, and cleans up `verifier_helpers.go`.

Done.

notation.go
```diff
@@ -215,15 +215,16 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, op
...
-	if len(verificationOutcomes) == 0 {
+	if numOfSignatureProcessed == 0 {
 		return ocispec.Descriptor{}, nil, ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("no signature is associated with %q, make sure the image was signed successfully", artifactRef)}
 	}
 
-	// check whether verification was successful or not
-	if verificationOutcomes[len(verificationOutcomes)-1].Error != nil {
+	// Verification Failed
+	if len(verificationOutcomes) == 0 {
 		return ocispec.Descriptor{}, verificationOutcomes, ErrorVerificationFailed{}
```
rgnote, line 224: Looks like we don't return any verification outcomes if there is no "good" signature associated with the given reference. That means, from notation cli, a user gets zero information on why the verification failed. All they see is `signature verification failed` message, which is very generic. They don't see the exact error messages associated with the "bad" signatures in order to understand the nature of the failure and potentially update their trust store or trust policy.

rgnote: @priteshbandi @shizhMSFT
This is the result of not processing and returning the outcomes of *all* signatures.

patrickzheng200: > Looks like we don't return any verification outcomes if there is no "good" signature associated with the given reference. That means, from notation cli, a user gets zero information on why the verification failed. All they see is `signature verification failed` message, which is very generic. They don't see the exact error messages associated with the "bad" signatures in order to understand the nature of the failure and potentially update their trust store or trust policy.

Just propose another option here:
1. Verification succeeded: return successful outcomes (there might be more than one signature succeeded) only.
2. Verification failed: still return all outcomes with details (all of them are failed).

@shizhMSFT @priteshbandi What do you think?

yizha1: My two cents:
- Limit the number of signatures attempted, for example, 50 (need to track this with an issue for later improvements)
- If one signature is verified, then the corresponding outcome should be recorded and returned. Library doesn't make any decision on what should be returned or not. The library returns the "fact". The debug option will benefit from this.

An example:
- If verification succeeded on the first signature, then overall verify succeeded, with debug enabled, only one successful signature info is printed out.
- if verification succeeded on the 4th signature, then overall verify succeeded, with debug enabled, the first 3 signatures with failed reason and the 4th successful signature are printed out
- if verification failed on all 50 signatures, then overall verify failed, with debug enabled, all the signatures with failed reasons are printed out. We also need to indicate the max number of 50 signatures are verified.

verifier/verifier.go
```diff
@@ -0,0 +1,507 @@
...
+				Type:   trustpolicy.TypeAuthenticity,
+				Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeAuthenticity],
+			}
+		}
+	} else {
+		return &notation.ValidationResult{
+			Type:   trustpolicy.TypeAuthenticity,
+			Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeAuthenticity],
+		}
+	}
```
shizhMSFT, line 372: nit:
```suggestion
	}

	return &notation.ValidationResult{
		Type:   trustpolicy.TypeAuthenticity,
		Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeAuthenticity],
	}
```

verifier/verifier.go
```diff
@@ -0,0 +1,507 @@
...
+			Error:  fmt.Errorf("digital signature has expired on %q", expiry.Format(time.RFC1123Z)),
+			Type:   trustpolicy.TypeExpiry,
+			Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeExpiry],
+		}
+	} else {
+		return &notation.ValidationResult{
+			Type:   trustpolicy.TypeExpiry,
+			Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeExpiry],
+		}
+	}
```
shizhMSFT, line 387: ni:
```suggestion
	}
	
	return &notation.ValidationResult{
		Type:   trustpolicy.TypeExpiry,
		Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeExpiry],
	}
```

verifier/verifier.go
```diff
@@ -0,0 +1,507 @@
...
+			Error:  err,
+			Type:   trustpolicy.TypeAuthenticTimestamp,
+			Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeAuthenticTimestamp],
+		}
+	} else {
+		return &notation.ValidationResult{
+			Type:   trustpolicy.TypeAuthenticTimestamp,
+			Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeAuthenticTimestamp],
+		}
+	}
```
shizhMSFT, line 437: nit:
```suggestion
	}

	return &notation.ValidationResult{
		Type:   trustpolicy.TypeAuthenticTimestamp,
		Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeAuthenticTimestamp],
	}
```

verifier/verifier.go
```diff
@@ -0,0 +1,507 @@
...
+	installedPlugin, err := v.pluginManager.Get(ctx, verificationPluginName)
+	if err != nil {
+		return nil, notation.ErrorVerificationInconclusive{Msg: fmt.Sprintf("error while locating the verification plugin %q, make sure the plugin is installed successfully before verifying the signature. error: %s", verificationPluginName, err)}
+	}
+	resp, err := installedPlugin.VerifySignature(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
```
shizhMSFT, line 506: nit:
```suggestion
	return installedPlugin.VerifySignature(ctx, req)
```

verifier/verifier.go
```diff
@@ -0,0 +1,507 @@
...
+		SignatureVerification: capabilitiesToVerify,
+	}
+
+	req := &proto.VerifySignatureRequest{
+		ContractVersion: proto.ContractVersion,
+		Signature:       signature,
+		TrustPolicy:     policy,
+		PluginConfig:    pluginConfig,
+	}
+	installedPlugin, err := v.pluginManager.Get(ctx, verificationPluginName)
```
shizhMSFT, line 497: It is redundant to get plugin again as we have already got it once before.

shizhMSFT (changes requested):

notation.go
patrickzheng200: > Let's keep it as it is for now. We can have a separate PR to resolve #201.

To summarize, keep it for now means we will have the following implementation:
1. Verification succeeded: the success outcome will be returned only. The verification process will stop at the first signature that gets verified successfully.
2. Verification failed: an empty verification outcome slice will be returned along with error `ErrorVerificationFailed`.

verifier/trustpolicy/trustpolicy_test.go
```diff
@@ -489,3 +494,41 @@ func TestApplicableTrustPolicy(t *testing.T) {
 		t.Fatalf("getApplicableTrustPolicy should return wildcard policy for registry scope \"some.registry.that/has.no.policy\"")
 	}
 }
+
+func TestLoadPolicyDocument(t *testing.T) {
```
shizhMSFT, line 498: nit:
```suggestion
func TestLoadDocument(t *testing.T) {
```

verifier/helpers.go
```diff
@@ -0,0 +1,183 @@
...
+	return certificates, nil
+}
+
+// isCriticalFailure checks whether a VerificationResult fails the entire signature verification workflow.
+// signature verification workflow is considered failed if there is a VerificationResult with "Enforced" as the action but the result was unsuccessful
+func isCriticalFailure(result *notation.ValidationResult) bool {
+	return result.Action == trustpolicy.ActionEnforce && result.Error != nil
+}
+
+func verifyX509TrustedIdentitiesCore(certs []*x509.Certificate, trustPolicy *trustpolicy.TrustPolicy) error {
```
shizhMSFT, line 85: This function `verifyX509TrustedIdentitiesCore` should be renamed to `verifyX509TrustedIdentities` and move to `verifier.go`.

The previous function `func verifyX509TrustedIdentities(trustPolicy *trustpolicy.TrustPolicy, outcome *notation.VerificationOutcome, authenticityResult *notation.ValidationResult)` should be removed.

shizhMSFT: Then line 150
```go
verifyX509TrustedIdentities(trustPolicy, outcome, authenticityResult)
```
becomes
```go
authenticityResult.Error = verifyX509TrustedIdentities(outcome.EnvelopeContent.SignerInfo.CertificateChain, trustPolicy)
```

verifier/verifier.go
```diff
@@ -0,0 +1,494 @@
...
+			if isCriticalFailure(revocationResult) {
+				return revocationResult.Error
+			}
+		}
+	}
+
+	return nil
+}
+
+func metadata(ctx context.Context, installedPlugin plugin.Plugin, pluginConfig map[string]string) (*proto.GetMetadataResponse, error) {
```
shizhMSFT, line 262: Suggest renaming function name to `getPluginMetadata` and move this function to `helpers.go` since it is not directly related to verify.

verifier/verifier.go
```diff
@@ -0,0 +1,494 @@
...
+		CertificateChain:      certChain,
+	}
+
+	policy := proto.TrustPolicy{
+		TrustedIdentities:     trustPolicy.TrustedIdentities,
+		SignatureVerification: capabilitiesToVerify,
+	}
+
+	req := &proto.VerifySignatureRequest{
+		ContractVersion: proto.ContractVersion,
```
shizhMSFT, line 487: @JeyJeyGao Should `ContractVersion` always be set by the package `plugin` instead of the caller?

verifier/verifier.go
```diff
@@ -0,0 +1,494 @@
...
+
+	// check if we need to verify using a plugin
+	var pluginCapabilities []proto.Capability
+	verificationPluginName, err := getVerificationPlugin(&outcome.EnvelopeContent.SignerInfo)
+	// use plugin, but getPluginName returns an error
+	if err != nil && err != errExtendedAttributeNotExist {
+		return err
+	}
+	var installedPlugin plugin.Plugin
+	if err == nil {
```
shizhMSFT, line 112: The logic will be more readable if 
```suggestion
	if verificationPluginName != "" {
```

verifier/verifier.go
```diff
@@ -0,0 +1,494 @@
...
+	// verify revocation
+	// check if we need to bypass the revocation check, since revocation can be skipped using a trust policy or a plugin may override the check
+	if outcome.VerificationLevel.Enforcement[trustpolicy.TypeRevocation] != trustpolicy.ActionSkip &&
+		!proto.CapabilityRevocationCheckVerifier.In(pluginCapabilities) {
+		// TODO perform X509 revocation check (not in RC1)
+		// https://github.com/notaryproject/notation-go/issues/110
+	}
+
+	// perform extended verification using verification plugin if present
+	if verificationPluginName != "" {
```
shizhMSFT, line 179: nit:
```suggestion
	if installedPlugin != nil {
```

verifier/verifier.go
```diff
@@ -0,0 +1,494 @@
...
+
+	// verify x509 trust store based authenticity
+	authenticityResult := verifyAuthenticity(ctx, trustPolicy, outcome)
+	outcome.VerificationResults = append(outcome.VerificationResults, authenticityResult)
+	if isCriticalFailure(authenticityResult) {
+		return authenticityResult.Error
+	}
+
+	// verify x509 trusted identity based authenticity (only if notation needs to perform this verification rather than a plugin)
+	if !proto.CapabilityTrustedIdentityVerifier.In(pluginCapabilities) {
```
shizhMSFT, line 149: It is a bit weird to write code like `!element.In(set)`. Usually, we have
```go
if !slice.Contains(pluginCapabilities, proto.CapabilityTrustedIdentityVerifier) {
```
which can be easily translated to natural language

> If plugin is NOT capable of verifying trusted identity, then ...

shizhMSFT: If we do rewrite the form of `!element.In(set)` to `!slice.Contains()`, we can remove `func (c Capability) In(capabilities []Capability) bool` in `plugin/proto/proto.go`.

internal/slice/slice.go
```diff
@@ -9,3 +9,13 @@ func Contains[E comparable](s []E, v E) bool {
 	}
 	return false
 }
+
+// ContainsAny reports whether v is present in s
+func ContainsAny(s []any, v any) bool {
```
shizhMSFT, line 14: nit: should we name package `slice` to package `slices` so that we can have consistent naming as https://pkg.go.dev/golang.org/x/exp/slices?

verifier/verifier.go
patrickzheng200: Good idea. Will do.

verifier/verifier.go
JeyJeyGao: Yes, it is. I'll update it in the next PR.

verifier/helpers.go
patrickzheng200: > Then line 150
> 
> ```go
> verifyX509TrustedIdentities(trustPolicy, outcome, authenticityResult)
> ```
> 
> becomes
> 
> ```go
> authenticityResult.Error = verifyX509TrustedIdentities(outcome.EnvelopeContent.SignerInfo.CertificateChain, trustPolicy)
> ```

Line 150 actually needs to be:
```
err = verifyX509TrustedIdentities(outcome.EnvelopeContent.SignerInfo.CertificateChain, trustPolicy)
if err != nil {
	authenticityResult.Error = err
}
```
because authenticityResult.Error can be non-nil before line 150, and if the verification level is not enforce, the code will still reach to line 150. If line 150 returns nil error, this nil error would overwrite the previous non-nil error.

shizhMSFT (approved):
LGTM

yizha1:
As per discussion and aligned during community call, this PR can be merged with two approvals from one org. The leftover issue is tracked here https://github.com/notaryproject/notation-go/issues/201.

JeyJeyGao (approved):
LGTM

PR #188, opened 2022-11-08
Fix VerifySignature command json marshaling

rgnote:
Temporary fix for https://github.com/notaryproject/notation-core-go/issues/88

### DIFF
--- a/internal/mock/mocks.go
+++ b/internal/mock/mocks.go
@@ -3,6 +3,7 @@ package mock
 import (
 	"context"
 	_ "embed"
+
 	"github.com/notaryproject/notation-core-go/signature"
 	"github.com/notaryproject/notation-go"
 	"github.com/notaryproject/notation-go/plugin"
--- a/internal/mock/mocks.go
+++ b/internal/mock/mocks.go
@@ -3,7 +3,6 @@ package mock
 import (
 	"context"
 	_ "embed"
-
 	"github.com/notaryproject/notation-core-go/signature"
 	"github.com/notaryproject/notation-go"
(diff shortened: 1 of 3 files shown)


verification/verifier_helpers.go
```diff
@@ -253,12 +253,12 @@ func (v *Verifier) executePlugin(ctx context.Context, trustPolicy *TrustPolicy,
...
 	}
-	var attributesToProcess []interface{}
-	extendedAttributes := make(map[interface{}]interface{})
+	var attributesToProcess []string
+	extendedAttributes := make(map[string]interface{})
 
 	for _, attr := range getNonPluginExtendedCriticalAttributes(signerInfo) {
-		extendedAttributes[attr.Key] = attr.Value
-		attributesToProcess = append(attributesToProcess, attr.Key)
+		extendedAttributes[attr.Key.(string)] = attr.Value
```
patrickzheng200, line 260: This line of code would panic if attr.Key is not of type string. Type assertion is needed here.

internal/mock/mocks.go
```diff
@@ -3,7 +3,6 @@ package mock
 import (
 	"context"
 	_ "embed"
-
```
shizhMSFT, line 6: It is better to keep those import block seperated.

verification/verifier_helpers.go
rgnote: This is handled in `getNonPluginExtendedCriticalAttributes`. It is returning only the string attributes

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

PR #190, opened 2022-11-08
update: Package registry refactoring

patrickzheng200:
This PR refactors package registry. Please only review files inside registry folder.
The original registry package is moved into the internal folder. Once this PR is merged, internal/registry folder shall be removed in a separate PR.

Signed-off-by: Patrick Zheng <patrickzheng@microsoft.com>

PR #191, opened 2022-11-08
update: Package notation refactoring

patrickzheng200:
PR review order: #190 -> #192 -> This PR -> #186.

This PR refactors the notation package, files intended to be reviewed in this PR:
1. files in notation folder
2. internal/mock

internal/common, verification/trustpolicy and verification/truststore pakcages shall be reviewed in #192 already. Therefore, they can be skipped in this PR.

Files in internal/mock_origin, internal/registry, and internal/notation.go are there just for not breaking the old dependencies. They will be removed once the refactoring is done.


notation/notation.go
```diff
@@ -0,0 +1,298 @@
...
+	"github.com/notaryproject/notation-go/registry"
+	"github.com/notaryproject/notation-go/verification/trustpolicy"
+	"github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+const AnnotationX509ChainThumbprint = "io.cncf.notary.x509chain.thumbprint#S256"
+
+// Descriptor describes the artifact that needs to be signed.
+type Descriptor struct {
```
JeyJeyGao, line 24: Do we still need it?

priteshbandi: +1 can we use ocispec.Descriptor ?

notation/notation.go
```diff
@@ -0,0 +1,314 @@
...
+	// present.
+	TSA timestamp.Timestamper
+
+	// TSAVerifyOptions is the verify option to verify the fetched timestamp
+	// signature.
+	// The `Intermediates` in the verify options will be ignored and
+	// re-contrusted using the certificates in the fetched timestamp signature.
+	// An empty list of `KeyUsages` in the verify options implies
+	// ExtKeyUsageTimeStamping.
+	TSAVerifyOptions x509.VerifyOptions
```
priteshbandi, line 67: Do we need to take this as input? Timestamps are as per RFC 3161 so we know what needs to be verified.

https://github.com/notaryproject/notaryproject/blob/main/specs/signature-specification.md#unsigned-attributes

notation/notation.go
```diff
@@ -0,0 +1,314 @@
...
+	if err != nil {
+		return Descriptor{}, nil, ErrorNoApplicableTrustPolicy{Msg: err.Error()}
+	}
+	// ignore the error since we already validated the policy document
+	verificationLevel, _ := trustpolicy.GetVerificationLevel(trustPolicy.SignatureVerification)
+	if verificationLevel.Name == trustpolicy.LevelSkip.Name {
+		verificationOutcomes = append(verificationOutcomes, &VerificationOutcome{VerificationLevel: verificationLevel})
+		return Descriptor{}, verificationOutcomes, nil
+	}
+	opts.VerificationLevel = verificationLevel
```
priteshbandi, line 226: Why are we modifying input?

notation/notation.go
```diff
@@ -0,0 +1,314 @@
...
+			sigBlob, sigBlobDesc, err := repo.FetchSignatureBlob(ctx, sigManifest)
+			if err != nil {
+				return ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("unable to retrieve digital signature with digest %q associated with %q from the registry, error : %s", sigBlobDesc.Digest, artifactRef, err.Error())}
+			}
+			_, outcome, err := verifier.Verify(ctx, sigBlob, opts)
+			if err != nil {
+				if outcome != nil && outcome.Error != nil {
+					outcome.SignatureBlobDescriptor = &sigBlobDesc
+					verificationOutcomes = append(verificationOutcomes, outcome)
+				}
```
priteshbandi, line 251: if error is `returned` we shouldn't rely on `outcome` to append outcome to verificationOutcomes

notation/notation.go
```diff
@@ -0,0 +1,314 @@
...
+				continue
+			}
+			outcome.SignatureBlobDescriptor = &sigBlobDesc
+			verificationOutcomes = append(verificationOutcomes, outcome)
```
priteshbandi, line 255: Should this be done irrespective if `Verify` returned error or not?

notation/notation.go
```diff
@@ -0,0 +1,314 @@
...
+	var success bool
+	var verifiedSigBlobDesc ocispec.Descriptor
+	err = repo.ListSignatures(ctx, artifactDescriptor, func(signatureManifests []ocispec.Descriptor) error {
+		if len(signatureManifests) < 1 {
+			return ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("no signatures are associated with %q, make sure the image was signed successfully", artifactRef)}
+		}
+		// if already verified successfully, no need to continue
+		if success {
+			return nil
+		}
```
priteshbandi, line 238: If success is set to true do we need to perform `len(signatureManifests) < 1` check?

notation/errors.go
```diff
@@ -0,0 +1,49 @@
+package notation
+
+// ErrorVerificationInconclusive is used when signature verification fails due to a runtime error (e.g. a network error)
+type ErrorVerificationInconclusive struct {
```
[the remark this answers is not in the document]
patrickzheng200: It's here because both notation package and verifier package use it. verifier package also imports notation package. If move errors.go to verifier package, there is import cycle.

notation/errors.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (e ErrorVerificationInconclusive) Error() string {
+	if e.Msg != "" {
+		return e.Msg
+	}
+	return "signature verification was inclusive due to an unexpected error"
+}
+
+// ErrorNoApplicableTrustPolicy is used when there is no trust policy that applies to the given artifact
+type ErrorNoApplicableTrustPolicy struct {
```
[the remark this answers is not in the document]
patrickzheng200: Same as above.

notation/errors.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (e ErrorNoApplicableTrustPolicy) Error() string {
+	if e.Msg != "" {
+		return e.Msg
+	}
+	return "there is no applicable trust policy for the given artifact"
+}
+
+// ErrorSignatureRetrievalFailed is used when notation is unable to retrieve the digital signature/s for the given artifact
+type ErrorSignatureRetrievalFailed struct {
```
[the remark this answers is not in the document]
patrickzheng200: Same as above.

notation/errors.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (e ErrorSignatureRetrievalFailed) Error() string {
+	if e.Msg != "" {
+		return e.Msg
+	}
+	return "unable to retrieve the digital signature from the registry"
+}
+
+// ErrorVerificationFailed is used when it is determined that the digital signature/s is not valid for the given artifact
+type ErrorVerificationFailed struct {
```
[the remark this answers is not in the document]
patrickzheng200: Same as above.

notation/notation.go
shizhMSFT: We can remove this option as it will be succeeded by the trust store and policy in the future.

notation/notation.go
```diff
@@ -0,0 +1,314 @@
...
+ 5. Digest in the signature does not match the digest present in the reference
+
+If each and every signature associated with the reference fail the verification,
+then Verify will return `ErrorVerificationFailed` error along with an array
+of `VerificationOutcome`.
+
+# Callers can pass the verification plugin config in VerifyOptions.PluginConfig
+
+For more details on signature verification, see https://github.com/notaryproject/notaryproject/blob/main/trust-store-trust-policy-specification.md#signature-verification
+*/
```
shizhMSFT, line 174: Can we change the style of comment block?

notation/notation.go
```diff
@@ -0,0 +1,314 @@
...
+
+	// VerificationLevel encapsulates the signature verification preset and its
+	// actions for each verification type
+	VerificationLevel *trustpolicy.VerificationLevel
+}
+
+// ValidationResult encapsulates the verification result (passed or failed)
+// for a verification type, including the desired verification action as
+//
+//	specified in the trust policy
```
shizhMSFT, line 99: This comment block seems not well formatted?

notation/notation.go
```diff
@@ -0,0 +1,314 @@
...
+type ValidationResult struct {
+	// Success is set to true if the verification was successful
+	Success bool
```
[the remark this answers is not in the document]
shizhMSFT: We should drop `Success` and use `Error` as the success indicator.

notation/notation.go
```diff
@@ -0,0 +1,314 @@
...
+}
+
+// Verifier is a generic interface for verifying an artifact.
+type Verifier interface {
+	// Verify verifies the signature blob and returns the signature blob
+	// descriptor upon successful verification.
+	Verify(ctx context.Context, signature []byte, opts VerifyOptions) (Descriptor, *VerificationOutcome, error)
+
+	// TrustPolicyDocument gets the validated trust policy document.
+	TrustPolicyDocument() (*trustpolicy.Document, error)
```
shizhMSFT, line 146: This method should not be a part of the `Verifier` interface and `*trustpolicy.Document` should be only in the `verifier` package and not here in the `notation` package.

notation/notation.go
```diff
@@ -0,0 +1,314 @@
...
+	VerificationLevel *trustpolicy.VerificationLevel
+	// VerificationResults contains the verifications performed on the signature
+	// and their results
+	VerificationResults []*ValidationResult
+	// SignedAnnotations contains arbitrary metadata relating to the target
+	// artifact that was signed
+	SignedAnnotations map[string]string
+	// Error that caused the verification to fail (if it fails)
+	Error error
+}
```
shizhMSFT: Those structures should be in the `verifier` package.

notation/notation.go
patrickzheng200: Changed to 
```
if err != nil {
    if outcome != nil {
        outcome.SignatureBlobDescriptor = &sigBlobDesc
	outcome.Error = err
	verificationOutcomes = append(verificationOutcomes, outcome)
    }
    continue
}
```
The append here is saving those signatures that failed the verification with details.

notation/notation.go
```diff
@@ -0,0 +1,314 @@
...
+	// Currently both `application/jose+json` and `application/cose` are
+	// supported.
+	SignatureMediaType string
+
+	// PluginConfig is a map of plugin configs.
+	PluginConfig map[string]string
+
+	// VerificationLevel encapsulates the signature verification preset and its
+	// actions for each verification type
+	VerificationLevel *trustpolicy.VerificationLevel
```
shizhMSFT: This option is not there in the design.

notation/notation.go
patrickzheng200: At this point, the signature verification has succeeded. This append is saving the signature with details.
For signatures that failed the verification, their outcomes are saved in the above step.

notation/notation.go
patrickzheng200: Nice catch! Moved `len(signatureManifests) < 1` after checking `success`.

notation/notation.go
patrickzheng200: Code updated.

notation/notation.go
patrickzheng200: Removed.

notation/notation.go
patrickzheng200: Both notation and verifier package are using these structures. verifier package also imports notation package. Putting these structures in the verifier package would cause import cycle.

shizhMSFT: We can discuss offline if there are technical issue like resolving an import cycle.

shizhMSFT: > BTW, what's reason would `verifier` package import package `notation`?

Oh, I see `verifier.New()` returns a `notation.Verifier`.

notation/notation.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		// process signatures
+		for _, sigManifest := range signatureManifests {
+			// get signature envelope
+			sigBlob, sigBlobDesc, err := repo.FetchSignatureBlob(ctx, sigManifest)
+			if err != nil {
+				return ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("unable to retrieve digital signature with digest %q associated with %q from the registry, error : %s", sigBlobDesc.Digest, artifactRef, err.Error())}
+			}
+			_, outcome, err := verifier.Verify(ctx, sigBlob, opts)
```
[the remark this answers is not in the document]
patrickzheng200: Do you mean sigBlobDesc? artifactDescriptor is the descriptor of the target artifact that got signed.

notation/errors.go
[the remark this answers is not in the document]
priteshbandi: Why does verify needs to depend on notation? can we break that circular dependency?

internal/mock/mocks.go
```diff
@@ -60,39 +65,40 @@ var (
...
 	return t.ResolveResponse, t.ResolveError
 }
 
-func (t Repository) ListSignatureManifests(ctx context.Context, manifestDigest digest.Digest) ([]registry.SignatureManifest, error) {
-	return t.ListSignatureManifestsResponse, t.ListSignatureManifestsError
+func (t Repository) ListSignatures(ctx context.Context, desc ocispec.Descriptor, fn func(signatureManifests []ocispec.Descriptor) error) error {
+	err := fn(t.ListSignaturesResponse)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
patrickzheng200: Yes. If `err == nil`, but we still want to mock ListSignaturesError, we will ` return t.ListSignaturesError` instead of nil. This is for unit test purpose.

notation/errors.go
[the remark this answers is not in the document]
patrickzheng200: That's because `verifier.New()` returns a `notation.Verifier` by design.

notation/notation.go
```diff
@@ -0,0 +1,285 @@
...
+			if err != nil {
+				outcome.Error = err
+				continue
+			}
+			outcome.SignedAnnotations = payload.TargetArtifact.Annotations
+			success = true
+			// Descriptor of the signature blob that gets verified successfully
+			verifiedSigBlobDesc = sigBlobDesc
+
+			return nil
```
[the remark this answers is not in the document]
patrickzheng200: > Can line 171-212 moved to verifier pkg? Basically have a `verifier.Verify` method similar to `signer.Sign`. For local verification we can have method ``verifier.LocalVerify`,

That's going to be a relatively big change of the verifier interface design. @shizhMSFT any suggestion/comment/concern on this one?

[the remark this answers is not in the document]
patrickzheng200: > Can line 171-212 moved to verifier pkg? Basically have a `verifier.Verify` method similar to `signer.Sign`. For local verification we can have method ``verifier.LocalVerify`,

Had a discussion with @shizhMSFT, we suggest to keep the current design. Verifying remote is based on verifying local. That's because when verifying remotely, we still need to Resolve and Fetch the remote signatures to local and then do the verification.

notation.go
```diff
@@ -63,31 +41,202 @@ type SignOptions struct {
...
+	}
+	err = repo.ListSignatures(ctx, artifactDescriptor, func(signatureManifests []ocispec.Descriptor) error {
+		// if already verified successfully, no need to continue
+		if success {
+			return nil
+		}
+
+		if len(signatureManifests) < 1 {
+			return ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("no signatures are associated with %q, make sure the image was signed successfully", artifactRef)}
+		}
```
shizhMSFT, line 171: `signatureManifests` is guaranteed to be non-zero length. If there is no signature manifes, `ListSignatures` will return nil without calling the callback function.

shizhMSFT: You may think other way to return this error.

notation.go
```diff
@@ -63,31 +41,202 @@ type SignOptions struct {
...
+
+		// process signatures
+		for _, sigManifest := range signatureManifests {
```
shizhMSFT, line 174: Renaming `sigManifest` to `sigManifestDesc` will be better to understand.

notation.go
```diff
@@ -63,31 +41,202 @@ type SignOptions struct {
...
+			// get signature envelope
+			sigBlob, sigBlobDesc, err := repo.FetchSignatureBlob(ctx, sigManifest)
+			if err != nil {
+				return ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("unable to retrieve digital signature with digest %q associated with %q from the registry, error : %s", sigBlobDesc.Digest, artifactRef, err.Error())}
+			}
+			payloadArtifactDescriptor, outcome, err := verifier.Verify(ctx, sigBlob, opts)
+			if err == nil && !equal(&payloadArtifactDescriptor, &artifactDescriptor) {
```
shizhMSFT, line 181: Since `equal` is deep equal, the `equal` will always fail if there's any annotation. 

Instead, we should check the MediaType, size, digest only and should not check the annotation. Try `content.Equal` from the `oras-go` package.

notation.go
```diff
@@ -63,31 +41,202 @@ type SignOptions struct {
...
+
+	if err != nil && !errors.Is(err, doneVerification) {
+		return ocispec.Descriptor{}, nil, err
+	}
+
+	// check whether verification was successful or not
+	if success {
+		// signature verification succeeds if there is at least one good
+		// signature
+		return verifiedSigBlobDesc, verificationOutcomes, nil
```
shizhMSFT, line 213: We should return `payloadArtifactDescriptor` on successful verification.

shizhMSFT (changes requested):

notation.go
```diff
@@ -63,31 +41,202 @@ type SignOptions struct {
...
+	// Error is set if there are any errors during the verification process
+	Error error
+}
+
+// VerificationOutcome encapsulates a signature blob's descriptor, its content,
+// the verification level and results for each verification type that was
+// performed.
+type VerificationOutcome struct {
+	// SignatureBlobDescriptor is descriptor of the signature envelope blob
+	SignatureBlobDescriptor *ocispec.Descriptor
```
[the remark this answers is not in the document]
patrickzheng200: This is the field that contains the signature blob descriptor with the signature's digest. Notation CLI needs it during printing when multiple signatures verification failed.

notation.go
```diff
@@ -63,31 +41,202 @@ type SignOptions struct {
...
+	}
+	verificationLevel := outcome.VerificationLevel
+	if verificationLevel.Name == trustpolicy.LevelSkip.Name {
+		verificationOutcomes = append(verificationOutcomes, &VerificationOutcome{VerificationLevel: verificationLevel})
+		return ocispec.Descriptor{}, verificationOutcomes, nil
+	}
+
+	// get signature manifests
+	var success bool
+	doneVerification := errors.New("done verification")
```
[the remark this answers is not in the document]
patrickzheng200: `const doneVerification = errors.New("done verification")` gives me an error: `errors.New("done verification") (value of type error) is not constant`. Maybe make it a var?

notation.go
```diff
@@ -63,31 +41,202 @@ type SignOptions struct {
...
+		return ocispec.Descriptor{}, nil, err
+	}
+	verificationLevel := outcome.VerificationLevel
+	if verificationLevel.Name == trustpolicy.LevelSkip.Name {
+		verificationOutcomes = append(verificationOutcomes, &VerificationOutcome{VerificationLevel: verificationLevel})
+		return ocispec.Descriptor{}, verificationOutcomes, nil
+	}
+
+	// get signature manifests
+	var success bool
```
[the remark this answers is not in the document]
patrickzheng200: It is still needed, the final step of Verify() function checks if success is true. If it's true we will `return targetArtifactDesc, verificationOutcomes, nil`.

notation.go
```diff
@@ -63,31 +42,186 @@ type SignOptions struct {
...
+			sigBlob, sigBlobDesc, err := repo.FetchSignatureBlob(ctx, sigManifestDesc)
+			if err != nil {
+				return ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("unable to retrieve digital signature with digest %q associated with %q from the registry, error : %s", sigBlobDesc.Digest, artifactRef, err.Error())}
+			}
+			payloadArtifactDescriptor, outcome, err := verifier.Verify(ctx, sigBlob, opts)
+			if err == nil && !content.Equal(payloadArtifactDescriptor, artifactDescriptor) {
+				err = errors.New("payloadArtifactDescriptor does not match artifactDescriptor")
+			}
+			if err != nil {
+				if outcome != nil {
```
priteshbandi, line 174: Is there a case where err is not nil but outcome is nil? If so then we might miss a value in `verificationOutcomes`

notation.go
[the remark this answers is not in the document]
priteshbandi: Why do we need to return targetArtifactDesc? The `VerificationOutcome` contains `EnvelopeContent` which will have descriptor that is in payload.

https://github.com/patrickzheng200/notation-go/blob/af27f20064cbc623058cd201102dfbbc8b74837a/notation.go#L112

notation.go
patrickzheng200: At this point, no. This logic is guaranteed in the implementation of verifier.Verify().

notation.go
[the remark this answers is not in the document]
patrickzheng200: The Payload.Content is a []byte instead of a descriptor.

notation.go
priteshbandi: yes but can there be regression in future? Should we emit some logs when it happens?

notation.go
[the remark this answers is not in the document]
priteshbandi: We will do `json.Unmarshal(outcome.EnvelopeContent.Payload.Content, payload)` only when there is successful verification and IMO its okay to do some repetitive work in favor of readability.

notation.go
patrickzheng200: How about after I add the log package into notation-go and we will use the logger to log it? I will add a TODO to track this for now.

notation.go
```diff
@@ -63,31 +42,188 @@ type SignOptions struct {
...
+					outcome.SignatureBlobDescriptor = &sigBlobDesc
+					outcome.Error = err
+					verificationOutcomes = append(verificationOutcomes, outcome)
+				}
+				// TODO: add log here to track when err != nil,
+				// but outcome == nil
+				continue
+			}
```
shizhMSFT, line 182: Always check error and then process. Otherwise, it is very easy to write buggy code.

shizhMSFT:
```suggestion
			if err != nil {
				if outcome == nil {
					// TODO: log fatal error
					return err
				}
				verificationOutcomes = append(verificationOutcomes, outcome)
				continue
			}
			if !content.Equal(payloadArtifactDescriptor, artifactDescriptor) {
				outcome.Error = errors.New("content descriptor mismatch")
				verificationOutcomes = append(verificationOutcomes, outcome)
				continue
			}
```

notation.go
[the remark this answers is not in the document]
shizhMSFT: In the most scenarios, only the descriptor of the signed artifact is interested. The `Verify()` function will be invoked as
```golang
desc, _, err := Verify(...)
if err != nil {
    return err
}
return downloadAndProcess(desc)
```
This design is to make the notation-go SDK more user friendly and easier to use.

notation.go
```diff
@@ -63,31 +42,188 @@ type SignOptions struct {
...
+	// will be returned.
+	Verify(ctx context.Context, signature []byte, opts VerifyOptions) (ocispec.Descriptor, *VerificationOutcome, error)
+}
+
+// Verify performs signature verification on each of the notation supported
+// verification types (like integrity, authenticity, etc.) and return the
+// verification outcomes.
+// For more details on signature verification, see
+// https://github.com/notaryproject/notaryproject/blob/main/specs/trust-store-trust-policy.md#signature-verification
+func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, opts VerifyOptions) (ocispec.Descriptor, []*VerificationOutcome, error) {
```
shizhMSFT, line 140: @yizha1 @patrickzheng200 @priteshbandi @rgnote I have security concerns on
```golang
func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, opts VerifyOptions) (ocispec.Descriptor, []*VerificationOutcome, error) {
```
returning `[]*VerificationOutcome`.

What if the remote registry is malicious and returns millions of invalid signature manifests? It will create millions of objects in `[]*VerificationOutcome`. Eventually, the process may be out of memory and get killed. It affects the availability if notation-go is integrated with a server application.

shizhMSFT: Several options here.
1. Remove `[]*VerificationOutcome` from the list of return parameters.
3. We set a limit on the number of attempted signatures.

priteshbandi: I think the underlying concern is: What's the maximum number of signatures we will evaluate (assuming all failed verification) before declaring overall verification as failed?
For the short term, we can put some reasonable limit(TBD - 50?) on the maximum number of signatures that the notation will evaluate. If an image contains more than that, the notation will fail verification. The catch here is that the order of signature artifact manifest descriptors returned by refers API should be deterministic.

notation.go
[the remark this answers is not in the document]
priteshbandi: Instead of a descriptor should we return a signature envelope that would contain the descriptor and other data?
The usecase where a signature envelope would be useful is returning the user-defined signed metadata as part of verification.

notation.go
patrickzheng200: Added a limit number 50 for now.

yizha1: @priteshbandi per your suggestion, could you create an issue to track the configuration of "maximum number of signatures"?

notation.go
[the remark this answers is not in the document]
patrickzheng200: > Instead of a descriptor should we return a signature envelope that would contain the descriptor and other data? The usecase where a signature envelope would be useful is returning the user-defined signed metadata as part of verification.

Oh yeah, for the signature envelope blob, I added a field called `RawSignature` inside VerificationOutcome. It's also included in this PR.

notation.go
```diff
@@ -63,31 +44,189 @@ type SignOptions struct {
...
+	} else if outcome.VerificationLevel.Name == trustpolicy.LevelSkip.Name {
+		return ocispec.Descriptor{}, []*VerificationOutcome{outcome}, nil
+	}
+
+	// get signature manifests
+	var targetArtifactDesc ocispec.Descriptor
+	artifactDescriptor, err := repo.Resolve(ctx, artifactRef)
+	if err != nil {
+		return ocispec.Descriptor{}, nil, ErrorSignatureRetrievalFailed{Msg: err.Error()}
+	}
```
shizhMSFT, line 160: nit: code clean up
```suggestion
	// passing nil signature to check 'skip'
	_, outcome, err := verifier.Verify(ctx, nil, opts)
	if err != nil {
		if outcome == nil {
			return ocispec.Descriptor{}, nil, err
		}
	} else if outcome.VerificationLevel.Name == trustpolicy.LevelSkip.Name {
		return ocispec.Descriptor{}, []*VerificationOutcome{outcome}, nil
	}

	// get signature manifests
	artifactRef := opts.ArtifactReference
	artifactDescriptor, err := repo.Resolve(ctx, artifactRef)
	if err != nil {
		return ocispec.Descriptor{}, nil, ErrorSignatureRetrievalFailed{Msg: err.Error()}
	}
	
	var verificationOutcomes []*VerificationOutcome
	var targetArtifactDesc ocispec.Descriptor
```

notation.go
```diff
@@ -63,31 +44,189 @@ type SignOptions struct {
...
+func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, opts VerifyOptions) (ocispec.Descriptor, []*VerificationOutcome, error) {
+	var verificationOutcomes []*VerificationOutcome
+	artifactRef := opts.ArtifactReference
+	// passing nil signature to check 'skip'
+	_, outcome, err := verifier.Verify(ctx, nil, opts)
+	if err != nil {
+		if outcome == nil {
+			return ocispec.Descriptor{}, nil, err
+		}
+	} else if outcome.VerificationLevel.Name == trustpolicy.LevelSkip.Name {
```
shizhMSFT, line 151: Should we compare the level directly instead of comparing the name?
```suggestion
	} else if outcome.VerificationLevel == trustpolicy.LevelSkip {
```

notation.go
shizhMSFT: Basically, only create variables when needed.

notation.go
```diff
@@ -63,31 +44,189 @@ type SignOptions struct {
...
+		return ocispec.Descriptor{}, []*VerificationOutcome{outcome}, nil
+	}
+
+	// get signature manifests
+	var targetArtifactDesc ocispec.Descriptor
+	artifactDescriptor, err := repo.Resolve(ctx, artifactRef)
+	if err != nil {
+		return ocispec.Descriptor{}, nil, ErrorSignatureRetrievalFailed{Msg: err.Error()}
+	}
+	count := 0
```
shizhMSFT, line 161: I am not able to observe `count++` anywhere.

notation.go
```diff
@@ -63,31 +44,189 @@ type SignOptions struct {
...
+	err = repo.ListSignatures(ctx, artifactDescriptor, func(signatureManifests []ocispec.Descriptor) error {
+		if count >= verificationSignatureNumLimit {
+			return fmt.Errorf("number of signatures attempted to be verified should be less than: %d", verificationSignatureNumLimit)
+		}
```
shizhMSFT, line 165: Should this check in the for loop?

notation.go
```diff
@@ -63,31 +44,189 @@ type SignOptions struct {
...
+				continue
+			}
+
+			// At this point, we've found a signature verified successfully
+			outcome.RawSignature = sigBlob
+			verificationOutcomes = append(verificationOutcomes, outcome)
+			// Descriptor of the signature blob that gets verified successfully
+			targetArtifactDesc = payloadArtifactDescriptor
+
+			return errDoneVerification
```
rgnote, line 194: Are we returning from verification on the first successful signature here? We wanted to process all signatures so that `notation inspect` command can show details of each signature. Similar to `signtool.exe` showing all signatures rather than just until a good signature.

notation.go
```diff
@@ -63,31 +44,189 @@ type SignOptions struct {
...
+				continue
+			}
+			if !content.Equal(payloadArtifactDescriptor, artifactDescriptor) {
+				outcome.Error = errors.New("content descriptor mismatch")
+				verificationOutcomes = append(verificationOutcomes, outcome)
+				continue
+			}
+
+			// At this point, we've found a signature verified successfully
+			outcome.RawSignature = sigBlob
```
shizhMSFT, line 189: We should remove this line since `verifier.Verify()` will set it.

notation.go
```diff
@@ -63,31 +44,189 @@ type SignOptions struct {
...
+
+	// get signature manifests
+	var targetArtifactDesc ocispec.Descriptor
+	artifactDescriptor, err := repo.Resolve(ctx, artifactRef)
+	if err != nil {
+		return ocispec.Descriptor{}, nil, ErrorSignatureRetrievalFailed{Msg: err.Error()}
+	}
+	count := 0
+	err = repo.ListSignatures(ctx, artifactDescriptor, func(signatureManifests []ocispec.Descriptor) error {
+		if count >= verificationSignatureNumLimit {
```
rgnote, line 163: Where are we incrementing this count?

notation.go
```diff
@@ -63,31 +44,189 @@ type SignOptions struct {
...
+	}
+
+	// check whether verification was successful or not
+	if verificationOutcomes[len(verificationOutcomes)-1].Error == nil {
+		// signature verification succeeds if there is at least one good
+		// signature
+		return targetArtifactDesc, verificationOutcomes, nil
+	}
+
+	return ocispec.Descriptor{}, verificationOutcomes, ErrorVerificationFailed{}
```
shizhMSFT, line 215: nit: code style check

Instead of doing
```go
err := do()
if err == nil {
    return process()
}
return err
```
try
```go
err := do()
if err != nil {
    return err
}
process()
```

```suggestion
	// check whether verification was successful or not
	if verificationOutcomes[len(verificationOutcomes)-1].Error != nil {
		return ocispec.Descriptor{}, verificationOutcomes, ErrorVerificationFailed{}
	}
	return targetArtifactDesc, verificationOutcomes, nil
```

notation.go
```diff
@@ -63,31 +44,189 @@ type SignOptions struct {
...
+	// get signature manifests
+	var targetArtifactDesc ocispec.Descriptor
+	artifactDescriptor, err := repo.Resolve(ctx, artifactRef)
+	if err != nil {
+		return ocispec.Descriptor{}, nil, ErrorSignatureRetrievalFailed{Msg: err.Error()}
+	}
+	count := 0
+	err = repo.ListSignatures(ctx, artifactDescriptor, func(signatureManifests []ocispec.Descriptor) error {
+		if count >= verificationSignatureNumLimit {
+			return fmt.Errorf("number of signatures attempted to be verified should be less than: %d", verificationSignatureNumLimit)
```
shizhMSFT, line 164: Should we wrap it into a `ErrorVerificationFailed`?

shizhMSFT (changes requested):

notation.go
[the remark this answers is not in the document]
shizhMSFT: Note that `RawSignature` is better than a signature envelope type since a bad signature cannot even be parsed.

notation.go
rgnote, line 164:
```suggestion
			return fmt.Errorf("number of signatures associated with an image should be less than: %d", verificationSignatureNumLimit)
```

notation.go
shizhMSFT: > Keep in mind that a specially crafted single signature may cause resource exhaustion too, for example, a signature with thousands of certificates in it.

That's covered since we have a size limit for a signature blob. Also, we have a size limit for signature manifest. Again, we also have a size limit for the return value of the referrers API as the underlying `oras-go` takes care of those security checks.

notation.go
patrickzheng200: This might not work as both of them are pointers to structs? I had to use reflect.DeepEqual to make it work.

notation.go
patrickzheng200: IMO, when we have one signature verified successfully, the whole verification is succeeded. In this case, general users would most likely only interest in the result or the succeeded signature. 
We will process all the signatures if none of them gets verified successfully. In this case, users might interest more in the failure details of the signatures.
Adding another option into VerifyOption seems a little bit over designed? It's making the system more complicated. @shizhMSFT any comments/suggestions/concerns on this one?

notation.go
```diff
@@ -63,31 +43,195 @@ type SignOptions struct {
...
+	// supported.
 	SignatureMediaType string
+
+	// PluginConfig is a map of plugin configs.
+	PluginConfig map[string]string
+
+	// MaxVerificationLimit is the maximum number of signature envelopes that
+	// can be associated with the target artifact. If not set by user, it will
+	// be set to 50 by default.
+	MaxVerificationLimit int
```
shizhMSFT, line 91: Since `VerifyOptions` is shared by `Verifier.Verify()` and `notation.Verify()`. Could you document that this option is scoped to `notation.Verify()`?

notation.go
```diff
@@ -63,31 +43,195 @@ type SignOptions struct {
...
+		// Set MaxVerificationLimit to 50 as default
+		opts.MaxVerificationLimit = 50
+	}
+	err = repo.ListSignatures(ctx, artifactDescriptor, func(signatureManifests []ocispec.Descriptor) error {
+		// process signatures
+		for _, sigManifestDesc := range signatureManifests {
+			if count >= opts.MaxVerificationLimit {
+				return ErrorVerificationFailed{Msg: fmt.Sprintf("number of signatures associated with an artifact should be less than: %d", opts.MaxVerificationLimit)}
+			}
+			count++
```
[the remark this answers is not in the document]
patrickzheng200: Nice catch!

shizhMSFT (approved):
LGTM

notation.go
rgnote: @shizhMSFT This is not out of scope for this PR. Before this PR, we are processing all signatures. The reason for processing all signatures is to support commands like notation inspect. With this refactoring, we are changing the functionality. I'm okay with that, but let's expose this as an option so that library consumers have a way.

shizhMSFT: Let's add an option so that invokers can choose to verify all signatures although we have a valid one.

shizhMSFT: @rgnote By going through the logic, I find there is an issue on the requirement of processing all signatures.

What if we have 100 signatures where the 42th one valid and others are invalid? Since it is over 50 signatures, the `Verify()` function will return error. Then the invoker will think the entire verification process failed although it in fact succeeded.

notation.go
```diff
@@ -63,31 +45,202 @@ type SignOptions struct {
...
+			// At this point, we've found a signature verified successfully
+			verificationOutcomes = append(verificationOutcomes, outcome)
+			// Descriptor of the signature blob that gets verified successfully
+			targetArtifactDesc = payloadArtifactDescriptor
+
+			return errDoneVerification
+		}
+
+		if count >= opts.MaxVerificationLimit {
+			return ErrorVerificationFailed{Msg: fmt.Sprintf("number of signatures associated with an artifact should be less than: %d", opts.MaxVerificationLimit)}
```
shizhMSFT, line 210: When it reaches the max verification limit should we return the previous outcomes along with the above error?

notation.go
```diff
@@ -63,31 +45,202 @@ type SignOptions struct {
...
+
+	// PluginConfig is a map of plugin configs.
+	PluginConfig map[string]string
+
+	// MaxVerificationLimit is the maximum number of signature envelopes that
+	// can be associated with the target artifact. If not set by user, it will
+	// be set to 50 by default.
+	// Note: this option is scoped to notation.Verify(). verifier.Verify() is
+	// for signle signature verification, and therefore, does not use it.
+	MaxVerificationLimit int
```
shizhMSFT, line 95: Will it be more accurate to say
```suggestion
	MaxSignatureAttempts int
```

shizhMSFT (approved):
LGTM

notation.go
```diff
@@ -63,31 +45,207 @@ type SignOptions struct {
...
+	var targetArtifactDesc ocispec.Descriptor
+	if opts.MaxSignatureAttempts <= 0 {
+		// Set MaxVerificationLimit to 50 as default
+		opts.MaxSignatureAttempts = maxVerificationLimitDefault
+	}
+	errExceededMaxVerificationLimit := ErrorVerificationFailed{Msg: fmt.Sprintf("number of signatures associated with an artifact should be less than: %d", opts.MaxSignatureAttempts)}
+	count := 0
+	err = repo.ListSignatures(ctx, artifactDescriptor, func(signatureManifests []ocispec.Descriptor) error {
+		// process signatures
+		for _, sigManifestDesc := range signatureManifests {
```
priteshbandi, line 177: Callout: we haven't implemented signature filtering https://github.com/notaryproject/notaryproject/blob/main/specs/signature-specification.md#signature-filtering

priteshbandi: Opened issue: https://github.com/notaryproject/notation-go/issues/197

notation.go
priteshbandi: Opened https://github.com/notaryproject/notation-go/issues/198 to track it.

priteshbandi (approved):
LGTM with comments addressed.

shizhMSFT (approved):
LGTM

notation.go
priteshbandi: had a discussion with @gokarnm on this and he suggested using timeout as the restricting mechanism which sounds more appt here.

We should also modify the `VerificationOutcome` behavior to return only successful signature verification outcome.

cc: @shizhMSFT @rgnote

shizhMSFT: @priteshbandi Can you create an issue for discussions and tracking?

shizhMSFT: I see it is tracked in #201

PR #192, opened 2022-11-08
feat: Added trustpolicy and truststore packages under verification

patrickzheng200:
PR review order: #190 -> This PR -> #191 -> #186

This PR adds the trustpolicy and truststore packages under verification.


verification/truststore/truststore.go
```diff
@@ -0,0 +1,126 @@
...
+type X509TrustStore interface {
+	// List named stores under storeType
+	// List(ctx context.Context, storeType Type) ([]string, error)
+
+	// GetCertificates returns certificates under storeType/namedStore
+	GetCertificates(ctx context.Context, storeType Type, namedStore string) ([]*x509.Certificate, error)
+}
+
+// NewX509TrustStore generates a new X509TrustStore
+func NewX509TrustStore(trustStorefs dir.SysFS) X509TrustStore {
```
JeyJeyGao, line 43: The constructor code is usually after the type struct.

internal/common/common.go
```diff
@@ -0,0 +1,97 @@
...
+
+	ldapv3 "github.com/go-ldap/ldap/v3"
+)
+
+const (
+	Wildcard    = "*"
+	X509Subject = "x509.subject"
+)
+
+// isPresent is a utility function to check if a string exists in an array
```
shizhMSFT, line 16:
```suggestion
// IsPresent is a utility function to check if a string exists in an array
```

internal/common/common.go
```diff
@@ -0,0 +1,97 @@
+package common
```
shizhMSFT, line 1: Try not to have a package named `common` or `utils`.

Consider
- `package slice` for `func Contains[E comparable](s []E, v E) bool` (originally `IsPresent`)
- `package pkix` for DN related methods
- `package registry` for `ValidateRegistryScopeFormat`

If internal methods are used within a package, then an unexported method will be better.

verification/trustpolicy/trustpolicy.go
```diff
@@ -0,0 +1,448 @@
...
+	TypeIntegrity          ValidationType = "integrity"
+	TypeAuthenticity       ValidationType = "authenticity"
+	TypeAuthenticTimestamp ValidationType = "authenticTimestamp"
+	TypeExpiry             ValidationType = "expiry"
+	TypeRevocation         ValidationType = "revocation"
+
+	ActionEnforce ValidationAction = "enforce"
+	ActionLog     ValidationAction = "log"
+	ActionSkip    ValidationAction = "skip"
+)
```
shizhMSFT, line 37: One `const` block for one type.
```suggestion
const (
	TypeIntegrity          ValidationType = "integrity"
	TypeAuthenticity       ValidationType = "authenticity"
	TypeAuthenticTimestamp ValidationType = "authenticTimestamp"
	TypeExpiry             ValidationType = "expiry"
	TypeRevocation         ValidationType = "revocation"
)

const (
	ActionEnforce ValidationAction = "enforce"
	ActionLog     ValidationAction = "log"
	ActionSkip    ValidationAction = "skip"
)
```

verification/trustpolicy/trustpolicy.go
```diff
@@ -0,0 +1,448 @@
...
+		Name: "skip",
+		Enforcement: map[ValidationType]ValidationAction{
+			TypeIntegrity:          ActionSkip,
+			TypeAuthenticity:       ActionSkip,
+			TypeAuthenticTimestamp: ActionSkip,
+			TypeExpiry:             ActionSkip,
+			TypeRevocation:         ActionSkip,
+		},
+	}
+
```
shizhMSFT, line 83: Same for `var` blocks.

verification/trustpolicy/trustpolicy.go
```diff
@@ -0,0 +1,448 @@
...
+	}
+)
+
+// Document represents a trustPolicy.json document
+type Document struct {
+	// Version of the policy document
+	Version string `json:"version"`
+	// TrustPolicies include each policy statement
+	TrustPolicies []TrustPolicy `json:"trustPolicies"`
+}
```
shizhMSFT, line 112: It will be better to have empty lines in between.

verification/trustpolicy/trustpolicy.go
```diff
@@ -0,0 +1,448 @@
...
+type SignatureVerification struct {
+	VerificationLevel string                              `json:"level"`
+	Override          map[ValidationType]ValidationAction `json:"override,omitempty"`
+}
+
+// Validate validates a policy document according to it's version's rule set.
+// if any rule is violated, returns an error
+func (policyDoc *Document) Validate() error {
+	// Constants
+	supportedPolicyVersions := []string{"1.0"}
```
shizhMSFT, line 138: Constants should be defined outside of the function.

verification/trustpolicy/trustpolicy.go
```diff
@@ -0,0 +1,448 @@
...
+		return applicablePolicy, nil
+	} else if wildcardPolicy != nil {
+		return wildcardPolicy, nil
+	} else {
+		return nil, fmt.Errorf("artifact %q has no applicable trust policy", artifactReference)
+	}
+}
+
+// deepCopy returns a pointer to the deeply copied TrustPolicy
+func deepCopy(t *TrustPolicy) *TrustPolicy {
```
shizhMSFT, line 425: It is better to rename it to
```go
func (t *TrustPolicy) clone() *TrustPolicy {
```

verification/truststore/truststore.go
```diff
@@ -0,0 +1,126 @@
...
+	Types = []Type{
+		TypeCA,
+		TypeSigningAuthority,
+	}
+)
+
+// X509TrustStore provide list and get behaviors for the trust store
+type X509TrustStore interface {
+	// List named stores under storeType
+	// List(ctx context.Context, storeType Type) ([]string, error)
```
shizhMSFT, line 36: It seems this method is not implemented yet.

verification/truststore/truststore.go
```diff
@@ -0,0 +1,126 @@
...
+		return nil, errors.New("storeType and namedStore cannot be empty")
+	}
+	path, err := trustStore.trustStorefs.SysPath(dir.X509TrustStoreDir(string(storeType), namedStore))
+	if err != nil {
+		return nil, err
+	}
+	// check path is valid
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		return nil, fmt.Errorf("%q does not exist", path)
+	}
```
shizhMSFT, line 64: There is a pitfall. See https://stefanxo.com/go-anti-patterns-os-isexisterr-os-isnotexisterr/

verification/truststore/truststore.go
patrickzheng200: I think we don't need this method. It is not used in the verification workflow.

internal/slice/slice.go
```diff
@@ -0,0 +1,16 @@
+package slice
+
+const (
+	Wildcard    = "*"
+	X509Subject = "x509.subject"
```
priteshbandi, line 5: If these are used only  in `trustpolicy.go` should we move them to `trustpolicy.go` file?

internal/slice/slice.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+// Contains is a utility function to check if a string exists in an array
+func Contains(val string, values []string) bool {
```
priteshbandi, line 9: Should we rename it to StringContains or ContainsString ?

verification/trustpolicy/trustpolicy.go
```diff
@@ -0,0 +1,486 @@
...
+		if registryScopeCount[key] > 1 {
+			return fmt.Errorf("registry scope %q is present in multiple trust policy statements, one registry scope value can only be associated with one statement", key)
+		}
+	}
+
+	// No error
+	return nil
+}
+
+func validateOverlappingDNs(policyName string, parsedDNs []parsedDN) error {
```
priteshbandi, line 377: Should we rename and move this method to pkix? perhaps  `HasOverlappingDNs` ?

verification/trustpolicy/trustpolicy.go
```diff
@@ -0,0 +1,486 @@
...
+		}
+	}
+	return false
+}
+
+// GetApplicableTrustPolicy returns a pointer to the deep copied TrustPolicy
+// statement that applies to the given registry URI. If no applicable trust
+// policy is found, returns an error
+// see https://github.com/notaryproject/notaryproject/blob/main/trust-store-trust-policy-specification.md#selecting-a-trust-policy-based-on-artifact-uri
+func GetApplicableTrustPolicy(trustPolicyDoc *Document, artifactReference string) (*TrustPolicy, error) {
```
priteshbandi, line 404: Should we make this method of type Document
```suggestion
func (policyDoc *Document) GetApplicableTrustPolicy(artifactReference string) (*TrustPolicy, error) {
```

verification/trustpolicy/trustpolicy.go
```diff
@@ -0,0 +1,486 @@
...
+		}
+	}
+
+	// No errors
+	return nil
+}
+
+// GetVerificationLevel returns VerificationLevel struct for the given
+// SignatureVerification struct throws error if SignatureVerification is invalid
+func GetVerificationLevel(signatureVerification SignatureVerification) (*VerificationLevel, error) {
```
priteshbandi, line 218: Should we make this member function of SignatureVerification type?

internal/slice/slice.go
shizhMSFT, line 9: Can we have something like https://pkg.go.dev/golang.org/x/exp/slices#Contains?

verification/truststore/truststore.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// x509TrustStore implements X509TrustStore
+type x509TrustStore struct {
+	trustStorefs dir.SysFS
+}
+
+// GetCertificates returns certificates under storeType/namedStore
+func (trustStore x509TrustStore) GetCertificates(ctx context.Context, storeType Type, namedStore string) ([]*x509.Certificate, error) {
+	if storeType == "" || namedStore == "" {
```
priteshbandi, line 51: IMO we should restrict most, if not all, the user-inputs to the following regex [a-zA-Z0-9_.-]+ i.e. alphanumeric and _, -.
This will avoid most of the [cross-platform issues](https://superuser.com/questions/358855/what-characters-are-safe-in-cross-platform-file-names-for-linux-windows-and-os). Also, we can either update spec with this PR or later(please open a issue, if we choose to do it later).


https://github.com/notaryproject/notation/pull/405#discussion_r1009061509

internal/slice/slice.go
shizhMSFT: The reason why we don't directly import `golang.org/x/exp` is that that repo is not stable. Here's the statement from `exp` package:
> **Warning**: Packages here are experimental and unreliable. Some may one day be promoted to the main repository or other subrepository, or they may be modified arbitrarily or even disappear altogether.

verification/trustpolicy/trustpolicy.go
```diff
@@ -0,0 +1,486 @@
...
+// clone returns a pointer to the deeply copied TrustPolicy
+func (t *TrustPolicy) clone() *TrustPolicy {
+	return &TrustPolicy{
+		Name:                  t.Name,
+		SignatureVerification: t.SignatureVerification,
+		RegistryScopes:        append([]string(nil), t.RegistryScopes...),
+		TrustedIdentities:     append([]string(nil), t.TrustedIdentities...),
+		TrustStores:           append([]string(nil), t.TrustStores...),
+	}
+}
```
shizhMSFT, line 443: Please put this block of code besides other member functions.

internal/slice/slice.go
patrickzheng200: They will also be used in the verifier package (this part is will be in PR #186 where verification package is refactored). If we put them into trustpolicy package, then they will become exported.

internal/slice/slice.go
patrickzheng200: Yeah, ContainsString might be better.

verification/trustpolicy/trustpolicy.go
patrickzheng200: This is only used in trustpolicy package, so instead of moving it into internal, we can keep it here unexported.

internal/slice/slice.go
patrickzheng200: I tried this before, but []string does not implement comparable.

verification/truststore/truststore.go
shizhMSFT: Will it be used in the `notation` CLI?

priteshbandi (approved):
LGTM

verification/truststore/truststore.go
```diff
@@ -50,17 +48,22 @@ type x509TrustStore struct {
...
 	if err != nil {
 		return nil, err
 	}
 	// check path is valid
-	if _, err := os.Stat(path); os.IsNotExist(err) {
-		return nil, fmt.Errorf("%q does not exist", path)
+	if _, err := os.Stat(path); err != nil {
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("%q does not exist", path)
+		}
```
shizhMSFT, line 66: Should we return err if it is an err but not a `IsNotExists` error?

verification/truststore/truststore.go
```diff
@@ -124,3 +127,18 @@ func validateCerts(certs []*x509.Certificate, path string) error {
...
 }
+
+// isValidStoreType checks if storeType is supported
+func isValidStoreType(storeType Type) bool {
+	for _, t := range Types {
+		if storeType == t {
+			return true
+		}
+	}
+	return false
```
shizhMSFT, line 138: How about
```suggestion
	return slice.Contains(Types, storeType)
```

PR #200, opened 2022-11-11
update: Package signature refactoring

patrickzheng200:
Files intended to be reviewed in this PR:
1. All files inside the signer package

(Refactoring of signer package in this PR incorporates the updated plugin and plugin/proto packages.)

PR #207, opened 2022-11-17
fix: update plugin to add ContractVersion

JeyJeyGao:
- Add ContractVersion argument when executing plugin request
- Check ContractVersion when getting plugin metadata
- Moved the path existence check from manager.Get() to NewCLIPlugin()

Test:
Added invalid contract version unit test.

### DIFF
--- a/internal/mock/mockfs/fs.go
+++ b/internal/mock/mockfs/fs.go
@@ -23,3 +23,9 @@ func NewSysFSMock(fsys fs.FS) dir.SysFS {
 		FS:   fsys,
 		root: ""}
 }
+
+func NewSysFSWithPathMock(fsys fs.FS, root string) dir.SysFS {
+	return sysFSMock{
+		FS:   fsys,
+		root: root}
+}
--- a/internal/mock/mockfs/fs.go
+++ b/internal/mock/mockfs/fs.go
@@ -24,7 +24,7 @@ func NewSysFSMock(fsys fs.FS) dir.SysFS {
 		root: ""}
 }
 
-func NewSysFSWithPathMock(fsys fs.FS, root string) dir.SysFS {
+func NewSysFSWithRootMock(fsys fs.FS, root string) dir.SysFS {
 	return sysFSMock{
 		FS:   fsys,
 		root: root}
--- a/internal/mock/mockfs/fs.go
+++ b/internal/mock/mockfs/fs.go
@@ -12,18 +12,22 @@ type sysFSMock struct {
 	root string
 }
 
+// SysPath returns the system path of the FS.
 func (s sysFSMock) SysPath(items ...string) (string, error) {
 	pathItems := []string{s.root}
 	pathItems = append(pathItems, items...)
 	return filepath.Join(pathItems...), nil
 }
 
+// NewSysFSMock returns a SysFS mock of the given FS.
 func NewSysFSMock(fsys fs.FS) dir.SysFS {
 	return sysFSMock{
 		FS:   fsys,
 		root: ""}
 }
 
+// NewSysFSWithRootMock returns a SysFS mock of the given fs and
+// a root directory
 func NewSysFSWithRootMock(fsys fs.FS, root string) dir.SysFS {
 	return sysFSMock{
 		FS:   fsys,
--- a/plugin/manager.go
+++ b/plugin/manager.go
@@ -18,7 +18,7 @@ var ErrNotRegularFile = errors.New("not regular file")
 
 // Manager manages plugins installed on the system.
 type Manager interface {
-	Get(ctx context.Context, name string) (Plugin, error)
+	Get(ctx context.Context, name string, pluginConfig map[string]string) (Plugin, error)
 	List(ctx context.Context) ([]string, error)
 }
 
@@ -35,7 +35,7 @@ func NewCLIManager(pluginFS dir.SysFS) *CLIManager {
 // Get returns a plugin on the system by its name.
 //
 // If the plugin is not found, the error is of type os.ErrNotExist.
-func (m *CLIManager) Get(ctx context.Context, name string) (Plugin, error) {
+func (m *CLIManager) Get(ctx context.Context, name string, pluginConfig map[string]string) (Plugin, error) {
 	// validate file existence
 	pluginPath := path.Join(name, binName(name))
 	fi, err := fs.Stat(m.pluginFS, pluginPath)
@@ -56,7 +56,7 @@ func (m *CLIManager) Get(ctx context.Context, name string) (Plugin, error) {
 	}
 
 	// validate and create plugin
-	return NewCLIPlugin(ctx, name, path)
+	return NewCLIPlugin(ctx, name, path, pluginConfig)
 }
 
 // List produces a list of the plugin names on the system.
--- a/plugin/manager.go
+++ b/plugin/manager.go
@@ -18,7 +18,7 @@ var ErrNotRegularFile = errors.New("not regular file")
 
 // Manager manages plugins installed on the system.
 type Manager interface {
-	Get(ctx context.Context, name string, pluginConfig map[string]string) (Plugin, error)
+	Get(ctx context.Context, name string) (Plugin, error)
 	List(ctx context.Context) ([]string, error)
 }
 
@@ -35,28 +35,15 @@ func NewCLIManager(pluginFS dir.SysFS) *CLIManager {
 // Get returns a plugin on the system by its name.
 //
 // If the plugin is not found, the error is of type os.ErrNotExist.
-func (m *CLIManager) Get(ctx context.Context, name string, pluginConfig map[string]string) (Plugin, error) {
-	// validate file existence
+func (m *CLIManager) Get(ctx context.Context, name string) (Plugin, error) {
 	pluginPath := path.Join(name, binName(name))
-	fi, err := fs.Stat(m.pluginFS, pluginPath)
-	if err != nil {
-		// Ignore any file which we cannot Stat
-		// (e.g. due to permissions or anything else).
-		return nil, err
-	}
-	if !fi.Mode().IsRegular() {
-		// Ignore non-regular files.
-		return nil, ErrNotRegularFile
-	}
-
-	// get path
 	path, err := m.pluginFS.SysPath(pluginPath)
 	if err != nil {
 		return nil, err
 	}
 
 	// validate and create plugin
-	return NewCLIPlugin(ctx, name, path, pluginConfig)
+	return NewCLIPlugin(ctx, name, path)
 }
 
 // List produces a list of the plugin names on the system.
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os/exec"
 	"path/filepath"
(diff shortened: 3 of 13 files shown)


shizhMSFT (commented):
Should we also check the contract version in the response by the package `plugin`?

patrickzheng200 (approved):
LGTM

patrickzheng200 (approved):
LGTM

plugin/manager.go
```diff
@@ -18,7 +18,7 @@ var ErrNotRegularFile = errors.New("not regular file")
 
 // Manager manages plugins installed on the system.
 type Manager interface {
-	Get(ctx context.Context, name string) (Plugin, error)
+	Get(ctx context.Context, name string, pluginConfig map[string]string) (Plugin, error)
```
shizhMSFT, line 21: @priteshbandi I'm confused why `get-plugin-metadata` has an optional `pluginConfig` in the request (see [spec](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#plugin-metadata)).

Could you elaborate a bit?

shizhMSFT: What's meaning of passing `pluginConfig` to `get-plugin-metadata`?

plugin/proto/metadata.go
```diff
@@ -42,6 +45,12 @@ func (resp *GetMetadataResponse) Validate() error {
 	if len(resp.SupportedContractVersions) == 0 {
 		return errors.New("empty supported contract versions")
 	}
+	if !resp.SupportsContract(ContractVersion) {
```
shizhMSFT, line 48: Should we unexport `Validate()` and call it in the `GetMetadata()` function?

plugin/plugin.go
```diff
@@ -88,27 +88,31 @@ func (p *CLIPlugin) GetMetadata(ctx context.Context, req *proto.GetMetadataReque
 // DescribeKey returns the KeySpec of a key.
 func (p *CLIPlugin) DescribeKey(ctx context.Context, req *proto.DescribeKeyRequest) (*proto.DescribeKeyResponse, error) {
 	var resp proto.DescribeKeyResponse
+	req.ContractVersion = proto.ContractVersion
 	err := run(ctx, p.name, p.path, req, &resp)
 	return &resp, err
```
shizhMSFT, line 93: Should we check the ContractVersion in the response before returning?

plugin/proto/metadata.go
shizhMSFT, line 48: Should we unexport `SupportsContract()` as the contract version set and check are encapsulated in the `plugin`, `plugin/proto` packages?

plugin/manager.go
shizhMSFT: What I can think about is passing `{"experimental":true}` or `{"support_old_contract":true}` to `get-plugin-metadata` so that the plugin can have different responses.

Are above scenarios valid?

plugin/plugin.go
JeyJeyGao: There is no ContractVersion in the response according to our [spec](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#describe-key).

plugin/proto/metadata.go
JeyJeyGao: Moved Validate() to plugin/plugin.go validate().

JeyJeyGao: moved SupportsContract to plugin/plugin.go supportsContract()

plugin/manager.go
priteshbandi: Yeah for example if a plugin use to have only `SIGNATURE_GENERATOR.RAW`capability now plugin owners want to launch `SIGNATURE_GENERATOR.ENVELOPE` capability but wants it to only enable when users explicitly passes the flag.

internal/plugin/manager/manager_test.go
```diff
@@ -1,332 +0,0 @@
-package manager
```
priteshbandi, line 1: Why we have removed all the tests?

priteshbandi:
@JeyJeyGao Please rebase the PR.

internal/plugin/manager/manager_test.go
JeyJeyGao: The internal/plugin package has been deleted now by this [commit](https://github.com/notaryproject/notation-go/commit/2573c88a5f62cf8c637bd43aceb8fb255c0e82c9).

plugin/manager.go
shizhMSFT: Makes sense.

signer/plugin.go
```diff
@@ -52,12 +52,6 @@ func (s *pluginSigner) Sign(ctx context.Context, desc ocispec.Descriptor, opts n
 	if err != nil {
 		return nil, nil, err
 	}
-	if !metadata.SupportsContract(proto.ContractVersion) {
```
shizhMSFT, line 55: Since #200 was merged recent, could you scan the package `signer` for updates like removing `ContractVersion: proto.ContractVersion` in the request?

plugin/plugin.go
```diff
@@ -164,3 +173,43 @@ func (c execCommander) Output(ctx context.Context, name string, command proto.Co
...
+// supportsContract return true if the metadata states that the
+// contract version is supported.
+func supportsContract(targetVer string, metadata *proto.GetMetadataResponse) bool {
+	for _, v := range metadata.SupportedContractVersions {
+		if v == targetVer {
+			return true
+		}
+	}
+	return false
+}
```
shizhMSFT, line 215: We can leverage `internal/slices.Contains` instead.

plugin/plugin.go
shizhMSFT: How about other like `GenerateSignature`, `GenerateEnvelope`, `VerifySignature`?

signer/plugin.go
JeyJeyGao: Updated.

plugin/plugin.go
JeyJeyGao: Updated.

plugin/plugin.go
JeyJeyGao: "supportedContractVersions" only exists in `get-plugin-metadata` response, so we can not check it for other commands.

plugin/plugin.go
```diff
@@ -164,3 +174,38 @@ func (c execCommander) Output(ctx context.Context, name string, command proto.Co
...
+		)
+	}
+	return nil
+}
+
+// supportsContract return true if the metadata states that the
+// contract version is supported.
+func supportsContract(targetVer string, metadata *proto.GetMetadataResponse) bool {
+	return slices.Contains(metadata.SupportedContractVersions, targetVer)
+}
```
shizhMSFT, line 211: nit: Seems we can replace line 198 with line 210.

plugin/plugin.go
```diff
@@ -57,58 +59,66 @@ type CLIPlugin struct {
...
 
 	// validate metadata
-	metadata, err := plugin.GetMetadata(ctx, &proto.GetMetadataRequest{})
+	metadata, err := plugin.GetMetadata(ctx, &proto.GetMetadataRequest{PluginConfig: pluginConfig})
 	if err != nil {
 		return nil, fmt.Errorf("failed to fetch metadata: %w", err)
 	}
 	if metadata.Name != name {
 		return nil, fmt.Errorf("executable name must be %q instead of %q", binName(metadata.Name), filepath.Base(path))
 	}
```
shizhMSFT, line 75: Should this check be in `plugin.GetMetadata()`?

Next question is "Do we really need to validate the plugin when calling `NewCLIPlugin()`?".

plugin/manager.go
```diff
@@ -35,7 +35,7 @@ func NewCLIManager(pluginFS dir.SysFS) *CLIManager {
 // Get returns a plugin on the system by its name.
 //
 // If the plugin is not found, the error is of type os.ErrNotExist.
-func (m *CLIManager) Get(ctx context.Context, name string) (Plugin, error) {
+func (m *CLIManager) Get(ctx context.Context, name string, pluginConfig map[string]string) (Plugin, error) {
 	// validate file existence
 	pluginPath := path.Join(name, binName(name))
 	fi, err := fs.Stat(m.pluginFS, pluginPath)
```
shizhMSFT, line 41: Should the operation of validating file existence in `NewCLIPlugin()` instead of `CLIManager.Get()`?

plugin/plugin.go
JeyJeyGao: 1. We need to validate the plugin name of reponse metadata that should be equal with user provided name. In `plugin.GetMetadata()` we don't have the user provided plugin name. 
2. We need to check the plugin name. When we return a new plugin, it should be a valid plugin or return an error. If we don't do the checking here, user also need to check it in another place.

plugin/plugin.go
JeyJeyGao: Updated.

plugin/plugin.go
shizhMSFT: > We need to validate the plugin name of reponse metadata that should be equal with user provided name. In plugin.GetMetadata() we don't have the user provided plugin name.

We do have the plugin name as the `name` parameter in `func NewCLIPlugin(ctx context.Context, name, path string, pluginConfig map[string]string) (*CLIPlugin, error)`

> We need to check the plugin name. When we return a new plugin, it should be a valid plugin or return an error. If we don't do the checking here, user also need to check it in another place.

Users always call `plugin.GetMetadata()` to determine the capability before use, don't they?

plugin/manager.go
JeyJeyGao: Updated. Moved to NewCLIPlugin()

plugin/plugin.go
JeyJeyGao: Updated.

plugin/plugin.go
```diff
@@ -58,57 +61,71 @@ type CLIPlugin struct {
...
+	if metadata.Name != p.name {
+		return nil, fmt.Errorf("executable name must be %q instead of %q", binName(metadata.Name), filepath.Base(p.path))
+	}
+	return &metadata, nil
 }
 
 // DescribeKey returns the KeySpec of a key.
 func (p *CLIPlugin) DescribeKey(ctx context.Context, req *proto.DescribeKeyRequest) (*proto.DescribeKeyResponse, error) {
 	var resp proto.DescribeKeyResponse
+	req.ContractVersion = proto.ContractVersion
```
priteshbandi, line 104: We should take `ContractVersion` in input as its value depends on the output of `et-plugin-metadata` command.  
Also, updating req struct inside of a method doesnt looks right and reduces readability

JeyJeyGao: Is it possible to let the plugin package to decide the ContractVersion and the package user don't need to care about the details? Because the ContractVersion checking is done in the plugin package, the package has enough information to decide the ContractVersion. When we have multiple ContractVersions in the future, we can have a multi-version selector in the package to set the ContractVersion.

Yes, it reduces the readability, so can we updated it like this to give a chance to let the user decide the version?
```go
if req.ContractVersion == ""
    req.ContractVersion = proto.ContractVersion
```

internal/mock/mockfs/fs.go
```diff
@@ -23,3 +23,9 @@ func NewSysFSMock(fsys fs.FS) dir.SysFS {
 		FS:   fsys,
 		root: ""}
 }
+
+func NewSysFSWithPathMock(fsys fs.FS, root string) dir.SysFS {
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

plugin/plugin.go
priteshbandi: So AFAIK contract version is driven by output of `get-plugin-metadata` command. If plugin package can get/store that value and use it for other commands then yes plugin package should full contract version.

Yes, this sounds better
```
if req.ContractVersion == ""
    req.ContractVersion = proto.ContractVersion
```

JeyJeyGao: - Added function doc to mention that the ContactVersion can be set automatically.
- We only have one version now, so the `get/store ContractVersion` logic is not required. It should be implemented when we really need it.

Is it acceptable?

internal/mock/mockfs/fs.go
```diff
@@ -23,3 +23,9 @@ func NewSysFSMock(fsys fs.FS) dir.SysFS {
 		FS:   fsys,
 		root: ""}
 }
+
+func NewSysFSWithRootMock(fsys fs.FS, root string) dir.SysFS {
```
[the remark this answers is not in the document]
JeyJeyGao: Added doc.

plugin/plugin.go
priteshbandi: Yes that sounds good.

priteshbandi (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #208, opened 2022-11-18
update: check SignatureMediaType in notation.Verify

patrickzheng200:
In this PR:
1. Created RemoteVerifyOptions for notation.Verify(), and therefore, VerifyOptions is now only for verifier.Verify().
2. In notation.Verify(), using signature media type fetched from registry.

### DIFF
--- a/notation.go
+++ b/notation.go
@@ -13,7 +13,6 @@ import (
 	"time"
 
 	"github.com/notaryproject/notation-core-go/signature"
-	"github.com/notaryproject/notation-go/internal/slices"
 	"github.com/notaryproject/notation-go/registry"
 	"github.com/notaryproject/notation-go/verifier/trustpolicy"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -141,12 +140,6 @@ type RemoteVerifyOptions struct {
 	// verified against to.
 	ArtifactReference string
 
-	// SignatureMediaTypes specifies the supported signature envelope types.
-	// If an empty list is present, all types are attempted.
-	// Currently both `application/jose+json` and `application/cose` are
-	// supported.
-	SignatureMediaTypes []string
-
 	// PluginConfig is a map of plugin configs.
 	PluginConfig map[string]string
 
@@ -203,11 +196,7 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, re
 			if err != nil {
 				return ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("unable to retrieve digital signature with digest %q associated with %q from the registry, error : %v", sigManifestDesc.Digest, artifactRef, err.Error())}
 			}
-			// checking the media type after fetching the blob is slow
-			// created an issue to track this: https://github.com/notaryproject/notation-go/issues/215
-			if len(remoteOpts.SignatureMediaTypes) != 0 && !slices.Contains(remoteOpts.SignatureMediaTypes, sigDesc.MediaType) {
-				continue
-			}
+			// using signature media type fetched from registry
 			opts.SignatureMediaType = sigDesc.MediaType
 
 			// verify each signature
--- a/notation.go
+++ b/notation.go
@@ -73,43 +73,6 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, opts Sig
 	return targetDesc, nil
 }
 
-// RemoteVerifyOptions contains parameters for notation.Verify.
-type RemoteVerifyOptions struct {
-	// ArtifactReference is the reference of the artifact that is been
-	// verified against to.
-	ArtifactReference string
-
-	// SignatureMediaTypes contains envelope types of the signature.
-	// Currently both `application/jose+json` and `application/cose` are
-	// supported.
-	SignatureMediaTypes []string
-
-	// PluginConfig is a map of plugin configs.
-	PluginConfig map[string]string
-
-	// MaxSignatureAttempts is the maximum number of signature envelopes that
-	// can be associated with the target artifact. If set to less than or equals
-	// to zero, an error will be returned.
-	// Note: this option is scoped to notation.Verify(). verifier.Verify() is
-	// for signle signature verification, and therefore, does not use it.
-	MaxSignatureAttempts int
-}
-
-// VerifyOptions contains parameters for verifier.Verify.
-type VerifyOptions struct {
-	// ArtifactReference is the reference of the artifact that is been
-	// verified against to.
-	ArtifactReference string
-
-	// SignatureMediaType is the envelope type of the signature.
-	// Currently both `application/jose+json` and `application/cose` are
-	// supported.
-	SignatureMediaType string
-
-	// PluginConfig is a map of plugin configs.
-	PluginConfig map[string]string
-}
-
 // ValidationResult encapsulates the verification result (passed or failed)
 // for a verification type, including the desired verification action as
 // specified in the trust policy
@@ -148,6 +111,21 @@ type VerificationOutcome struct {
 	Error error
 }
 
+// VerifyOptions contains parameters for Verifier.Verify.
+type VerifyOptions struct {
+	// ArtifactReference is the reference of the artifact that is been
+	// verified against to.
+	ArtifactReference string
+
+	// SignatureMediaType is the envelope type of the signature.
+	// Currently both `application/jose+json` and `application/cose` are
+	// supported.
+	SignatureMediaType string
+
+	// PluginConfig is a map of plugin configs.
+	PluginConfig map[string]string
+}
+
 // Verifier is a generic interface for verifying an artifact.
 type Verifier interface {
 	// Verify verifies the signature blob and returns the outcome upon
@@ -155,6 +133,27 @@ type Verifier interface {
 	// If nil signature is present and the verification level is not 'skip',
 	// an error will be returned.
 	Verify(ctx context.Context, desc ocispec.Descriptor, signature []byte, opts VerifyOptions) (*VerificationOutcome, error)
+}
+
+// RemoteVerifyOptions contains parameters for notation.Verify.
+type RemoteVerifyOptions struct {
+	// ArtifactReference is the reference of the artifact that is been
+	// verified against to.
+	ArtifactReference string
+
+	// SignatureMediaTypes specifies the supported signature envelope types.
+	// If an empty list is present, all types are attempted.
+	// Currently both `application/jose+json` and `application/cose` are
+	// supported.
+	SignatureMediaTypes []string
+
+	// PluginConfig is a map of plugin configs.
+	PluginConfig map[string]string
+
+	// MaxSignatureAttempts is the maximum number of signature envelopes that
+	// can be associated with the target artifact. If set to less than or equals
+	// to zero, an error will be returned.
+	MaxSignatureAttempts int
 }
 
 // Verify performs signature verification on each of the notation supported
@@ -204,6 +203,8 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, re
 			if err != nil {
 				return ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("unable to retrieve digital signature with digest %q associated with %q from the registry, error : %v", sigManifestDesc.Digest, artifactRef, err.Error())}
 			}
+			// checking the media type after fetching the blob is slow
+			// created an issue to track this: https://github.com/notaryproject/notation-go/issues/215
 			if len(remoteOpts.SignatureMediaTypes) != 0 && !slices.Contains(remoteOpts.SignatureMediaTypes, sigDesc.MediaType) {
 				continue
 			}
--- a/notation.go
+++ b/notation.go
@@ -184,12 +184,13 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, re
 	}
 
 	var verificationOutcomes []*VerificationOutcome
+	errExceededMaxVerificationLimit := ErrorVerificationFailed{Msg: fmt.Sprintf("total number of signatures associated with an artifact should be less than: %d", remoteOpts.MaxSignatureAttempts)}
 	numOfSignatureProcessed := 0
 	err = repo.ListSignatures(ctx, artifactDescriptor, func(signatureManifests []ocispec.Descriptor) error {
 		// process signatures
 		for _, sigManifestDesc := range signatureManifests {
 			if numOfSignatureProcessed >= remoteOpts.MaxSignatureAttempts {
-				return ErrorVerificationFailed{Msg: fmt.Sprintf("total number of signatures associated with an artifact should be less than: %d", remoteOpts.MaxSignatureAttempts)}
+				break
 			}
 			numOfSignatureProcessed++
 			// get signature envelope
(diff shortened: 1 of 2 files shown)


notation.go
```diff
@@ -177,10 +177,19 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, op
...
 			if err != nil {
 				return ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("unable to retrieve digital signature with digest %q associated with %q from the registry, error : %v", sigManifestDesc.Digest, artifactRef, err.Error())}
 			}
+			// opts.SignatureMediaType is non-emtpy but does not match the
+			// sigDesc.MediaType, skip the current signature.
+			if opts.SignatureMediaType != "" && opts.SignatureMediaType != sigDesc.MediaType {
+				continue
+			}
+			// Otherwise, use the MediaType fetched from repo
+			opts.SignatureMediaType = sigDesc.MediaType
```
JeyJeyGao, line 190: A little confusing. Correct me if I'm wrong.
Suppose we have 2 sigantures and opts.SignatureMediaType is empty:
1. when processing the first one `sigCose1`:  opts.SignatureMediaType = "application/cose"
2. when processing the second one `sigJWS2`: this signature will be skipped.
Because opts.SignatureMediaType is `empty`, user don't want to skip any signature type.

priteshbandi (commented):
Why are we even taking `opts.SignatureMediaType` as input? Won't hurt multi-signature envelope support?

notation.go
patrickzheng200: Code updated.

patrickzheng200:
> Why are we even taking `opts.SignatureMediaType` as input? 
This enables users to pick their trusted signature format. Signature format that they don't trust will not be processed even if it might pass the verification process.

> Also, will checking check SignatureMediaType in notation.Verify hurt multi-signature envelope support?
No, it won't hurt multi-signature format support. 
Basically, there are 3 scenarios:
1. VerifyOptions.SignatureMediaType is empty. In this case, MediaType fetched from repo will be used.
2. VerifyOptions.SignatureMediaType is not empty but does not match MediaType fetched from repo. In this case, this certain signature will be skipped, and the program will continue to verify the next signature.
3. VerifyOptions.SignatureMediaType equals to MediaType fetched from repo. This MediaType will be used to verify the signature.
To enable both JWS and COSE, user could just leave VerifyOptions.SignatureMediaType as empty. (I updated the code to reflect this.)

shizhMSFT (approved):
LGTM

notation.go
JeyJeyGao: Is it possible that the user reuses the `VerifyOptions` but the user doesn't know we changed the `opts.SignatureMediaType` which causing a issue?

JeyJeyGao (approved):
LGTM

notation.go
```diff
@@ -169,6 +169,7 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, op
 	}
 	errExceededMaxVerificationLimit := ErrorVerificationFailed{Msg: fmt.Sprintf("total number of signatures associated with an artifact should be less than: %d", opts.MaxSignatureAttempts)}
 	numOfSignatureProcessed := 0
+	userSignaureMediaType := opts.SignatureMediaType
```
priteshbandi, line 172: `opts.SignatureMediaType` should be an array to allow user to trust multiple signature formats. Also, an empty array here means they trust all.

priteshbandi:
> > Why are we even taking `opts.SignatureMediaType` as input?
> 
> This enables users to pick their trusted signature format. Signature format that they don't trust will not be processed even if it might pass the verification process.
> 
The question here is, if a user trust notary v2 for signing and verifying, why wont they trust a signature format that has been vetted by notary v2 ?

> > Also, will checking check SignatureMediaType in notation.Verify hurt multi-signature envelope support?
> 
> No, it won't hurt multi-signature format support. Basically, there are 3 scenarios:
> 
> 1. VerifyOptions.SignatureMediaType is empty. In this case, MediaType fetched from repo will be used.
> 2. VerifyOptions.SignatureMediaType is not empty but does not match MediaType fetched from repo. In this case, this certain signature will be skipped, and the program will continue to verify the next signature.
> 3. VerifyOptions.SignatureMediaType equals to MediaType fetched from repo. This MediaType will be used to verify the signature.
> 
> To enable both JWS and COSE, user could just leave VerifyOptions.SignatureMediaType as empty. (I updated the code to reflect this.)

Yes, that's an implementation detail. I am wondering why will a user want to trust one signature format but not the other. Also, if an implementation does that then it's not abiding by notartv2 spec.

notation.go
patrickzheng200: Added unit test. As for changing it to array, IMO it's not very necessary at this point since we only support two formats. An empty user input means supporting both of them. Besides, changing VerifyOption would require Ratify to change accordingly.

patrickzheng200:
> > > Why are we even taking `opts.SignatureMediaType` as input?
> > 
> > 
> > This enables users to pick their trusted signature format. Signature format that they don't trust will not be processed even if it might pass the verification process.
> 
> The question here is, if a user trust notary v2 for signing and verifying, why wont they trust a signature format that has been vetted by notary v2 ?
> 
> > > Also, will checking check SignatureMediaType in notation.Verify hurt multi-signature envelope support?
> > 
> > 
> > No, it won't hurt multi-signature format support. Basically, there are 3 scenarios:
> > 
> > 1. VerifyOptions.SignatureMediaType is empty. In this case, MediaType fetched from repo will be used.
> > 2. VerifyOptions.SignatureMediaType is not empty but does not match MediaType fetched from repo. In this case, this certain signature will be skipped, and the program will continue to verify the next signature.
> > 3. VerifyOptions.SignatureMediaType equals to MediaType fetched from repo. This MediaType will be used to verify the signature.
> > 
> > To enable both JWS and COSE, user could just leave VerifyOptions.SignatureMediaType as empty. (I updated the code to reflect this.)
> 
> Yes, that's an implementation detail. I am wondering why will a user want to trust one signature format but not the other. Also, if an implementation does that then it's not abiding by notartv2 spec.

Please don't get me wrong, I don't mean users should or need to pick one specific signature format here. Leaving the field empty is the default behavior, i.e. taking both signature formats is our default.  The purpose of this PR is not enabling users to pick a signature format. On the contrary, previous to this PR, the logic is only using opts.SignatureMediaType. This PR actually fixes this issue and takes SignatureMediaType from registry.

notation.go
priteshbandi: If this is external facing method used by ratify, I would suggest changing this to array to avoid any breaking changes post RC1.

priteshbandi:
> > > > Why are we even taking `opts.SignatureMediaType` as input?
> > > 
> > > 
> > > This enables users to pick their trusted signature format. Signature format that they don't trust will not be processed even if it might pass the verification process.
> > 
> > 
> > The question here is, if a user trust notary v2 for signing and verifying, why wont they trust a signature format that has been vetted by notary v2 ?
> > > > Also, will checking check SignatureMediaType in notation.Verify hurt multi-signature envelope support?
> > > 
> > > 
> > > No, it won't hurt multi-signature format support. Basically, there are 3 scenarios:
> > > 
> > > 1. VerifyOptions.SignatureMediaType is empty. In this case, MediaType fetched from repo will be used.
> > > 2. VerifyOptions.SignatureMediaType is not empty but does not match MediaType fetched from repo. In this case, this certain signature will be skipped, and the program will continue to verify the next signature.
> > > 3. VerifyOptions.SignatureMediaType equals to MediaType fetched from repo. This MediaType will be used to verify the signature.
> > > 
> > > To enable both JWS and COSE, user could just leave VerifyOptions.SignatureMediaType as empty. (I updated the code to reflect this.)
> > 
> > 
> > Yes, that's an implementation detail. I am wondering why will a user want to trust one signature format but not the other. Also, if an implementation does that then it's not abiding by notartv2 spec.
> 
> Please don't get me wrong, I don't mean users should or need to pick one specific signature format here. Leaving the field empty is the default behavior, i.e. taking both signature formats is our default. The purpose of this PR is not enabling users to pick a signature format. On the contrary, previous to this PR, the logic is only using opts.SignatureMediaType. This PR actually fixes this issue and takes SignatureMediaType from registry.

if the purpose of this PR is not enabling users to pick a signature format, can we always use mediatype returned by signature manifest i.e. user shouldn't be able to choose?

notation.go
priteshbandi: I am confused, either we need to take this as an array or we should completely drop this functionality and always use `mediaType` present signature artifact manifest.

patrickzheng200:
> > > > > Why are we even taking `opts.SignatureMediaType` as input?
> > > > 
> > > > 
> > > > This enables users to pick their trusted signature format. Signature format that they don't trust will not be processed even if it might pass the verification process.
> > > 
> > > 
> > > The question here is, if a user trust notary v2 for signing and verifying, why wont they trust a signature format that has been vetted by notary v2 ?
> > > > > Also, will checking check SignatureMediaType in notation.Verify hurt multi-signature envelope support?
> > > > 
> > > > 
> > > > No, it won't hurt multi-signature format support. Basically, there are 3 scenarios:
> > > > 
> > > > 1. VerifyOptions.SignatureMediaType is empty. In this case, MediaType fetched from repo will be used.
> > > > 2. VerifyOptions.SignatureMediaType is not empty but does not match MediaType fetched from repo. In this case, this certain signature will be skipped, and the program will continue to verify the next signature.
> > > > 3. VerifyOptions.SignatureMediaType equals to MediaType fetched from repo. This MediaType will be used to verify the signature.
> > > > 
> > > > To enable both JWS and COSE, user could just leave VerifyOptions.SignatureMediaType as empty. (I updated the code to reflect this.)
> > > 
> > > 
> > > Yes, that's an implementation detail. I am wondering why will a user want to trust one signature format but not the other. Also, if an implementation does that then it's not abiding by notartv2 spec.
> > 
> > 
> > Please don't get me wrong, I don't mean users should or need to pick one specific signature format here. Leaving the field empty is the default behavior, i.e. taking both signature formats is our default. The purpose of this PR is not enabling users to pick a signature format. On the contrary, previous to this PR, the logic is only using opts.SignatureMediaType. This PR actually fixes this issue and takes SignatureMediaType from registry.
> 
> if the purpose of this PR is not enabling users to pick a signature format, can we always use mediatype returned by signature manifest i.e. user shouldn't be able to choose?

Code updated. Changed the field to an array.

notation.go
```diff
@@ -94,6 +95,21 @@ type VerifyOptions struct {
 	MaxSignatureAttempts int
```
shizhMSFT, line 95: The documentation of this field needs to be updated.

notation.go
```diff
@@ -156,31 +178,38 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, op
...
 			numOfSignatureProcessed++
 			// get signature envelope
-			sigBlob, _, err := repo.FetchSignatureBlob(ctx, sigManifestDesc)
+			sigBlob, sigDesc, err := repo.FetchSignatureBlob(ctx, sigManifestDesc)
 			if err != nil {
 				return ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("unable to retrieve digital signature with digest %q associated with %q from the registry, error : %v", sigManifestDesc.Digest, artifactRef, err.Error())}
 			}
+			if len(remoteOpts.SignatureMediaTypes) != 0 && !slices.Contains(remoteOpts.SignatureMediaTypes, sigDesc.MediaType) {
+				continue
+			}
```
shizhMSFT, line 209: The performance of this check is slow since it checks the media type after fetching the blob. Could you create an issue to track it?

patrickzheng200: created an issue to track this. (#215)

shizhMSFT (approved):
LGTM

patrickzheng200:
> 

@priteshbandi Oh yeah, I've updated the code. PTAL.

notation.go
```diff
@@ -141,12 +134,33 @@ type Verifier interface {
...
+	// ArtifactReference is the reference of the artifact that is been
+	// verified against to.
+	ArtifactReference string
+
+	// PluginConfig is a map of plugin configs.
+	PluginConfig map[string]string
+
+	// MaxSignatureAttempts is the maximum number of signature envelopes that
+	// can be associated with the target artifact. If set to less than or equals
+	// to zero, an error will be returned.
```
priteshbandi, line 148:
```suggestion
	// MaxSignatureAttempts is the maximum number of signature envelopes that
	// will be processed for verification. If set to less than or equals
	// to zero, an error will be returned.
```

notation.go
```diff
@@ -156,31 +170,36 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, op
...
 	if err != nil {
 		return ocispec.Descriptor{}, nil, ErrorSignatureRetrievalFailed{Msg: err.Error()}
 	}
 
 	var verificationOutcomes []*VerificationOutcome
-	if opts.MaxSignatureAttempts <= 0 {
-		return ocispec.Descriptor{}, verificationOutcomes, ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("verifyOptions.MaxSignatureAttempts expects a positive number, got %d", opts.MaxSignatureAttempts)}
+	if remoteOpts.MaxSignatureAttempts <= 0 {
+		return ocispec.Descriptor{}, verificationOutcomes, ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("verifyOptions.MaxSignatureAttempts expects a positive number, got %d", remoteOpts.MaxSignatureAttempts)}
 	}
```
priteshbandi, line 184: We should perform this check at line 173 .

Reason: Why to resolve a reference when we are going to throw an error.

shizhMSFT (approved):
LGTM

notation.go
```diff
@@ -156,31 +170,38 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, op
...
 	if err != nil {
 		return ocispec.Descriptor{}, nil, ErrorSignatureRetrievalFailed{Msg: err.Error()}
 	}
 
 	var verificationOutcomes []*VerificationOutcome
-	if opts.MaxSignatureAttempts <= 0 {
-		return ocispec.Descriptor{}, verificationOutcomes, ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("verifyOptions.MaxSignatureAttempts expects a positive number, got %d", opts.MaxSignatureAttempts)}
-	}
-	errExceededMaxVerificationLimit := ErrorVerificationFailed{Msg: fmt.Sprintf("total number of signatures associated with an artifact should be less than: %d", opts.MaxSignatureAttempts)}
+	errExceededMaxVerificationLimit := ErrorVerificationFailed{Msg: fmt.Sprintf("total number of signatures associated with an artifact should be less than: %d", remoteOpts.MaxSignatureAttempts)}
```
shizhMSFT, line 187: nit: we can merge this line to line 223

patrickzheng200: And we can move line 222 to line 193, i.e. use the return to replace the break statement in the for loop.

priteshbandi (approved):
LGTM

shizhMSFT (approved):
LGTM

notation.go
shizhMSFT: The change related to this comment is reverted.

PR #210, opened 2022-11-21
Add: added set data structure

### LINKED ISSUE #203: Add set data structure at internal/container/set
Created an issue to track this.

patrickzheng200:
This PR adds a generic Set data structure into the library.

Resolves #203.

### DIFF
--- a/internal/container/set.go
+++ b/internal/container/set.go
@@ -1,14 +1,23 @@
 package set
 
-// Set is a map as a set data structure
+// Set is a map as a set data structure.
 type Set[T comparable] map[T]struct{}
 
-// Add adds the element of type T into the Set
+// Add adds the element of type T into the Set.
 func (s Set[T]) Add(elem T) {
 	s[elem] = struct{}{}
 }
 
-// New creats an empty Set for elements of type T
+// Contains checks if element exists in the Set.
+func (s Set[T]) Contains(elem T) bool {
+	if _, ok := s[elem]; ok {
+		return true
+	}
+
+	return false
+}
+
+// New creats an empty Set for elements of type T.
 func New[T comparable]() Set[T] {
 	return make(map[T]struct{})
 }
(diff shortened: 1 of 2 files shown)


internal/container/set.go
```diff
@@ -0,0 +1,23 @@
+package set
+
+// Set is a map as a set data structure.
+type Set[T comparable] map[T]struct{}
```
JeyJeyGao, line 4:
```suggestion
type Set[T any] map[T]struct{}
```
interface{} can be the element of a map, so is it better?

internal/container/set.go
```diff
@@ -0,0 +1,23 @@
...
+	s[elem] = struct{}{}
+}
+
+// Contains checks if element exists in the Set.
+func (s Set[T]) Contains(elem T) bool {
+	if _, ok := s[elem]; ok {
+		return true
+	}
+
+	return false
```
JeyJeyGao, line 17: Can we return `ok` directly?

internal/container/set.go
patrickzheng200: As our notation-go library is still using go 1.18, your suggested change would cause error: "incomparable map key type T (missing comparable constraint)". For more details, please see https://github.com/golang/go/issues/51338.

internal/container/set.go
patrickzheng200: Nice catch!

jondonas (approved):

JeyJeyGao (approved):
LGTM

internal/container/set.go
```diff
@@ -0,0 +1,21 @@
...
+}
+
+// Contains checks if element exists in the Set.
+func (s Set[T]) Contains(elem T) bool {
+	_, ok := s[elem]
+
+	return ok
+}
+
+// New creats an empty Set for elements of type T.
```
shizhMSFT, line 18: nit: typo: 
```suggestion
// New creates an empty Set for elements of type T.
```

Issue #212, opened 2022-11-22
Behavior of VerifyOptions.MaxSignatureAttempts


patrickzheng200:
This issue tracks the behavior of VerifyOptions.MaxSignatureAttempts during notation.Verify(). As of 11//22/2022, we will have the following logics:

1. Default of VerifyOptions.MaxSignatureAttempts is 0, when a user decides to use default, notation.Verify() would verify unlimited number of signatures associated with the artifact.
2. User gives a positive number to VerifyOptions.MaxSignatureAttempts, notation.Verify() would verify at most VerifyOptions.MaxSignatureAttempts number of signatures associated with the artifact.
3. User gives a negative number to VerifyOptions.MaxSignatureAttempts, same behavior as case 1.

@shizhMSFT @yizha1 @priteshbandi @rgnote Please leave your concerns/comments/suggestions if there's any.

priteshbandi:
Should we use 0 and -ve to default for unlimited or user can pass `math.MaxInt64` ?

patrickzheng200:
Closed this issue since we decided to not make change to current behavior in notation-go. In notation CLI, we will pass in a larger number such as `math.MaxInt64` as unlimited.

PR #213, opened 2022-11-22
update: updated MaxSignatureAttempts logic

patrickzheng200:
This PR updates logic of MaxSignatureAttempts based on issue #212.

### DIFF
--- a/notation.go
+++ b/notation.go
@@ -172,7 +172,7 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, op
 			// user specifies a positive opts.MaxSignatureAttempts
 			// Tracked by issue: https://github.com/notaryproject/notation-go/issues/212
 			if opts.MaxSignatureAttempts > 0 && numOfSignatureProcessed >= opts.MaxSignatureAttempts {
-				break
+				return errExceededMaxVerificationLimit
 			}
 			numOfSignatureProcessed++
 			// get signature envelope
@@ -194,10 +194,6 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, op
 			verificationOutcomes = append(verificationOutcomes, outcome)
 
 			return errDoneVerification
-		}
-
-		if opts.MaxSignatureAttempts > 0 && numOfSignatureProcessed >= opts.MaxSignatureAttempts {
-			return errExceededMaxVerificationLimit
 		}
 
 		return nil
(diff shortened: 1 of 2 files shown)


notation.go
```diff
@@ -164,15 +164,14 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, op
...
 	errExceededMaxVerificationLimit := ErrorVerificationFailed{Msg: fmt.Sprintf("total number of signatures associated with an artifact should be less than: %d", opts.MaxSignatureAttempts)}
 	numOfSignatureProcessed := 0
 	err = repo.ListSignatures(ctx, artifactDescriptor, func(signatureManifests []ocispec.Descriptor) error {
 		// process signatures
 		for _, sigManifestDesc := range signatureManifests {
-			if numOfSignatureProcessed >= opts.MaxSignatureAttempts {
+			// user specifies a positive opts.MaxSignatureAttempts
+			// Tracked by issue: https://github.com/notaryproject/notation-go/issues/212
+			if opts.MaxSignatureAttempts > 0 && numOfSignatureProcessed >= opts.MaxSignatureAttempts {
 				break
```
JeyJeyGao, line 175: Can we return `errExceededMaxVerificationLimit` directly?

JeyJeyGao (approved):
LGTM

priteshbandi:
Why are we updating this logic? Can we keep this code and caller can pass math.MaxInt64 which is unlimited for all practical purpose? These not only simplifies our code logic and makes it more readable for caller.

patrickzheng200:
> Why are we updating this logic? Can we keep this code and caller can pass math.MaxInt64 which is unlimited for all practical purpose? These not only simplifies our code logic and makes it more readable for caller.

Just want to clarify, the original logic has the `opts.MaxSignatureAttempts <= 0` check, and the default of `opts.MaxSignatureAttempts` is 0. It means that caller has to pass a value to `opts.MaxSignatureAttempts`, otherwise notation.Verify() would fail. Is this what we want? This PR intends to remove this check, so if caller ignore this field, notation.Verify() will still process and verify unlimited number of signatures.

shizhMSFT:
Let's close this PR due to security considerations on unlimited signature verification.

patrickzheng200:
> Let's close this PR due to security considerations on unlimited signature verification.

Makes sence. PR closed.

patrickzheng200:
> Why are we updating this logic? Can we keep this code and caller can pass math.MaxInt64 which is unlimited for all practical purpose? These not only simplifies our code logic and makes it more readable for caller.

Makes sence. Closed this PR, instead will set a large number such as math.MaxInt64 inside notation CLI.

PR #222, opened 2022-11-30
Pass expiry to envelope-generator plugin

priteshbandi:
There will be two PRs, one in notation-go([PR#222](https://github.com/notaryproject/notation-go/pull/222)) and other in notation([PR#458](https://github.com/notaryproject/notation/pull/458)) repo.

*Tests are failing because this PR depends on https://github.com/notaryproject/notation-go/pull/222*

Issue: https://github.com/notaryproject/notation/issues/443

### DIFF
--- a/notation.go
+++ b/notation.go
@@ -32,8 +32,8 @@ type SignOptions struct {
 	// supported.
 	SignatureMediaType string
 
-	// Expiry identifies the expiration time of the resulted signature.
-	Expiry time.Time
+	// ExpiryDuration identifies the expiry duration of the resulted signature.
+	ExpiryDuration *time.Duration
 
 	// Sets or overrides the plugin configuration.
 	PluginConfig map[string]string
@@ -52,6 +52,17 @@ type Signer interface {
 // remote.
 // The descriptor of the sign content is returned upon sucessful signing.
 func Sign(ctx context.Context, signer Signer, repo registry.Repository, opts SignOptions) (ocispec.Descriptor, error) {
+	// Input validation for expiry duration
+	if opts.ExpiryDuration != nil {
+		expDurInSec := opts.ExpiryDuration.Seconds()
+		if expDurInSec != float64(int64(expDurInSec)) {
+			return ocispec.Descriptor{}, fmt.Errorf("expiry duration supports minimum granulaity of second")
+		}
+		if expDurInSec <= 0 {
+			return ocispec.Descriptor{}, fmt.Errorf("expiry duration cannot be zero or negative")
+		}
+	}
+
 	targetDesc, err := repo.Resolve(ctx, opts.ArtifactReference)
 	if err != nil {
 		return ocispec.Descriptor{}, err
--- a/notation.go
+++ b/notation.go
@@ -56,7 +56,7 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, opts Sig
 	if opts.ExpiryDuration != nil {
 		expDurInSec := opts.ExpiryDuration.Seconds()
 		if expDurInSec != float64(int64(expDurInSec)) {
-			return ocispec.Descriptor{}, fmt.Errorf("expiry duration supports minimum granulaity of second")
+			return ocispec.Descriptor{}, fmt.Errorf("expiry duration supports minimum granularity of second")
 		}
 		if expDurInSec <= 0 {
 			return ocispec.Descriptor{}, fmt.Errorf("expiry duration cannot be zero or negative")
--- a/notation.go
+++ b/notation.go
@@ -32,8 +32,9 @@ type SignOptions struct {
 	// supported.
 	SignatureMediaType string
 
-	// ExpiryDuration identifies the expiry duration of the resulted signature.
-	ExpiryDuration *time.Duration
+	// ExpiryDuration identifies the expiry duration of the resulted signature. Zero value
+	// represents no expiry duration.
+	ExpiryDuration time.Duration
 
 	// Sets or overrides the plugin configuration.
 	PluginConfig map[string]string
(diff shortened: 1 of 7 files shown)


signer/plugin_test.go
```diff
@@ -13,6 +13,8 @@ import (
 	"time"
 
 	"github.com/notaryproject/notation-core-go/signature"
+	_ "github.com/notaryproject/notation-core-go/signature/cose"
+	_ "github.com/notaryproject/notation-core-go/signature/jws"
```
priteshbandi, line 17: Currently most of the tests in notation-go signer package are not executing because`signature.RegisteredEnvelopeTypes()` returns empty array. These lines fixes that issue

notation.go
```diff
@@ -52,6 +52,17 @@ type Signer interface {
 // remote.
 // The descriptor of the sign content is returned upon sucessful signing.
 func Sign(ctx context.Context, signer Signer, repo registry.Repository, opts SignOptions) (ocispec.Descriptor, error) {
+	// Input validation for expiry duration
+	if opts.ExpiryDuration != nil {
+		expDurInSec := opts.ExpiryDuration.Seconds()
+		if expDurInSec != float64(int64(expDurInSec)) {
+			return ocispec.Descriptor{}, fmt.Errorf("expiry duration supports minimum granulaity of second")
```
patrickzheng200, line 59: nit:
```suggestion
			return ocispec.Descriptor{}, fmt.Errorf("expiry duration supports minimum granularity of second")
```

notation.go
```diff
@@ -52,6 +52,17 @@ type Signer interface {
 // remote.
 // The descriptor of the sign content is returned upon sucessful signing.
 func Sign(ctx context.Context, signer Signer, repo registry.Repository, opts SignOptions) (ocispec.Descriptor, error) {
+	// Input validation for expiry duration
+	if opts.ExpiryDuration != nil {
+		expDurInSec := opts.ExpiryDuration.Seconds()
+		if expDurInSec != float64(int64(expDurInSec)) {
```
shizhMSFT, line 58: Converting from `int64` to `float64` then to `int64` and then to `float64` seems not right.

Suggesting
```suggestion
		if opts.ExpiryDuration % time.Second != 0 {
```

signer/plugin.go
```diff
@@ -104,6 +104,12 @@ func (s *pluginSigner) generateSignatureEnvelope(ctx context.Context, desc ocisp
 		PayloadType:           envelope.MediaTypePayloadV1,
 		PluginConfig:          s.mergeConfig(opts.PluginConfig),
 	}
+
+	if opts.ExpiryDuration != nil {
+		// expiry duration will always be a positive value
+		req.ExpiryDurationInSeconds = uint64(opts.ExpiryDuration.Seconds())
```
shizhMSFT, line 110: Same as above
```suggestion
		req.ExpiryDurationInSeconds = uint64(opts.ExpiryDuration / time.Second)
```

signer/signer.go
```diff
@@ -90,8 +90,9 @@ func (s *genericSigner) Sign(ctx context.Context, desc ocispec.Descriptor, opts
 		SigningAgent:  signingAgent, // TODO: include external signing plugin's name and version. https://github.com/notaryproject/notation-go/issues/80
 	}
 
-	if !opts.Expiry.IsZero() {
-		signReq.Expiry = opts.Expiry
+	// Add expiry only if ExpiryDuration is not nil and its value is not zero
+	if opts.ExpiryDuration != nil && opts.ExpiryDuration.Seconds() != 0 {
+		signReq.Expiry = signReq.SigningTime.Add(*opts.ExpiryDuration)
 	}
```
shizhMSFT, line 96: If we don't add expiry if its value is zero, then why we need to make `opts.ExpiryDuration` a `*time.Duration` pointer? Will `ExpiryDuration time.Duration` be better?

priteshbandi: Having expiry as pointer(`*time.Duration`) helps to convey message that expiry is optional, without need of any documentation. Having it as `time.Duration` means now caller/user will need to read documentation to understand how will it behaves.

jondonas (approved):

signer/signer.go
```diff
@@ -90,8 +90,9 @@ func (s *genericSigner) Sign(ctx context.Context, desc ocispec.Descriptor, opts
 		SigningAgent:  signingAgent, // TODO: include external signing plugin's name and version. https://github.com/notaryproject/notation-go/issues/80
 	}
 
-	if !opts.Expiry.IsZero() {
-		signReq.Expiry = opts.Expiry
+	// Add expiry only if ExpiryDuration is not nil and its value is not zero
+	if opts.ExpiryDuration != nil && opts.ExpiryDuration.Seconds() != 0 {
```
shizhMSFT, line 94: nit:
```suggestion
	if opts.ExpiryDuration != nil && *opts.ExpiryDuration != 0 {
```

signer/signer.go
shizhMSFT: In the golang conversion, `time.Duration` is more preferable than `*time.Duration`. As a golang library / module, it is better to be consistent with the built-in libraries.

shizhMSFT: Additionally, for library code, taking `*time.Duration` may cause usability issue. 

For instance,
```go
const expiryForCertianApp time.Duration = 30 * 24 * time.Hour // one month

signOpts := notation.SignOptions{
	ArtifactReference:  opts.reference,
	SignatureMediaType: mediaType,
	ExpiryDuration:     &expiryForCertianApp,
	PluginConfig:       pluginConfig,
}
```
The above code will have compile error since you cannot take an address of a const. However, constant expiry duration is common.

Instead of the above code, developers have to do
```go
expriyDuration := expiryForCertianApp
signOpts := notation.SignOptions{
	ArtifactReference:  opts.reference,
	SignatureMediaType: mediaType,
	ExpiryDuration:     &expriyDuration,
	PluginConfig:       pluginConfig,
}
```
which does not sound right.

priteshbandi: I have seen golang code use pointers for time.Duration in case where absence of value has a meaning ([ref1]( https://github.com/grpc/grpc-go/blob/v1.51.0/internal/serviceconfig/serviceconfig.go#L139-L145), [ref2](https://github.com/grpc/grpc-go/blob/master/xds/internal/xdsclient/xdsresource/type_rds.go#L137), [ref3](https://github.com/kubernetes-sigs/controller-runtime/blob/c83076e9f79258f23bcb200cf2391cf4ade939cf/pkg/config/v1alpha1/types.go#L56)). Yes, there is some inconvenience while writing code but having cleaner API/method definition will reduce potential bugs.

priteshbandi: As per discussion updated PR to use `time.Duration`

patrickzheng200 (approved):
LGTM. Looks like there's a unit test failed.

shizhMSFT (approved):
LGTM

PR #223, opened 2022-12-01
feat: added tag reference log for notation.Sign and notation.Verify

patrickzheng200:
This PR adds tag reference logs in notation.Sign and notation.Verify.

### DIFF
--- a/notation.go
+++ b/notation.go
@@ -201,7 +201,6 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, re
 		return ocispec.Descriptor{}, nil, ErrorSignatureRetrievalFailed{Msg: err.Error()}
 	}
 	if !isDigestReference(artifactRef) {
-		fmt.Println("here")
 		// artifactRef is not a digest reference
 		ref, err := orasRegistry.ParseReference(artifactRef)
 		if err != nil {
--- a/notation.go
+++ b/notation.go
@@ -57,16 +57,16 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, opts Sig
 	logger := log.GetLogger(ctx)
 
 	artifactRef := opts.ArtifactReference
-	targetDesc, err := repo.Resolve(ctx, artifactRef)
-	if err != nil {
-		return ocispec.Descriptor{}, err
-	}
 	ref, err := orasRegistry.ParseReference(artifactRef)
 	if err != nil {
 		return ocispec.Descriptor{}, err
 	}
 	if ref.Reference == "" {
 		return ocispec.Descriptor{}, errors.New("reference is missing digest or tag")
+	}
+	targetDesc, err := repo.Resolve(ctx, artifactRef)
+	if err != nil {
+		return ocispec.Descriptor{}, err
 	}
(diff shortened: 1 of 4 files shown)


notation.go
```diff
@@ -262,3 +291,13 @@ func generateAnnotations(signerInfo *signature.SignerInfo) (map[string]string, e
 		annotationX509ChainThumbprint: string(val),
 	}, nil
 }
+
+func isDigestReference(reference string) bool {
```
shizhMSFT, line 295: Instead doing your own way, you may try [Reference.ValidateReferenceAsDigest()](https://pkg.go.dev/oras.land/oras-go/v2@v2.0.0-rc.5/registry#Reference.ValidateReferenceAsDigest)

notation.go
```diff
@@ -52,10 +55,23 @@ type Signer interface {
...
 	if err != nil {
 		return ocispec.Descriptor{}, err
 	}
+	if !isDigestReference(artifactRef) {
+		// artifactRef is not a digest reference
+		ref, err := orasRegistry.ParseReference(artifactRef)
+		if err != nil {
+			return ocispec.Descriptor{}, err
+		}
+		logger.Warnf("Always sign the artifact using digest(`@sha256:...`) rather than a tag(`:%s`) because tags are mutable and a tag reference can point to a different artifact than the one signed", ref.ReferenceOrDefault())
```
shizhMSFT, line 71: `oras-go` will not use the default `latest`. Therefore, it should be `ref.Reference` instead of `ref.ReferenceOrDefault()`.

notation.go
```diff
@@ -52,10 +54,26 @@ type Signer interface {
...
+	if err != nil {
+		return ocispec.Descriptor{}, err
+	}
+	ref, err := orasRegistry.ParseReference(artifactRef)
 	if err != nil {
 		return ocispec.Descriptor{}, err
 	}
+	if ref.Reference == "" {
+		return ocispec.Descriptor{}, errors.New("reference is missing digest or tag")
+	}
```
shizhMSFT, line 70: We can do this check before `repo.Resolve` to fail fast.

notation.go
```diff
@@ -176,16 +196,30 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, re
...
 	if err != nil {
 		return ocispec.Descriptor{}, nil, ErrorSignatureRetrievalFailed{Msg: err.Error()}
 	}
+	ref, err := orasRegistry.ParseReference(artifactRef)
+	if err != nil {
+		return ocispec.Descriptor{}, nil, ErrorSignatureRetrievalFailed{Msg: err.Error()}
+	}
+	if ref.Reference == "" {
+		return ocispec.Descriptor{}, nil, ErrorSignatureRetrievalFailed{Msg: "reference is missing digest or tag"}
+	}
```
shizhMSFT, line 211: same here

notation.go
patrickzheng200: > We can do this check before `repo.Resolve` to fail fast.

Oh yeah, nice catch!

rgnote (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #228, opened 2022-12-06
Improving debug log for plugin

priteshbandi:
Improving debug log for plugin

### DIFF
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -164,14 +164,11 @@ func run(ctx context.Context, pluginName string, pluginPath string, req proto.Re
 	// execute request
 	stdout, stderr, err := executor.Output(ctx, pluginPath, req.Command(), data)
 	if err != nil {
-		logger.Debugf("Plugin %s returned error", req.Command())
-		if exitError, ok := err.(*exec.ExitError); ok {
-			logger.Debugf("Plugin %s exit code: %s", req.Command(), exitError.ExitCode())
-		}
+		logger.Debugf("plugin %s execution status: %v", req.Command(), err)
+		logger.Debugf("Plugin %s returned error: %s", req.Command(), string(stderr))
 		var re proto.RequestError
 		jsonErr := json.Unmarshal(stderr, &re)
 		if jsonErr != nil {
-			logger.Debugf("Plugin %s error response: %s", req.Command(), jsonErr)
 			return proto.RequestError{
 				Code: proto.ErrorCodeGeneric,
 				Err:  fmt.Errorf("response is not in JSON format. error: %v stderr: %v", err, stderr)}
@@ -179,7 +176,7 @@ func run(ctx context.Context, pluginName string, pluginPath string, req proto.Re
 		return re
 	}
 
-	logger.Debugf("Plugin %s response: %+v", req.Command(), string(stdout))
+	logger.Debugf("Plugin %s response: %s", req.Command(), string(stdout))
 	// deserialize response
 	err = json.Unmarshal(stdout, resp)
 	if err != nil {
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -84,8 +84,6 @@ func NewCLIPlugin(ctx context.Context, name, path string) (*CLIPlugin, error) {
 
 // GetMetadata returns the metadata information of the plugin.
 func (p *CLIPlugin) GetMetadata(ctx context.Context, req *proto.GetMetadataRequest) (*proto.GetMetadataResponse, error) {
-	logger := log.GetLogger(ctx)
-	logger.Debugf("Plugin get-plugin-metadata request: %+v", req)
 	var metadata proto.GetMetadataResponse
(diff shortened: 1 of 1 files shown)


plugin/plugin.go
```diff
@@ -152,29 +148,34 @@ func (p *CLIPlugin) GenerateEnvelope(ctx context.Context, req *proto.GenerateEnv
...
 func run(ctx context.Context, pluginName string, pluginPath string, req proto.Request, resp interface{}) error {
+	logger := log.GetLogger(ctx)
+
 	// serialize request
 	data, err := json.Marshal(req)
 	if err != nil {
 		return fmt.Errorf("%s: failed to marshal request object: %w", pluginName, err)
 	}
 
+	logger.Debugf("Plugin %s request: %+v", req.Command(), string(data))
```
JeyJeyGao, line 170: Some requests contain binary data and we may don't want to log them.

priteshbandi: Why not? Our spec says binary data would be base64 encoded.

plugin/plugin.go
```diff
@@ -98,90 +96,90 @@ func (p *CLIPlugin) GetMetadata(ctx context.Context, req *proto.GetMetadataReque
...
 		if jsonErr != nil {
+			logger.Debugf("Plugin %s error response: %s", req.Command(), jsonErr)
 			return proto.RequestError{
 				Code: proto.ErrorCodeGeneric,
 				Err:  fmt.Errorf("response is not in JSON format. error: %v stderr: %v", err, stderr)}
 		}
 		return re
 	}
 
+	logger.Debugf("Plugin %s response: %+v", req.Command(), string(stdout))
```
shizhMSFT, line 182:
```suggestion
	logger.Debugf("Plugin %s response: %s", req.Command(), string(stdout))
```

plugin/plugin.go
```diff
@@ -98,90 +96,90 @@ func (p *CLIPlugin) GetMetadata(ctx context.Context, req *proto.GetMetadataReque
...
 	stdout, stderr, err := executor.Output(ctx, pluginPath, req.Command(), data)
 	if err != nil {
+		logger.Debugf("Plugin %s returned error", req.Command())
+		if exitError, ok := err.(*exec.ExitError); ok {
+			logger.Debugf("Plugin %s exit code: %s", req.Command(), exitError.ExitCode())
+		}
 		var re proto.RequestError
 		jsonErr := json.Unmarshal(stderr, &re)
 		if jsonErr != nil {
+			logger.Debugf("Plugin %s error response: %s", req.Command(), jsonErr)
```
shizhMSFT, line 174:
```suggestion
			logger.Debugf("Plugin %s error response: %v", req.Command(), jsonErr)
```

plugin/plugin.go
```diff
@@ -98,90 +96,90 @@ func (p *CLIPlugin) GetMetadata(ctx context.Context, req *proto.GetMetadataReque
...
 		return fmt.Errorf("%s: failed to marshal request object: %w", pluginName, err)
 	}
 
+	logger.Debugf("Plugin %s request: %s", req.Command(), string(data))
 	// execute request
 	stdout, stderr, err := executor.Output(ctx, pluginPath, req.Command(), data)
 	if err != nil {
+		logger.Debugf("Plugin %s returned error", req.Command())
+		if exitError, ok := err.(*exec.ExitError); ok {
+			logger.Debugf("Plugin %s exit code: %s", req.Command(), exitError.ExitCode())
```
shizhMSFT, line 169:
```suggestion
			logger.Debugf("Plugin %s exit code: %d", req.Command(), exitError.ExitCode())
```

plugin/plugin.go
shizhMSFT: or may be `%w`.

plugin/plugin.go
```diff
@@ -98,90 +96,90 @@ func (p *CLIPlugin) GetMetadata(ctx context.Context, req *proto.GetMetadataReque
...
-	logger.Debugf("Plugin describe-key request: %+v", req)
-	var resp proto.DescribeKeyResponse
 	if req.ContractVersion == "" {
 		req.ContractVersion = proto.ContractVersion
 	}
+
+	var resp proto.DescribeKeyResponse
 	err := run(ctx, p.name, p.path, req, &resp)
-	logger.Debugf("Plugin describe-key response: %+v", resp)
 	return &resp, err
```
shizhMSFT, line 112: Should we do
```go
var resp proto.DescribeKeyResponse
if err := run(...); err != nil {
    return nil, err
}
return &resp, nil
```
and for all other commands?

priteshbandi: IMO current format is more compact.

shizhMSFT: But not in a good code style and not memory efficient.

shizhMSFT (approved):
LGTM

patrickzheng200 (approved):
LGTM

PR #235, opened 2022-12-19
update: logs and error messages

patrickzheng200:
This PR adds more logs to the library and improves error messages.

This PR intends to resolve following issues:
https://github.com/notaryproject/notation/issues/462
https://github.com/notaryproject/notation/issues/128 -> comments related to trust policy

### DIFF
--- a/verifier/trustpolicy/trustpolicy.go
+++ b/verifier/trustpolicy/trustpolicy.go
@@ -153,6 +153,11 @@ type SignatureVerification struct {
 // Validate validates a policy document according to its version's rule set.
 // if any rule is violated, returns an error
 func (policyDoc *Document) Validate() error {
+	// sanity check
+	if policyDoc == nil {
+		return errors.New("trust policy document cannot be nil")
+	}
+
 	// Validate Version
 	if policyDoc.Version == "" {
 		return errors.New("trust policy document is missing or has empty version, it must be specified")
--- a/verifier/trustpolicy/trustpolicy.go
+++ b/verifier/trustpolicy/trustpolicy.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"regexp"
 	"strings"
 
@@ -150,10 +151,13 @@ type SignatureVerification struct {
 	Override          map[ValidationType]ValidationAction `json:"override,omitempty"`
 }
 
-// Validate validates a policy document according to it's version's rule set.
+// Validate validates a policy document according to its version's rule set.
 // if any rule is violated, returns an error
 func (policyDoc *Document) Validate() error {
 	// Validate Version
+	if policyDoc.Version == "" {
+		return errors.New("trust policy document is missing or has empty version, it must be specified")
+	}
 	if !slices.Contains(supportedPolicyVersions, policyDoc.Version) {
 		return fmt.Errorf("trust policy document uses unsupported version %q", policyDoc.Version)
 	}
@@ -173,10 +177,10 @@ func (policyDoc *Document) Validate() error {
 		}
 		policyStatementNameCount[statement.Name]++
 
-		// Verify signature verification level is valid
+		// Verify signature verification is valid
 		verificationLevel, err := statement.SignatureVerification.GetVerificationLevel()
 		if err != nil {
-			return fmt.Errorf("trust policy statement %q uses invalid signatureVerification value %q", statement.Name, statement.SignatureVerification.VerificationLevel)
+			return fmt.Errorf("trust policy statement %q has invalid signatureVerification with error: %v", statement.Name, err)
 		}
 
 		// Any signature verification other than "skip" needs a trust store and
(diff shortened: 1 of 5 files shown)


verifier/verifier.go
```diff
@@ -119,11 +122,14 @@ func (v *verifier) Verify(ctx context.Context, desc ocispec.Descriptor, signatur
 	payload := &envelope.Payload{}
 	err = json.Unmarshal(outcome.EnvelopeContent.Payload.Content, payload)
 	if err != nil {
+		logger.Debug("Failed to unmarshal the payload content in the signature blob to envelope.Payload")
```
shizhMSFT, line 125: Should it be an error level warning?
```suggestion
		logger.Error("Failed to unmarshal the payload content of the signature envelope")
```

shizhMSFT (approved):
LGTM with suggestions

verifier/verifier.go
patrickzheng200: Nice catch!

JeyJeyGao (approved):
LGTM

verifier/trustpolicy/trustpolicy.go
```diff
@@ -150,10 +151,13 @@ type SignatureVerification struct {
 	Override          map[ValidationType]ValidationAction `json:"override,omitempty"`
 }
 
-// Validate validates a policy document according to it's version's rule set.
+// Validate validates a policy document according to its version's rule set.
 // if any rule is violated, returns an error
 func (policyDoc *Document) Validate() error {
 	// Validate Version
+	if policyDoc.Version == "" {
```
shizhMSFT, line 158: Can `policyDoc` be `nil`?

verifier/trustpolicy/trustpolicy.go
```diff
@@ -173,10 +177,10 @@ func (policyDoc *Document) Validate() error {
 		}
 		policyStatementNameCount[statement.Name]++
 
-		// Verify signature verification level is valid
+		// Verify signature verification is valid
 		verificationLevel, err := statement.SignatureVerification.GetVerificationLevel()
 		if err != nil {
-			return fmt.Errorf("trust policy statement %q uses invalid signatureVerification value %q", statement.Name, statement.SignatureVerification.VerificationLevel)
+			return fmt.Errorf("trust policy statement %q has invalid signatureVerification with error: %v", statement.Name, err)
```
shizhMSFT, line 183: Do we need to specify `%w` instead of `%v` for errors?

verifier/trustpolicy/trustpolicy.go
```diff
@@ -263,22 +267,29 @@ func LoadDocument() (*Document, error) {
 	policyDocument := &Document{}
 	err = json.NewDecoder(jsonFile).Decode(policyDocument)
 	if err != nil {
-		return nil, err
+		if errors.Is(err, io.EOF) {
+			return nil, errors.New("empty trustpolicy.json file. See a valid trustpolicy.json example at https://github.com/notaryproject/notaryproject/blob/main/specs/trust-store-trust-policy.md#trust-policy")
+		}
+		return nil, errors.New("invalid trustpolicy.json file. See a valid trustpolicy.json example at https://github.com/notaryproject/notaryproject/blob/main/specs/trust-store-trust-policy.md#trust-policy")
```
shizhMSFT, line 273: Instead of `main`, we should use a fixed version tag.

verifier/trustpolicy/trustpolicy.go
```diff
@@ -263,22 +267,29 @@ func LoadDocument() (*Document, error) {
 	policyDocument := &Document{}
 	err = json.NewDecoder(jsonFile).Decode(policyDocument)
 	if err != nil {
-		return nil, err
+		if errors.Is(err, io.EOF) {
+			return nil, errors.New("empty trustpolicy.json file. See a valid trustpolicy.json example at https://github.com/notaryproject/notaryproject/blob/main/specs/trust-store-trust-policy.md#trust-policy")
+		}
```
shizhMSFT, line 272: I have concern that `io.EOF` does not imply empty file. Are there any other situations that may return `io.EOF`? May be `early EOF`? or `early EOF` will just be a generic json decoder errror?

verifier/trustpolicy/trustpolicy.go
patrickzheng200: It's checked before the Validate() method is called. Thus, in notation's workflow, it's safe. However, since it's a public method, applications that want to import the trustpolicy package and call Validate() directly shall check nil on their side. Otherwise, a panic will be triggered.

verifier/trustpolicy/trustpolicy.go
patrickzheng200: I replaced it with `v1.0.0-rc.1`.

verifier/trustpolicy/trustpolicy.go
patrickzheng200: Good point. I guess we could just remove this check? since it's one of the invalid file types anyway.

verifier/trustpolicy/trustpolicy.go
shizhMSFT: Let's check it.

verifier/trustpolicy/trustpolicy.go
```diff
@@ -173,10 +181,10 @@ func (policyDoc *Document) Validate() error {
 		}
 		policyStatementNameCount[statement.Name]++
 
-		// Verify signature verification level is valid
+		// Verify signature verification is valid
 		verificationLevel, err := statement.SignatureVerification.GetVerificationLevel()
 		if err != nil {
-			return fmt.Errorf("trust policy statement %q uses invalid signatureVerification value %q", statement.Name, statement.SignatureVerification.VerificationLevel)
+			return fmt.Errorf("trust policy statement %q has invalid signatureVerification with error: %w", statement.Name, err)
```
priteshbandi, line 187:
```suggestion
			return fmt.Errorf("trust policy statement %q has invalid signatureVerification. Error: %w", statement.Name, err)
```

priteshbandi (approved):
LGTM with suggestions

PR #236, opened 2022-12-21
Fix error message

priteshbandi:
Fix error message: Instead of printing bytes in error message convert them to string.


rgnote (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #241, opened 2023-01-12
feat: support OCI image manifest

patrickzheng200:
This PR is the implementation of the [spec](https://github.com/notaryproject/notaryproject/blob/main/specs/signature-specification.md#backward-compatibility).

Tested with [COSE, JWS] + [OCI artifact manifest, OCI image manifest] on ACR.

### DIFF
--- a/notation.go
+++ b/notation.go
@@ -252,7 +252,7 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, re
 				break
 			}
 			numOfSignatureProcessed++
-			logger.Infof("Processing signature with mediaType: %v and digest: %v", sigManifestDesc.MediaType, sigManifestDesc.Digest)
+			logger.Infof("Processing signature with manifest mediaType: %v and digest: %v", sigManifestDesc.MediaType, sigManifestDesc.Digest)
 			// get signature envelope
 			sigBlob, sigDesc, err := repo.FetchSignatureBlob(ctx, sigManifestDesc)
 			if err != nil {
--- a/notation.go
+++ b/notation.go
@@ -252,7 +252,7 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, re
 				break
 			}
 			numOfSignatureProcessed++
-			logger.Infof("Processing signature with digest: %v", sigManifestDesc.Digest)
+			logger.Infof("Processing signature with mediaType: %v and digest: %v", sigManifestDesc.MediaType, sigManifestDesc.Digest)
 			// get signature envelope
 			sigBlob, sigDesc, err := repo.FetchSignatureBlob(ctx, sigManifestDesc)
 			if err != nil {
--- a/notation.go
+++ b/notation.go
@@ -42,17 +42,6 @@ type SignOptions struct {
 
 	// SigningAgent sets the signing agent name
 	SigningAgent string
-
-	// OCIimageManifest is the flag to specify if user wants to use OCI image
-	// manifest to store signatures in remote registries.
-	// By default, Notation will use OCI artifact manifest to store signatures.
-	// If OCIimageManifest flag is set to true, Notation will instead use
-	// OCI image manifest.
-	// Note, Notation will not automatically convert between these two types
-	// on any occasion.
-	// OCI artifact manifest: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/artifact.md
-	// OCI image manifest: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md
-	OCIimageManifest bool
 }
 
 // Signer is a generic interface for signing an artifact.
@@ -107,7 +96,7 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, opts Sig
 	}
 	logger.Debugf("Generated annotations: %+v", annotations)
 	logger.Debugf("Pushing signature of artifact descriptor: %+v, signature media type: %v", targetDesc, opts.SignatureMediaType)
-	_, _, err = repo.PushSignature(ctx, opts.SignatureMediaType, sig, targetDesc, annotations, opts.OCIimageManifest)
+	_, _, err = repo.PushSignature(ctx, opts.SignatureMediaType, sig, targetDesc, annotations)
 	if err != nil {
 		logger.Error("Failed to push the signature")
 		return ocispec.Descriptor{}, err
--- a/notation.go
+++ b/notation.go
@@ -42,6 +42,17 @@ type SignOptions struct {
 
 	// SigningAgent sets the signing agent name
 	SigningAgent string
+
+	// OCIImageManifest is the flag to specify if user wants to use OCI image
+	// manifest to store signatures in remote registries.
+	// By default, Notation will use OCI artifact manifest to store signatures.
+	// If OCIImageManifest flag is set to true, Notation will instead use
+	// OCI image manifest to store signatures.
+	// Note, Notation will not automatically convert between these two types
+	// on any occasion.
+	// OCI artifact manifest: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/artifact.md
+	// OCI image manifest: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md
+	OCIImageManifest bool
 }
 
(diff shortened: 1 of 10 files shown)


notation.go
```diff
@@ -42,6 +42,17 @@ type SignOptions struct {
...
+	// OCIimageManifest is the flag to specify if user wants to use OCI image
+	// manifest to store signatures in remote registries.
+	// By default, Notation will use OCI artifact manifest to store signatures.
+	// If OCIimageManifest flag is set to true, Notation will instead use
+	// OCI image manifest.
+	// Note, Notation will not automatically convert between these two types
+	// on any occasion.
+	// OCI artifact manifest: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/artifact.md
+	// OCI image manifest: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md
+	OCIimageManifest bool
```
shizhMSFT, line 55: To push the signature using OCI image manifest or OCI artifact manifest is not related to the `signer.Sign()` itself. We should introduce `RemoteSignOptions` like `RemoteVerifyOptions`. However, this solution is still not appropriate since whether a repository supports OCI artifact manifest is a property of the `registry.Repository`.

Therefore, I would like to propose the following changes in the package `registry`:
```go
// repositoryClient implements Repository
type repositoryClient struct {
	registry.Repository
	RepositoryOptions 
}

type RepositoryOptions struct {
	PushImageManifest bool
}

func NewRepositoryWithOptions(repo registry.Repository, opts RepositoryOptions) Repository {
	return &repositoryClient{
		Repository:        repo,
		RepositoryOptions: opts,
	}
}

```

registry/repository.go
```diff
@@ -63,48 +63,62 @@ func (c *repositoryClient) FetchSignatureBlob(ctx context.Context, desc ocispec.
...
 	opts := oras.PackOptions{
 		Subject:             &subject,
 		ManifestAnnotations: annotations,
 	}
 
-	return oras.Pack(ctx, c.Repository.Manifests(), ArtifactTypeNotation, []ocispec.Descriptor{blobDesc}, opts)
+	// use OCI image manifest to store signatures
+	if ociImageManifest {
+		opts.PackImageManifest = true
+	}
```
shizhMSFT, line 121: nit:
```go
opts := oras.PackOptions{
	Subject:             &subject,
	ManifestAnnotations: annotations,
	PackImageManifest:   ociImageManifest,
}
```

registry/repository.go
```diff
@@ -42,14 +42,14 @@ func (c *repositoryClient) ListSignatures(ctx context.Context, desc ocispec.Desc
...
 	if err != nil {
 		return nil, ocispec.Descriptor{}, err
 	}
-	if len(sigManifest.Blobs) != 1 {
-		return nil, ocispec.Descriptor{}, fmt.Errorf("signature manifest requries exactly one signature envelope blob, got %d", len(sigManifest.Blobs))
+	if len(signatureBlobs) != 1 {
+		return nil, ocispec.Descriptor{}, fmt.Errorf("signature manifest requries exactly one signature envelope blob, got %d", len(signatureBlobs))
 	}
-	sigDesc := sigManifest.Blobs[0]
+	sigDesc := signatureBlobs[0]
```
shizhMSFT, line 52: Should this block of code in `getSignatureBlobDesc()` since `getSignatureBlobDesc()` should return just one descriptor by its method name?

registry/repository.go
```diff
@@ -16,15 +16,31 @@ const (
...
 // repositoryClient implements Repository
 type repositoryClient struct {
 	registry.Repository
+	RepositoryOptions
 }
 
-// NewRepository returns a new Repository
-func NewRepository(repo registry.Repository) Repository {
+// NewRepositoryWithOptions returns a new Repository.
+func NewRepositoryWithOptions(repo registry.Repository, opts RepositoryOptions) Repository {
```
shizhMSFT, line 40: We may still want to keep the original constructor so that we will have 2 constructors.

registry/repository.go
```diff
@@ -16,15 +16,31 @@ const (
...
+	// OCIimageManifest specifies if user wants to use OCI image manifest
+	// to store signatures in remote registries.
+	// By default, Notation will use OCI artifact manifest to store signatures.
+	// If OCIimageManifest flag is set to true, Notation will instead use
+	// OCI image manifest.
+	// Note, Notation will not automatically convert between these two types
+	// on any occasion.
+	// OCI artifact manifest: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/artifact.md
+	// OCI image manifest: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md
+	OCIimageManifest bool
```
shizhMSFT, line 30:
```suggestion
	OCIImageManifest bool
```

registry/repository.go
```diff
@@ -69,42 +81,59 @@ func (c *repositoryClient) PushSignature(ctx context.Context, mediaType string,
...
+	return signatureBlobs[0], nil
 }
 
 // uploadSignatureManifest uploads the signature manifest to the registry
-func (c *repositoryClient) uploadSignatureManifest(ctx context.Context, subject, blobDesc ocispec.Descriptor, annotations map[string]string) (ocispec.Descriptor, error) {
+func (c *repositoryClient) uploadSignatureManifest(ctx context.Context, subject, blobDesc ocispec.Descriptor, annotations map[string]string, ociImageManifest bool) (ocispec.Descriptor, error) {
 	opts := oras.PackOptions{
 		Subject:             &subject,
 		ManifestAnnotations: annotations,
+		PackImageManifest:   ociImageManifest,
```
shizhMSFT, line 135: `ociImageManifest` can be accessed directly from `c.OCIImageManifest` instead of the function parameter.

registry/repository.go
patrickzheng200: In that case, do we still need the `RepositoryOptions`? Looks like `OCIImageManifest == false` has the same behavior as the original constructor. So an alternative might be:
```
// repositoryClient implements Repository
type repositoryClient struct {
	registry.Repository
	OCIImageManifest bool
}

func NewRepository(repo registry.Repository) Repository {
	return &repositoryClient{
		Repository: repo,
	}
}

func NewRepositoryWithImageManifest(repo registry.Repository) Repository {
	return &repositoryClient{
		Repository:       repo,
		OCIImageManifest: true,
	}
}
```

registry/repository.go
patrickzheng200: Nice catch!

registry/repository.go
shizhMSFT: The constructor
```go
func NewRepository(repo registry.Repository) Repository
```
should be used for most users with all default settings.

The constructor
```go
func NewRepositoryWithOptions(repo registry.Repository, opts RepositoryOptions) Repository
```
should be used for advanced users. Since we may introduce more options in the future, `RepositoryOptions` is designed for extensibility.

patrickzheng200: I see. Will update the code.

shizhMSFT (approved):
LGTM

registry/repository_test.go
```diff
@@ -287,12 +302,34 @@ func (c mockRemoteClient) Do(req *http.Request) (*http.Response, error) {
 		default:
 			return &http.Response{}, fmt.Errorf(msg)
 		}
+	case "/v2/test/referrers/sha256:0000000000000000000000000000000000000000000000000000000000000000":
+		return &http.Response{
+			StatusCode: http.StatusOK,
+			Body:       io.NopCloser(bytes.NewReader([]byte(validPageImage))),
+			Header: map[string][]string{
+				"Oras-Api-Version": {"oras/1.1"},
+			},
```
shizhMSFT, line 311: We don't support ORAS artifact spec any longer. Should we remove those ORAS artifact spec related code?

registry/repository_test.go
```diff
@@ -68,6 +69,17 @@ const (
 			}
 		]
 	}`
+	validPageImage = `
+	{
+		"references": [{}],
```
shizhMSFT, line 74: Do we need this `references` field?

patrickzheng200: Code updated.

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

PR #242, opened 2023-01-13
feat: add support for signed user metadata

byronchien:
Adds support for signed user metadata in `notation sign` and `notation verify`. [Relevant spec](https://github.com/notaryproject/notation/pull/498)

example sign usage:
```
chienb@a07817b52895 notation % notation sign $IMAGE --user-metadata io.wabbit-networks.buildId=123 --user-metadata io.wabbit-networks.buildTime=123
Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(`:v1`) because tags are mutable and a tag reference can point to a different artifact than the one signed.
Resolved artifact tag `v1` to digest `sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b` before signing.
Successfully signed localhost:5000/net-monitor@sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b

```

example metadata displayed on verification (without metadata flag)
```
chienb@a07817b52895 notation % notation verify $IMAGE
Resolved artifact tag `v1` to digest `sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b` before verification.
Warning: The resolved digest may not point to the same signed artifact, since tags are mutable.
Successfully verified signature for localhost:5000/net-monitor@sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b

The artifact was signed with the following user metadata.

KEY                          VALUE
io.wabbit-networks.buildTime   123
io.wabbit-networks.buildId     123
```

example verification:
```
chienb@a07817b52895 notation % notation verify $IMAGE --user-metadata io.wabbit-networks.buildTime=123
Resolved artifact tag `v1` to digest `sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b` before verification.
Warning: The resolved digest may not point to the same signed artifact, since tags are mutable.
Successfully verified signature for localhost:5000/net-monitor@sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b

The artifact was signed with the following user metadata.

KEY                            VALUE
io.wabbit-networks.buildTime   123
io.wabbit-networks.buildId     123
```

example verification failure
```
chienb@a07817b52895 notation % notation verify $IMAGE --user-metadata foo=bar
Resolved artifact tag `v1` to digest `sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b` before verification.
Warning: The resolved digest may not point to the same signed artifact, since tags are mutable.
Error: signature verification failed: signature verification failed
```

Signed-off-by: Byron Chien <chienb@amazon.com>


byronchien:
related: https://github.com/notaryproject/notation/pull/507

verifier/verifier.go
```diff
@@ -135,18 +135,27 @@ func (v *verifier) Verify(ctx context.Context, desc ocispec.Descriptor, signatur
...
 	outcome.VerificationResults = append(outcome.VerificationResults, integrityResult)
 	if integrityResult.Error != nil {
 		logVerificationResult(logger, integrityResult)
 		return integrityResult.Error
 	}
 
+	// verify user metadata
+	if len(userMetadata) > 0 {
+		logger.Info("User Metadata flag is present. Checking signature metadata for specified values.")
+		err := verifyUserMetadata(logger, envContent, userMetadata)
```
priteshbandi, line 153: Should this happen after below code? https://github.com/notaryproject/notation-go/blob/6c88d3de1283d0d573bafae9d1567026367bfcb6/verifier/verifier.go#L130-L134?

In that way we can avoid parsing payload twice and only perform annotation check if descriptor is correct

verifier/verifier.go
```diff
@@ -412,6 +421,28 @@ func verifyAuthenticity(ctx context.Context, trustPolicy *trustpolicy.TrustPolic
...
+	if err != nil {
+		logger.Error("Failed to unmarshal the payload content in the signature blob to envelope.Payload")
+		return err
+	}
+
+	logger.Debugf("Signature metadata: %v", payload.TargetArtifact.Annotations)
+
+	for k, v := range userMetadata {
+		if payload.TargetArtifact.Annotations[k] != v {
+			logger.Debug("Error: specified metadata is not present in the signature")
```
priteshbandi, line 438: I would add information which key value pair is not present in the signature.

verifier/verifier.go
```diff
@@ -412,6 +420,20 @@ func verifyAuthenticity(ctx context.Context, trustPolicy *trustpolicy.TrustPolic
 	}
 }
 
+func verifyUserMetadata(logger log.Logger, payload *envelope.Payload, userMetadata map[string]string) error {
+	logger.Debugf("Verifying that metadata %v is present in signature", userMetadata)
+	logger.Debugf("Signature metadata: %v", payload.TargetArtifact.Annotations)
+
+	for k, v := range userMetadata {
+		if payload.TargetArtifact.Annotations[k] != v {
+			logger.Infof("Error: specified metadata %s=%s is not present in the signature", k, v)
```
patrickzheng200, line 429:
```suggestion
			logger.Errorf("User required metadata %s=%s is not present in the signature", k, v)
```

notation.go
```diff
@@ -40,6 +41,9 @@ type SignOptions struct {
 	// PluginConfig sets or overrides the plugin configuration.
 	PluginConfig map[string]string
 
+	// UserMetadata contains key-value pairs that are added to the signature payload
+	UserMetadata map[string]string
```
shizhMSFT, line 45: [ocispec.Descirptor](https://pkg.go.dev/github.com/opencontainers/image-spec/specs-go/v1#Descriptor) already has an field of `Annotations`. Should we reuse it?

notation.go
```diff
@@ -156,6 +160,9 @@ type VerifyOptions struct {
 
 	// PluginConfig is a map of plugin configs.
 	PluginConfig map[string]string
+
+	// UserMetadata contains key-value pairs that must be present in the signature
+	UserMetadata map[string]string
```
shizhMSFT, line 165: [ocispec.Descirptor](https://pkg.go.dev/github.com/opencontainers/image-spec/specs-go/v1#Descriptor) already has an field of `Annotations`. Should we reuse it?

notation.go
```diff
@@ -181,6 +188,9 @@ type RemoteVerifyOptions struct {
 	// will be processed for verification. If set to less than or equals
 	// to zero, an error will be returned.
 	MaxSignatureAttempts int
+
+	// UserMetadata contains key-value pairs that must be present in the signature
+	UserMetadata map[string]string
```
shizhMSFT, line 193: We need to keep this option if we reuse `Annotations` of [ocispec.Descirptor](https://pkg.go.dev/github.com/opencontainers/image-spec/specs-go/v1#Descriptor).

signer/signer.go
```diff
@@ -23,6 +24,13 @@ import (
 // signingAgent is the unprotected header field used by signature.
 const signingAgent = "Notation/1.0.0"
 
+var (
+	reservedAnnotationPrefixes = []string{
+		"org.opencontainers",
```
shizhMSFT, line 29: Why `org.opencontainer` is a reserved prefix?

Users may want those [standard annotations](https://pkg.go.dev/github.com/opencontainers/image-spec/specs-go/v1#pkg-constants).

notation.go
byronchien: i don't think it makes sense for verify since a descriptor is not passed into verify, but will change for sign

notation.go
byronchien: in both cases, the descriptor is fetched in notation.Sign/notation.Verify.
- i think adding annotations to the descriptor before calling signer.Sign makes sense
- for verify, the artifact descriptor passed in is being validated, so re-using the annotations there to pass the user-metadata that we're looking for doesn't make sense

shizhMSFT: Since `notation-go` is a library, developers can invoke `Signer.Sign` and `Verifier.Verify` directly.

As you have mentioned, `notation.Sign` and `notation.Verify` does not have an entrypoint for user metadata. Hence, we do need an option for the user metadata. For `Verify`, we can add it in the `RemoteVerifyOptions`. For `Sign`, we may introduce a new struct `RemoteSignOptions` just like `RemoteVerifyOptions`.

priteshbandi: @byronchien, I might have mislead you.
I agree with @shizhMSFT, we are already passing a descriptor for the `Sign` method so we don't need a new metadata variable in `notation.SignOption`. Instead, what can pass a descriptor with metadata already filled in as annotations.

signer/signer.go
```diff
@@ -133,3 +144,27 @@ func (s *genericSigner) Sign(ctx context.Context, desc ocispec.Descriptor, opts
 	// TODO: re-enable timestamping https://github.com/notaryproject/notation-go/issues/78
 	return sig, &envContent.SignerInfo, nil
 }
+
+func constructPayload(ctx context.Context, desc ocispec.Descriptor, userMetadata map[string]string) (*envelope.Payload, error) {
+	logger := log.GetLogger(ctx)
+
+	if desc.Annotations == nil && len(userMetadata) > 0 {
+		desc.Annotations = map[string]string{}
+	}
```
priteshbandi, line 153: Why are we overriding existing annotations instead we should append new ones?

notation.go
```diff
@@ -240,6 +251,7 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, re
 	}
 
 	var verificationOutcomes []*VerificationOutcome
+	var failedOutcomes []*VerificationOutcome
```
priteshbandi, line 254: Why does failedOutcomes needs to be an array? Can we use use a boolean value to check if verification was successful but metadata match failed?

byronchien: i guess it doesn't necessarily need to be an array. my reasoning was that it might be easier to prioritize different error messages based on different failure outcomes in the future (not just a metadata failure), but this can also just be a single boolean/message that's set if the outcome error is metadata verification failure.

notation.go
byronchien: ok, i think i understand now, will add a RemoteSignOptions, remove usermetadata from SignOptions, and add the annotations to the descriptor before Signer.Sign

notation.go
priteshbandi: Since we dont have check on number of signatures user can verify adding an unbounded array will aggravate the unbounded memory consumption issue.

signer/signer.go
byronchien: this is initializing annotations to an empty map if the existing descriptor has a nil annotation map to avoid trying to write to a nil map.

notation.go
```diff
@@ -22,6 +24,31 @@ import (
...
+	// supported.
+	SignatureMediaType string
+
+	// ExpiryDuration identifies the expiry duration of the resulted signature. Zero value
+	// represents no expiry duration.
+	ExpiryDuration time.Duration
+
+	// PluginConfig sets or overrides the plugin configuration.
+	PluginConfig map[string]string
+
```
priteshbandi, line 45: Can we embed SignOptions into RemoteSignOptions?

you can do is embed SignOptions inside RemoteSignOptions.
E.g.

notation.go
```diff
@@ -243,6 +317,9 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, re
 	errExceededMaxVerificationLimit := ErrorVerificationFailed{Msg: fmt.Sprintf("total number of signatures associated with an artifact should be less than: %d", remoteOpts.MaxSignatureAttempts)}
 	numOfSignatureProcessed := 0
 
+	var verificationFailedErr error
+	verificationFailedErr = ErrorVerificationFailed{}
```
priteshbandi, line 321:
```suggestion
	var verificationFailedErr error = ErrorVerificationFailed{}
```

priteshbandi (approved):
LGTM

notation.go
```diff
@@ -325,3 +390,17 @@ func generateAnnotations(signerInfo *signature.SignerInfo) (map[string]string, e
...
+	if err != nil {
+		return nil, errors.New("Failed to unmarshal the payload content in the signature blob to envelope.Payload")
+	}
+
+	if payload.TargetArtifact.Annotations == nil {
+		return map[string]string{}, nil
+	}
+
+	return payload.TargetArtifact.Annotations, nil
+}
```
shizhMSFT, line 406: nit: code style: This code block should be inserted at line 196 right after the definition of the `VerificationOutcome` struct

notation.go
```diff
@@ -325,3 +390,17 @@ func generateAnnotations(signerInfo *signature.SignerInfo) (map[string]string, e
 		annotationX509ChainThumbprint: string(val),
 	}, nil
 }
+
+func (outcome *VerificationOutcome) GetUserMetadata() (map[string]string, error) {
```
shizhMSFT, line 394: According to [effective go](https://go.dev/doc/effective_go#Getters), the `Get` prefix should be removed.
```suggestion
func (outcome *VerificationOutcome) UserMetadata() (map[string]string, error) {
```

notation.go
```diff
@@ -325,3 +390,17 @@ func generateAnnotations(signerInfo *signature.SignerInfo) (map[string]string, e
...
+	var payload envelope.Payload
+	err := json.Unmarshal(outcome.EnvelopeContent.Payload.Content, &payload)
```
shizhMSFT, line 396: In case of a bad outcome, `outcome.EnvelopeContent` can be `nil` and thus will panic.

verifier/verifier.go
```diff
@@ -412,6 +420,20 @@ func verifyAuthenticity(ctx context.Context, trustPolicy *trustpolicy.TrustPolic
 	}
 }
 
+func verifyUserMetadata(logger log.Logger, payload *envelope.Payload, userMetadata map[string]string) error {
+	logger.Debugf("Verifying that metadata %v is present in signature", userMetadata)
+	logger.Debugf("Signature metadata: %v", payload.TargetArtifact.Annotations)
+
+	for k, v := range userMetadata {
+		if payload.TargetArtifact.Annotations[k] != v {
```
shizhMSFT, line 428: In case of an edge case, users may require a field to be present but empty.
```suggestion
		if got, ok := payload.TargetArtifact.Annotations[k]; !ok || got != v {
```

example_remoteSign_test.go
```diff
@@ -46,9 +46,9 @@ func Example_remoteSign() {
 	exampleRepo := registry.NewRepository(remoteRepo)
 
 	// exampleSignOptions is an example of notation.SignOptions.
-	exampleSignOptions := notation.SignOptions{
-		ArtifactReference:  exampleArtifactReference,
-		SignatureMediaType: exampleSignatureMediaType,
+	exampleSignOptions := notation.RemoteSignOptions{
+	    ArtifactReference: exampleArtifactReference,
+	    SignatureMediaType: exampleSignatureMediaType
```
patrickzheng200, line 51:
```suggestion
	    SignatureMediaType: exampleSignatureMediaType,
```

patrickzheng200 (approved):
LGTM with one comment

notation.go
```diff
@@ -194,6 +185,24 @@ type VerificationOutcome struct {
 	Error error
 }
 
+func (outcome *VerificationOutcome) UserMetadata() (map[string]string, error) {
+	if outcome.EnvelopeContent == nil {
+		return map[string]string{}, nil
```
shizhMSFT, line 190: If `outcome.EnvelopeContent` is nil, should we return the outcome error instead of an empty map?

yizha1:
@byronchien  could you fix the conflicts?

byronchien:
resolved conflict

shizhMSFT (approved):
LGTM

PR #245, opened 2023-01-19
update: support OCI image manifest

patrickzheng200:
This PR targets on [issue 219](https://github.com/notaryproject/notaryproject/issues/219) and [issue 220](https://github.com/notaryproject/notaryproject/issues/220).

In this update, Notation will store signatures of type OCI artifact manifest only when:
     1. The registry supports OCI artifact manifest.
     2. The registry supports the Referrers API.

### DIFF
--- a/registry/repository.go
+++ b/registry/repository.go
@@ -84,6 +84,18 @@ func (c *repositoryClient) FetchSignatureBlob(ctx context.Context, desc ocispec.
 // linked signature envelope blob. Upon successful, PushSignature returns
 // signature envelope blob and manifest descriptors.
 func (c *repositoryClient) PushSignature(ctx context.Context, mediaType string, blob []byte, subject ocispec.Descriptor, annotations map[string]string) (blobDesc, manifestDesc ocispec.Descriptor, err error) {
+	// when uploading OCI artifact manifest, Notation requires the registry
+	// to support the Referrers API as well.
+	// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers
+	if !c.OCIImageManifest {
+		err := c.Repository.Referrers(ctx, subject, "", func(referrers []ocispec.Descriptor) error {
+			return nil
+		})
+		if err != nil {
+			return ocispec.Descriptor{}, ocispec.Descriptor{}, fmt.Errorf("failed to ping Referrers API on uploading OCI artifact manifest with error: %v. Try OCI image manifest instead", err)
+		}
+	}
+
 	blobDesc, err = oras.PushBytes(ctx, c.Repository.Blobs(), mediaType, blob)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


registry/repository.go
```diff
@@ -143,5 +143,17 @@ func (c *repositoryClient) uploadSignatureManifest(ctx context.Context, subject,
...
+	// to support the Referrers API as well.
+	// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers
+	if !c.OCIImageManifest {
+		err := c.Repository.Referrers(ctx, subject, "", func(referrers []ocispec.Descriptor) error {
+			return nil
+		})
+		if err != nil {
+			return ocispec.Descriptor{}, fmt.Errorf("failed to ping Referrers API on uploading OCI artifact manifest with error: %v. Try OCI image manifest instead", err)
+		}
+	}
```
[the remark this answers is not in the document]
patrickzheng200: Moved the code in front of line 87 to early break.

registry/repository.go
```diff
@@ -84,6 +84,18 @@ func (c *repositoryClient) FetchSignatureBlob(ctx context.Context, desc ocispec.
 // linked signature envelope blob. Upon successful, PushSignature returns
 // signature envelope blob and manifest descriptors.
 func (c *repositoryClient) PushSignature(ctx context.Context, mediaType string, blob []byte, subject ocispec.Descriptor, annotations map[string]string) (blobDesc, manifestDesc ocispec.Descriptor, err error) {
+	// when uploading OCI artifact manifest, Notation requires the registry
+	// to support the Referrers API as well.
+	// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers
+	if !c.OCIImageManifest {
+		err := c.Repository.Referrers(ctx, subject, "", func(referrers []ocispec.Descriptor) error {
```
shizhMSFT, line 91: It is better to use a non-existing descriptor for better performance.

registry/repository.go
```diff
@@ -84,6 +84,18 @@ func (c *repositoryClient) FetchSignatureBlob(ctx context.Context, desc ocispec.
...
+			return nil
+		})
+		if err != nil {
+			return ocispec.Descriptor{}, ocispec.Descriptor{}, fmt.Errorf("failed to ping Referrers API on uploading OCI artifact manifest with error: %v. Try OCI image manifest instead", err)
+		}
+	}
```
shizhMSFT, line 97: I'm actually thinking if this logic should be in `notation-go` or in the `notation CLI`.

patrickzheng200: As we discussed, we still will have this overhead if we move this logic to the CLI side. Hence, I'll keep it here for now.

patrickzheng200:
As discussed, will move this logic to CLI side. PR closed.

PR #246, opened 2023-01-22
feat!: add signingkeys.json validation check

priteshbandi:
Change to validate signingkeys are valid(unique name) for both read and write operation.


byronchien (approved):

config/keys.go
```diff
@@ -65,5 +71,27 @@ func LoadSigningKeys() (*SigningKeys, error) {
...
+	}
+
 	return &config, nil
 }
+
+func validateKeys(config *SigningKeys) error {
+	keys := config.Keys
+	uniqMap := make(map[string]struct{}, len(keys))
+	for _, key := range keys {
+		if _, ok := uniqMap[key.Name]; ok {
```
JeyJeyGao, line 86: Do we need to check an empty string key name?

config/keys.go
```diff
@@ -65,5 +71,27 @@ func LoadSigningKeys() (*SigningKeys, error) {
...
+			return fmt.Errorf("malformed %s: multiple keys with name '%s' found", dir.PathSigningKeys, key.Name)
+		}
+		uniqMap[key.Name] = struct{}{}
+	}
+
+	if _, ok := uniqMap[config.Default]; !ok {
```
JeyJeyGao, line 92: If default key is not specified, it is also an valid signingkeys.json file?

priteshbandi: That's correct, default key is optional.

config/keys.go
priteshbandi: We don't have that in specification but it's a good point, should we restrict it to alpha numeric?

config/keys.go
```diff
@@ -65,5 +71,27 @@ func LoadSigningKeys() (*SigningKeys, error) {
...
+	if err := validateKeys(&config); err != nil {
+		return nil, err
+	}
+
 	return &config, nil
 }
+
+func validateKeys(config *SigningKeys) error {
+	keys := config.Keys
+	uniqMap := make(map[string]struct{}, len(keys))
```
patrickzheng200, line 84: We do have a [set](https://github.com/notaryproject/notation-go/blob/main/internal/container/set.go) data structure implemented in notation-go.

config/keys.go
patrickzheng200, line 92: IIRC, config.Default can be an empty string in the signingkeys.json file. So, missing config.Default should be fine here?

priteshbandi: Fixed

config/keys.go
```diff
@@ -65,5 +72,37 @@ func LoadSigningKeys() (*SigningKeys, error) {
...
+
+	if config.Default != nil {
+		defaultKey := *config.Default
+		if len(defaultKey) == 0 {
+			return fmt.Errorf("malformed %s: default key name cannot be empty", dir.PathSigningKeys)
+		}
+
+		if !uniqueKeyNames.Contains(defaultKey) {
+		return fmt.Errorf("malformed %s: default key '%s' not found", dir.PathSigningKeys, defaultKey)
+		}
```
patrickzheng200, line 104: This logic prevents a deletion of the default key in signingkeys.json. For example, if the sigingkeys.json is:
```
{
    "default": "alpine2",
    "keys": [
        {
            "name": "alpine",
            "keyPath": "dir/alpine.key",
            "certPath": "dir/alpine.crt"
        },
        {
            "name": "alpine2",
            "keyPath": "dir/alpine2.key",
            "certPath": "dir/alpine2.crt"
        }
    ]
}
```
A `notation key delete alpine2` command on the CLI side will fail the `validateKeys` because in the process of deleting the default key `alpine2`, the default key field would be set as `""`.  But we expect the command to be successful with the result:
```
{
    "default": "",
    "keys": [
        {
            "name": "alpine",
            "keyPath": "dir/alpine.key",
            "certPath": "dir/alpine.crt"
        }
    ]
}
```

priteshbandi: yes, that's correct will update notation to assign nil value to default instead of empty string

patrickzheng200: > yes, that's correct will update notation to assign nil value to default instead of empty string

Oh, I see. Just an open discussion, is there any particular reason that we should do that? Since we do not allow empty key name, an empty string for `default` field means `default` is not set. For library code, if we set `Default` as a `*string` instead of `string`, users might need extra work to fill it. (for example, iirc, in golang, we cannot take address of a const.)

config/keys.go
```diff
@@ -65,5 +72,37 @@ func LoadSigningKeys() (*SigningKeys, error) {
...
+	}
+
+	if config.Default != nil {
+		defaultKey := *config.Default
+		if len(defaultKey) == 0 {
+			return fmt.Errorf("malformed %s: default key name cannot be empty", dir.PathSigningKeys)
+		}
+
+		if !uniqueKeyNames.Contains(defaultKey) {
+		return fmt.Errorf("malformed %s: default key '%s' not found", dir.PathSigningKeys, defaultKey)
```
patrickzheng200, line 103: nit:
this line needs indentation

priteshbandi: Actually refactored the code to no directly manupulate SigningKeys, instead added getters and setters for it.

config/keys.go
priteshbandi: Actually refactored the code to not directly manipulate SigningKeys struct, instead added getters and setters for it.

config/keys.go
```diff
@@ -29,29 +36,151 @@ type KeySuite struct {
...
+		ExternalKey: &ExternalKey{
+			ID:           id,
+			PluginName:   pluginName,
+			PluginConfig: pluginConfig,
+		},
+	}
+
+	err =  s.addCore(ks, markDefault)
+	logger.Debugf("Added key with name %s - {%+v}", keyName, ks)
+	return err
```
patrickzheng200, line 107: When err is not nil, the log still claims "Added key with name...". Should we do something like:
```suggestion
	if err :=  s.addCore(ks, markDefault); err != nil {
	    logger.Errorf(xxx)
	    return err
	}
	logger.Debugf("Added key with name %s - {%+v}", keyName, ks)
	return nil
```

config/keys.go
```diff
@@ -29,29 +36,151 @@ type KeySuite struct {
...
+func (s *SigningKeys) GetDefault() (KeySuite, error) {
+	if s.Default == nil {
+		return KeySuite{}, errors.New("default signing key not set." +
+			" Please set default singing key or specify a key name")
+	}
+
+	return s.Get(*s.Default)
+}
+
+// Remove deletes given signing keys
```
patrickzheng200, line 134: We should add a brief description for the outputs of this func.

config/keys.go
```diff
@@ -29,29 +36,151 @@ type KeySuite struct {
...
+		return errorKeyNameEmpty
+	}
+
+	if !slices.ContainsIsser(s.Keys, keyName) {
+		return fmt.Errorf("key with name '%s' not found", keyName)
+	}
+
+	if s.Default == nil || *s.Default != keyName {
+		s.Default = &keyName
+	}
```
patrickzheng200, line 167: This can simply be:
```suggestion
	s.Default = &keyName
```

config/keys.go
```diff
@@ -29,29 +36,151 @@ type KeySuite struct {
...
+
+	return nil
+}
+
 // Save config to file
```
patrickzheng200, line 172: How about:
```suggestion
// Save SigningKeys to signingkeys.json
```

config/keys.go
```diff
@@ -29,29 +36,151 @@ type KeySuite struct {
...
-func NewSigningKeys() *SigningKeys {
-	return &SigningKeys{Keys: []KeySuite{}}
+	if err := validateKeys(s); err != nil {
+		return err
+	}
+
+	return save(path, s)
 }
 
 // LoadSigningKeys reads the config from file
```
patrickzheng200, line 186: Same here:
```suggestion
// LoadSigningKeys reads the signingkeys.json file
```

patrickzheng200 (approved):
LGTM. We could wait for Shiwei in case he has extra comments on the PR.

config/keys.go
```diff
@@ -29,32 +36,154 @@ type KeySuite struct {
...
 
-// Is checks whether the given name is equal with the Name variable
-func (k KeySuite) Is(name string) bool {
-	return k.Name == name
-}
+var errorKeyNameEmpty = errors.New("key name cannot be empty")
+var errKeyNotFound = errors.New("signing key not found")
 
-// SigningKeys reflects the signingkeys.json file.
+// SigningKeys reflects the signings.json file.
```
patrickzheng200, line 42: nit: the name of the file should be `signingkeys.json` right?

config/keys.go
```diff
@@ -65,5 +194,74 @@ func LoadSigningKeys() (*SigningKeys, error) {
...
+
+	if err := fn(signingKeys); err != nil {
+		return err
+	}
+
+	if err := signingKeys.Save(); err != nil {
+		return err
+	}
+
+	return nil
```
shizhMSFT, line 221: nit: code can be simplified
```suggestion
	return signingKeys.Save()
```

config/keys.go
```diff
@@ -65,5 +194,74 @@ func LoadSigningKeys() (*SigningKeys, error) {
...
+
+	if err := validateKeys(&config); err != nil {
+		return nil, err
+	}
+
 	return &config, nil
 }
+
+// LoadExecSaveSigningKeys loads signing key, executes given function and then saves the signing key
+func LoadExecSaveSigningKeys(fn func(keys *SigningKeys) error) error {
```
shizhMSFT, line 206: Will `UpdateSigningKeys` a better API name for this utility function?

priteshbandi: We already have methods called UpdateDefualt, Remove which updates the `SigningKeys` struct but actully doesnt saves them on disk, IMO its better to  have `LoadExecSaveSigningKeys` which shows what function is doing.

config/keys.go
```diff
@@ -29,32 +36,154 @@ type KeySuite struct {
...
-func (k KeySuite) Is(name string) bool {
-	return k.Name == name
-}
+var errorKeyNameEmpty = errors.New("key name cannot be empty")
+var errKeyNotFound = errors.New("signing key not found")
 
 // SigningKeys reflects the signingkeys.json file.
 type SigningKeys struct {
-	Default string     `json:"default"`
+	Default *string    `json:"default"`
```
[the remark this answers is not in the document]
priteshbandi: Fix the issue by adding `omit,empty`

internal/slices/slices.go
```diff
@@ -19,3 +19,29 @@ func ContainsAny(s []any, v any) bool {
...
+// ContainsIsser reports whether name is present in s.
+func ContainsIsser[E isser](s []E, name string) bool {
+	return IndexIsser(s, name) >= 0
+}
+
+// DeleteIssr removes the elements s[i:i+1] from s,
+// returning the modified slice.
+func DeleteIssr[S ~[]E, E isser](s S, i int) S {
+return append(s[:i], s[i+1:]...)
+}
```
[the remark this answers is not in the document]
priteshbandi: I did try that but was getting following error
`config/keys.go:149:25: []KeySuite does not implement ~[]any ([]KeySuite missing in ~[]any)`

shizhMSFT (approved):
LGTM

PR #253, opened 2023-01-29
fix: Appends annotations returned by plugin to signature manifest's annotations

priteshbandi:
As per spec([link](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#generate-envelope)) annotations returned by plugin should be appened to signature manifest's annotations and this PR introduces aforementioned behavior

### DIFF
--- a/notation.go
+++ b/notation.go
@@ -49,7 +49,7 @@ type SignOptions struct {
 // and packing in various signature formats.
 type Signer interface {
 	// Sign signs the artifact described by its descriptor,
-	// and returns the signature and SignerInfo.
+	// and returns the signature, annotations and SignerInfo.
 	Sign(ctx context.Context, desc ocispec.Descriptor, opts SignOptions) ([]byte, map[string]string, *signature.SignerInfo, error)
 }
 
@@ -85,12 +85,12 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, opts Sig
 		logger.Infof("Resolved artifact tag `%s` to digest `%s` before signing", ref.Reference, targetDesc.Digest.String())
 	}
 
-	sig, ants, signerInfo, err := signer.Sign(ctx, targetDesc, opts)
+	sig, pluginAnnotations, signerInfo, err := signer.Sign(ctx, targetDesc, opts)
 	if err != nil {
 		return ocispec.Descriptor{}, err
 	}
 	logger.Debug("Generating annotation")
-	annotations, err := generateAnnotations(signerInfo, ants)
+	annotations, err := generateAnnotations(signerInfo, pluginAnnotations)
 	if err != nil {
 		return ocispec.Descriptor{}, err
 	}
--- a/notation.go
+++ b/notation.go
@@ -50,7 +50,7 @@ type SignOptions struct {
 type Signer interface {
 	// Sign signs the artifact described by its descriptor,
 	// and returns the signature and SignerInfo.
-	Sign(ctx context.Context, desc ocispec.Descriptor, opts SignOptions) ([]byte, *signature.SignerInfo, error)
+	Sign(ctx context.Context, desc ocispec.Descriptor, opts SignOptions) ([]byte, map[string]string, *signature.SignerInfo, error)
 }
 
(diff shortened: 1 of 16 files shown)


notation.go
```diff
@@ -50,7 +50,7 @@ type SignOptions struct {
 type Signer interface {
 	// Sign signs the artifact described by its descriptor,
 	// and returns the signature and SignerInfo.
```
patrickzheng200, line 52: nit:
we should add a description for the newly added output term as well.

notation.go
```diff
@@ -85,12 +85,12 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, opts Sig
 		logger.Infof("Resolved artifact tag `%s` to digest `%s` before signing", ref.Reference, targetDesc.Digest.String())
 	}
 
-	sig, signerInfo, err := signer.Sign(ctx, targetDesc, opts)
+	sig, ants, signerInfo, err := signer.Sign(ctx, targetDesc, opts)
```
patrickzheng200, line 88: Looks like `ants` is non-nil only when signer is pluginSigner with generate-envelope capability. How about renaming it to something like `pluginAnnotations`?

signer/plugin.go
```diff
@@ -63,26 +65,26 @@ func (s *pluginSigner) Sign(ctx context.Context, desc ocispec.Descriptor, opts n
...
-func (s *pluginSigner) generateSignature(ctx context.Context, desc ocispec.Descriptor, opts notation.SignOptions, metadata *proto.GetMetadataResponse) ([]byte, *signature.SignerInfo, error) {
+func (s *pluginSigner) generateSignature(ctx context.Context, desc ocispec.Descriptor, opts notation.SignOptions, metadata *proto.GetMetadataResponse) ([]byte, map[string]string, *signature.SignerInfo, error) {
 	logger := log.GetLogger(ctx)
 	logger.Debug("Generating signature by plugin")
 	config := s.mergeConfig(opts.PluginConfig)
 	// Get key info.
 	key, err := s.describeKey(ctx, config)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil,err
```
byronchien, line 78: supernit: space 
```suggestion
		return nil, nil, nil, err
```

byronchien (approved):

notation.go
```diff
@@ -49,8 +49,8 @@ type SignOptions struct {
 // and packing in various signature formats.
 type Signer interface {
 	// Sign signs the artifact described by its descriptor,
-	// and returns the signature and SignerInfo.
-	Sign(ctx context.Context, desc ocispec.Descriptor, opts SignOptions) ([]byte, *signature.SignerInfo, error)
+	// and returns the signature, annotations and SignerInfo.
+	Sign(ctx context.Context, desc ocispec.Descriptor, opts SignOptions) ([]byte, map[string]string, *signature.SignerInfo, error)
```
patrickzheng200, line 53: Just an open discussion, another design option here is creating a new unexported interface, something like:
```
in notation.go:
    type signerAnnotation interface {
        PluginAnnotations(xxx) map[string]string
    }
    in func Sign(xxx) xxx {
         do a type assertion on the signer input, if it's a signerAnnotation, get its plugin annotations.
    }
in signer/plugin.go:
    type pluginSigner struct {
	plugin       plugin.SignPlugin
	keyID        string
	pluginConfig map[string]string
        pluginAnnotations map[string]string
    }
    // pluginSigner will implement notation.signerAnnotation
    func (s *pluginSigner) PluginAnnotations(xxx) map[string]string {
        return s.pluginAnnotations
    }
    inside func (s *pluginSigner) generateSignatureEnvelope(xxx) xxx {
         add logic: s.pluginAnnotations = resp.Annotations
    } 
```
In the above design, we are no longer breaking change the notation.Signer interface, and only pluginSigner is able to return annotations. (generic signer will not implement signerAnnotation as it doesn't generate annotation during the Sign process.) In the user's perspective, nothing really changes, the adding pluginAnnotations process happened under the hood.

priteshbandi:
Closing in favor of https://github.com/notaryproject/notation-go/pull/262

Issue #254, opened 2023-01-29
trust store validation error message needs improvement


JeyJeyGao:
The trust store validation error message needs improvement. Here is the [code](https://github.com/notaryproject/notation-go/blob/c90d5213aefafbe60a361398447217fad018a1aa/verifier/helpers.go#L58).

```go
storeType, name, found := strings.Cut(trustStore, ":")
if !found {
	return nil, fmt.Errorf("trust policy statement %q is missing separator in trust store value %q", policy.Name, trustStore)
}
```
We had better provide a valid trust store format to help user figure out the value. We can add "Format <TrustStoreType>:<TrustStoreName> is required" to the end of the error message.

patrickzheng200:
To align with the spec, we could say `Format {store-type}:{store-name} is required`

Issue #252, opened 2023-01-28
verification error log should be a warning level rather than info level


JeyJeyGao:
Here is the [code](https://github.com/notaryproject/notation-go/blob/c90d5213aefafbe60a361398447217fad018a1aa/notation.go#L268). 

```go
// verify each signature
outcome, err := verifier.Verify(ctx, artifactDescriptor, sigBlob, opts)
if err != nil {
        // this message should be an warning level log.
	logger.Infof("Signature %v failed verification with error: %v", sigManifestDesc.Digest, err)
	if outcome == nil {
		logger.Error("Got nil outcome. Expecting non-nil outcome on verification failure")
		return err
	}
	continue
}
```

Issue #251, opened 2023-01-28
trusted identity parsing error was not returned


JeyJeyGao:
As shown in [the code](https://github.com/notaryproject/notation-go/blob/c90d5213aefafbe60a361398447217fad018a1aa/internal/pkix/pkix.go#L16), the error was handled but didn't return the real error which can help the user locate the issue.

```go
	dn, err := ldapv3.ParseDN(name)
	if err != nil {
		return nil, fmt.Errorf("distinguished name (DN) %q is not valid, it must contain 'C', 'ST', and 'O' RDN attributes at a minimum, and follow RFC 4514 standard", name)
	}
```

patrickzheng200:
Nice catch! Will create a PR and add the corresponding logic.

PR #256, opened 2023-01-30
chore: logs and tests clean-up

patrickzheng200:
This PR targets on issues #251, #252 and #254.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 
 # Custom
 coverage.txt
+tmp
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@
 
 # Custom
 coverage.txt
-tmp
--- a/verifier/helpers.go
+++ b/verifier/helpers.go
@@ -55,7 +55,7 @@ func loadX509TrustStores(ctx context.Context, scheme signature.SigningScheme, po
 
 		storeType, name, found := strings.Cut(trustStore, ":")
 		if !found {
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -5,4 +5,5 @@
 *.sublime-workspace
 
 # Custom
-coverage.txt
\ No newline at end of file
+coverage.txt
+tmp
```
JeyJeyGao, line 9: Why we need it?

patrickzheng200: Running `go test -v .` at the root dir would generate a folder tmp for trust store due to the example_test files.

JeyJeyGao (approved):
LGTM

.gitignore
priteshbandi: Shouldn't we be cleaning that file after test run?

verifier/helpers.go
```diff
@@ -55,7 +55,7 @@ func loadX509TrustStores(ctx context.Context, scheme signature.SigningScheme, po
 
 		storeType, name, found := strings.Cut(trustStore, ":")
 		if !found {
-			return nil, fmt.Errorf("trust policy statement %q is missing separator in trust store value %q", policy.Name, trustStore)
+			return nil, fmt.Errorf("trust policy statement %q is missing separator in trust store value %q. Format <TrustStoreType>:<TrustStoreName> is required", policy.Name, trustStore)
```
priteshbandi, line 58: nit: Should we reword it to:
`trust policy statement %q is missing separator in trust store value %q. The required format is <TrustStoreType>:<TrustStoreName>`

priteshbandi (approved):
LGTM with a nit

shizhMSFT (approved):
LGTM

PR #261, opened 2023-02-02
Add a method for parsing the descriptor from a payload

byronchien:
Adds a helper to get the descriptor from the payload. Needed here instead of in notation-core-go since the payload needs to be parsed to envelope.Payload (in notation-go internal)

Signed-off-by: Byron Chien <chienb@amazon.com>


notation.go
```diff
@@ -325,3 +326,18 @@ func generateAnnotations(signerInfo *signature.SignerInfo) (map[string]string, e
 		annotationX509ChainThumbprint: string(val),
 	}, nil
 }
+
+func GetDescriptorFromPayload(payload *signature.Payload) (*ocispec.Descriptor, error) {
```
patrickzheng200, line 330: We need a func doc for this function because it's exported.

notation.go
```diff
@@ -325,3 +326,18 @@ func generateAnnotations(signerInfo *signature.SignerInfo) (map[string]string, e
...
+	if payload == nil {
+		return nil, errors.New("Empty payload")
```
patrickzheng200, line 332: nit:
```suggestion
		return nil, errors.New("empty payload")
```

notation.go
```diff
@@ -325,3 +326,18 @@ func generateAnnotations(signerInfo *signature.SignerInfo) (map[string]string, e
...
+	}
+
+	var parsedPayload envelope.Payload
+
+	err := json.Unmarshal(payload.Content, &parsedPayload)
+	if err != nil {
+		return nil, errors.New("Failed to unmarshall the payload content to envelope.Payload")
```
patrickzheng200, line 339: nit:
```suggestion
		return nil, errors.New("failed to unmarshall the payload content to envelope.Payload")
```

patrickzheng200 (commented):
Just an open discussion, where are we using this new method? Looks like it's a helper function but not used in notation.go.

byronchien:
> Just an open discussion, where are we using this new method? Looks like it's a helper function but not used in notation.go.

using this function in inspect ([PR here](https://github.com/notaryproject/notation/pull/528)). Added the helper in notation-go rather than notation since it relies on a type not exported from notation-go (envelope.Payload).

patrickzheng200:
> > Just an open discussion, where are we using this new method? Looks like it's a helper function but not used in notation.go.
> 
> using this function in inspect ([PR here](https://github.com/notaryproject/notation/pull/528)). Added the helper in notation-go rather than notation since it relies on a type not exported from notation-go (envelope.Payload).

I see. IMO, `notation.go` should only expose APIs related to the core processes such as Sign and Verify. As a library code, this is to keep `notation-go` as clean as possible. To add this helper function, notation CLI actually has its own [internal/envelope](https://github.com/notaryproject/notation/blob/main/internal/envelope/envelope.go) file. We could add envelope.Payload struct to this file and move this helper function into Notation CLI.

byronchien:
> > > Just an open discussion, where are we using this new method? Looks like it's a helper function but not used in notation.go.
> > 
> > 
> > using this function in inspect ([PR here](https://github.com/notaryproject/notation/pull/528)). Added the helper in notation-go rather than notation since it relies on a type not exported from notation-go (envelope.Payload).
> 
> I see. IMO, `notation.go` should only expose APIs related to the core processes such as Sign and Verify. As a library code, this is to keep `notation-go` as clean as possible. Notation CLI also has its own [internal/envelope](https://github.com/notaryproject/notation/blob/main/internal/envelope/envelope.go) file. We could add envelope.Payload struct to this file and move this helper function into Notation CLI. (Note: `notation CLI` is one of the users of `notation-go` library. there will be non-CLI users as well.)

makes sense, will add the struct and helper to notation instead

notation.go
```diff
@@ -325,3 +326,20 @@ func generateAnnotations(signerInfo *signature.SignerInfo) (map[string]string, e
 		annotationX509ChainThumbprint: string(val),
 	}, nil
 }
+
+// GetDescriptorFromPayload parses a signature payload and returns the descriptor
+// that was signed.
+func GetDescriptorFromPayload(payload *signature.Payload) (*ocispec.Descriptor, error) {
```
shizhMSFT, line 332: According to[ effective go](https://go.dev/doc/effective_go#Getters), `Get` prefix should be dropped.
```suggestion
func DescriptorFromPayload(payload *signature.Payload) (*ocispec.Descriptor, error) {
```

notation.go
```diff
@@ -325,3 +326,20 @@ func generateAnnotations(signerInfo *signature.SignerInfo) (map[string]string, e
 		annotationX509ChainThumbprint: string(val),
 	}, nil
 }
+
+// GetDescriptorFromPayload parses a signature payload and returns the descriptor
+// that was signed.
```
shizhMSFT, line 331: Note: the descriptor was signed but may not be trusted.

notation.go
shizhMSFT, line 332: The word `Payload` in `DescriptorFromPayload` is too broad. Consider `DescriptorFromSignaturePayload`.

notation.go
```diff
@@ -325,3 +326,20 @@ func generateAnnotations(signerInfo *signature.SignerInfo) (map[string]string, e
...
+func GetDescriptorFromPayload(payload *signature.Payload) (*ocispec.Descriptor, error) {
+	if payload == nil {
+		return nil, errors.New("empty payload")
+	}
+
+	var parsedPayload envelope.Payload
+
+	err := json.Unmarshal(payload.Content, &parsedPayload)
```
shizhMSFT, line 339: Security concern: payload media type is not checked before parsing.

byronchien: can parse the payload to an intermediate struct to check mediatype before parsing to the final payload struct.

byronchien:
added DescriptorFromSignaturePayload to notation in [#528](https://github.com/notaryproject/notation/pull/528)

PR #262, opened 2023-02-02
fix: Appends annotations returned by plugin to signature manifest's annotations

priteshbandi:
Another way to implement https://github.com/notaryproject/notation-go/pull/253 as per suggestion from @patrickzheng200 

-------
As per spec([link](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#generate-envelope)) annotations returned by plugin should be appened to signature manifest's annotations and this PR introduces aforementioned behavior


signer/plugin.go
```diff
@@ -21,12 +21,13 @@ import (
 // pluginSigner signs artifacts and generates signatures.
 // It implements notation.Signer
 type pluginSigner struct {
-	plugin       plugin.SignPlugin
-	keyID        string
-	pluginConfig map[string]string
+	plugin              plugin.SignPlugin
+	keyID               string
+	pluginConfig        map[string]string
+	manifestAnnotations map[string]string
```
patrickzheng200, line 27: nit: can we rename it to `pluginAnnotations`? since it's annotations from the plugin.

signer/plugin.go
```diff
@@ -44,6 +45,10 @@ func NewFromPlugin(plugin plugin.Plugin, keyID string, pluginConfig map[string]s
 	}, nil
 }
 
+func (s *pluginSigner) PluginAnnotations() map[string]string {
```
patrickzheng200, line 48: We need a func doc for this function and mention that `pluginSigner implements notation.signerAnnotation`.

patrickzheng200 (commented):
Looks like need to fix the unit tests, otherwise, LGTM.

patrickzheng200 (approved):
LGTM

rgnote (approved):
LGTM

PR #269, opened 2023-02-03
Create config with user only permission

priteshbandi:
Create config with user only permission


config/base.go
```diff
@@ -19,6 +19,11 @@ func save(filePath string, cfg interface{}) error {
 		return err
 	}
 	defer file.Close()
+
+	if err := os.Chmod(filePath, 0600); err != nil {
```
JeyJeyGao, line 23:
```suggestion
	if err := file.Chmod(0600); err != nil {
```
nit

config/base.go
```diff
@@ -19,6 +19,11 @@ func save(filePath string, cfg interface{}) error {
...
+		return err
+	}
```
shizhMSFT, line 25: Instead of creating and then chmod, please consider replace `os.Create` with `os.OpenFile(filePath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)`

config/keys.go
```diff
@@ -148,7 +148,7 @@ func (s *SigningKeys) Remove(keyName ...string) ([]string, error) {
 		}
 		s.Keys = slices.Delete(s.Keys, idx)
 		deletedNames = append(deletedNames, name)
-		if *s.Default == name {
+		if s.Default !=nil && *s.Default == name {
 			s.Default = nil
 		}
```
priteshbandi, line 153: This is actually a bug fix which I found while working on this PR

patrickzheng200 (approved):
LGTM

byronchien (approved):

PR #271, opened 2023-02-07
Update: added ErrorPushSignatureFailed

patrickzheng200:
This PR adds a `ErrorPushSignatureFailed` error type so that users of notation.Sign() would be able to check error type on Signing failures. 
On notation CLI side, if the error is of type `ErrorPushSignatureFailed`, a suggestion of using OCI image manifest will be printed out.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -1,5 +1,17 @@
 package notation
 
+//
+type ErrorPushSignatureFailed struct {
+	Msg string
+}
+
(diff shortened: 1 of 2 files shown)


JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

PR #279, opened 2023-02-20
feat: allow users overriding NOTATION_CONFIG and NOTATION_LIBEXEC

ningziwen:
https://github.com/notaryproject/notation/issues/473


shizhMSFT:
There is an on-going discussion at https://github.com/notaryproject/notation/issues/473

PR #285, opened 2023-03-07
fix: added truststore.ValidateCerts

### LINKED ISSUE #284: trust store certificates validation issue
https://github.com/notaryproject/notation-go/blob/9920fb73937e9b8dbaccc2886ceefefd5a4a3803/verifier/truststore/truststore.go#L107-L127

the `validateCerts` function was called in `X509TrustStore` interface which can be bypassed because the interface can be override by notation-go users. Then, it may cause an leaf cert can be used as a certificate in trust store and cause the verification only checks the leaf certificate.

patrickzheng200:
This PR aims to resolve #284.


verifier/helpers.go
```diff
@@ -34,6 +34,26 @@ var VerificationPluginHeaders = []string{
 
 var errExtendedAttributeNotExist = errors.New("extended attribute not exist")
 
+// ValidateCerts ensures certs from input are CA certificates or self-signed.
+// It's designed to be used when user implements their own
+// truststore.X509TrustStore.
+func ValidateCerts(certs []*x509.Certificate) error {
```
shizhMSFT, line 40: Should the following implementation also use this function to avoid duplicated code?
https://github.com/notaryproject/notation-go/blob/5cf34ace712920f8455857a5d7397ba49f4c00b1/verifier/truststore/truststore.go#L107

patrickzheng200: Reformed the code. Moved it into the truststore package as it's a function for validating certs in a trust store.

priteshbandi (approved):
LGTM

verifier/truststore/truststore.go
```diff
@@ -89,40 +87,35 @@ func (trustStore *x509TrustStore) GetCertificates(ctx context.Context, storeType
...
-func validateCerts(certs []*x509.Certificate, path string) error {
-	// to prevent any trust store misconfigurations, ensure there is at least
-	// one certificate from each file.
+// ValidateCerts ensures certs from trust store are CA certificates or
+// self-signed.
+func ValidateCerts(certs []*x509.Certificate, storeType Type, namedStore string) error {
 	if len(certs) < 1 {
-		return fmt.Errorf("could not parse a certificate from %q, every file in a trust store must have a PEM or DER certificate in it", path)
+		return fmt.Errorf("could not parse a certificate from %s/%s, every file in a trust store must have a PEM or DER certificate in it", storeType, namedStore)
 	}
```
shizhMSFT, line 106: The function `func ValidateCerts(certs []*x509.Certificate, storeType Type, namedStore string) error` is not generic enough than previous `func ValidateCertificateChain(certs []*x509.Certificate) error`. Also, other implementation may not read certificates from file so that PEM and DER may not be applicable.

verifier/truststore/truststore.go
```diff
@@ -89,40 +87,35 @@ func (trustStore *x509TrustStore) GetCertificates(ctx context.Context, storeType
 		if err != nil {
 			return nil, fmt.Errorf("error while reading certificates from %q: %w", joinedPath, err)
 		}
-
-		if err := validateCerts(certs, joinedPath); err != nil {
+		if err := ValidateCerts(certs, storeType, namedStore); err != nil {
```
shizhMSFT, line 90: How about

```go
func ValidateCertificateChain(certs []*x509.Certificate) error {
	if len(certs) < 1 {
		return errors.New("empty certificate chain")
	}

	for _, cert := range certs {
		if !cert.IsCA {
			if err := cert.CheckSignature(cert.SignatureAlgorithm, cert.RawTBSCertificate, cert.Signature); err != nil {
				return fmt.Errorf(
					"certificate with subject %q is not a CA certificate or self-signed signing certificate",
					cert.Subject,
				)
			}
		}
	}

	return nil
}
```

```suggestion
		// to prevent any trust store misconfigurations, ensure there is at least
		// one certificate from each file.
		if len(certs) < 1 {
			return fmt.Errorf("%s: could not parse a certificate from. every file in a trust store must have a PEM or DER certificate in it", joinedPath)
		}
		if err := ValidateCertificateChain(certs); err != nil {
			return fmt.Errorf("%s: %w", joinedPath, err)
		}
```

shizhMSFT (approved):
LGTM

priteshbandi (approved):

PR #294, opened 2023-03-28
chore: update err msg for policy verification

qweeah:
This PR rephrased error message returned during trust policy verification.

### DIFF
--- a/verifier/trustpolicy/trustpolicy.go
+++ b/verifier/trustpolicy/trustpolicy.go
@@ -186,7 +186,7 @@ func (policyDoc *Document) Validate() error {
 		// Verify signature verification is valid
 		verificationLevel, err := statement.SignatureVerification.GetVerificationLevel()
 		if err != nil {
-			return fmt.Errorf("trust policy statement %q has invalid signatureVerification. Error: %w", statement.Name, err)
+			return fmt.Errorf("trust policy statement %q has invalid signatureVerification: %w", statement.Name, err)
 		}
 
 		// Any signature verification other than "skip" needs a trust store and
@@ -289,7 +289,7 @@ func LoadDocument() (*Document, error) {
 // SignatureVerification struct throws error if SignatureVerification is invalid
 func (signatureVerification *SignatureVerification) GetVerificationLevel() (*VerificationLevel, error) {
 	if signatureVerification.VerificationLevel == "" {
-		return nil, errors.New("trust policy statement is missing or has empty signature verification level, it must be specified")
+		return nil, errors.New("signature verification configuration is missing or contains empty verification level, it must be specified")
 	}
 
 	var baseLevel *VerificationLevel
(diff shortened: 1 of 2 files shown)


patrickzheng200 (commented):
Have you checked the following?
1. run `go test -v .` under trustpolicy dir. All tests need to be passed.
2. run the notation CLI e2e test again, because there are test cases expecting this error message. Changing the error message would fail the e2e test.

qweeah:
> 2\. Changing the error message would fail the e2e test. All e2e tests need to be passed as well.

Shouldn't that be done when bumping notation-go?

patrickzheng200:
> > 2. Changing the error message would fail the e2e test. All e2e tests need to be passed as well.
> 
> Shouldn't that be done when bumping notation-go?

oh yeah, if you need to change the e2e, you could include the change in this PR: https://github.com/notaryproject/notation/pull/593 because this change is connected closely with your CLI trust policy command.
Merge the notation-go PR first -> go mod tidy on notation CLI side -> merge the notation CLI PR.

patrickzheng200 (dismissed):
LGTM

verifier/trustpolicy/trustpolicy.go
```diff
@@ -289,7 +289,7 @@ func LoadDocument() (*Document, error) {
 // SignatureVerification struct throws error if SignatureVerification is invalid
 func (signatureVerification *SignatureVerification) GetVerificationLevel() (*VerificationLevel, error) {
 	if signatureVerification.VerificationLevel == "" {
-		return nil, errors.New("trust policy statement is missing or has empty signature verification level, it must be specified")
+		return nil, errors.New("signature verification configuration is missing or contains empty verification level, it must be specified")
```
priteshbandi, line 292: This is first time I am seeing `signature verification configuration`, why do we want to move away from trust policy statement?

qweeah: The name comes from the corresponding field called `signatureVerification` in the [schema spec](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy-schema)

If we want to be accurate, how about changing the error message to `verification level cannot be empty`

qweeah: Why this change is needed: the validation error comes from the verification level, the original msg might make notation user think the statement itself is missing.

byronchien (approved):

patrickzheng200 (approved):
LGTM

priteshbandi (approved):
LGTM

shizhMSFT (approved):
LGTM

shizhMSFT:
I'm not able to merge this PR due to
image
caused by recent change by https://github.com/notaryproject/notation-core-go/pull/135

shizhMSFT:
Updated the branch policy to match the workflow defined in `notation-core-go`.

qweeah:
> @qweeah Have you checked the following?
> 
> 1. run `go test -v .` under trustpolicy dir. All tests need to be passed.
> 2. run the notation CLI e2e test again, because there are test cases expecting this error message. Changing the error message would fail the e2e test. All e2e tests need to be passed as well.

@patrickzheng200 I bumped notation-go including this change and ran an [E2E ](https://github.com/qweeah/notation/actions/runs/4636725254/jobs/8204886757) in my own fork, looks like no test will fail thus  no further change required on CLI side.

PR #295, opened 2023-03-29
feat: adding OCSP revocation checks to Verify

kody-kimberl:
This PR adds OCSP revocation checking to the Verify function using the notation-core-go's revocation package

Based on my design from https://github.com/notaryproject/notation-core-go/issues/132 and the specification [here](https://github.com/notaryproject/notaryproject/blob/main/specs/trust-store-trust-policy.md#certificate-revocation-evaluation).

This PR addresses part of the following issue:
 - https://github.com/notaryproject/notation-core-go/issues/124

It is dependent on https://github.com/notaryproject/notation-core-go/pull/134

Signed-off-by: Kody Kimberl [kody.kimberl.work@gmail.com](mailto:kody.kimberl.work@gmail.com)


verifier/verifier.go
```diff
@@ -63,6 +67,24 @@ func New(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore
...
+	result, err := New(trustPolicy, trustStore, pluginManager)
+	if err != nil {
+		return nil, err
+	}
+	if v, ok := result.(*verifier); !ok {
+		return result, nil
+	} else {
+		v.revocationClient = revocationClient
+		return v, nil
+	}
```
byronchien, line 84: i'm confused, why is this type check needed?

verifier/verifier.go
```diff
@@ -518,6 +545,52 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+			importantErr = true
+		case ocsp.NoOCSPServerError:
+			// Should this be an important error?
+			logger.Debugf("%v", t)
+			logger.Debug("1 or more certificates in the chain did not have OCSP configured. Ignoring these certificates")
+			importantErr = false
+		case ocsp.TimeoutError:
+			logger.Debugf("%v", t)
+			logger.Debug("Revocation unavailable. OCSP requests timed out")
+			importantErr = false
```
byronchien, line 573: is it right to fail open if revocation / ocsp can't be reached? this seems like an error that should fail the verification. users also have the option to fail open with trust policy level.

verifier/verifier.go
```diff
@@ -518,6 +545,52 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+		importantErr = false
+	} else {
+		// If there is an error and it is designated as unimportant, log but pass the validation
+		switch t := revocationErr.(type) {
+		case ocsp.RevokedError:
+			importantErr = true
+		case ocsp.UnknownStatusError:
+			importantErr = true
+		case ocsp.NoOCSPServerError:
+			// Should this be an important error?
```
byronchien, line 566: imo i think no ocsp being present should not be an important error.

verifier/verifier.go
kody-kimberl: The New function return type is notation.Verifier rather than the verifier struct that we know it should be. Thus, a cast is needed. It should never actually fail though

byronchien (approved):

byronchien (dismissed):

verifier/verifier.go
```diff
@@ -518,6 +544,40 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+		for _, err := range serverErrs {
+			if err != nil {
+				logger.Debugf("error for certificate #%d in chain: %v", (i + 1), err)
+			}
+		}
+	}
+
+	switch revocation.ResultFromErrors(revocationErrs) {
+	case revocation.OK:
+		logger.Debug("no important errors encountered while checking revocation, status is OK")
```
rgnote, line 570: Can we combine revocationErrs and the result (`revocation.*`) into a struct so that we don't need to make multiple calls to `revocation` package? Having a simpler package interface would be ideal.

verifier/verifier.go
rgnote: Can we get rid of the if-else block since it is causing confusion and simply assign the `revocationClient` to `v` and return. It is fine to check `ok` field and fail (though it never happens ideally)

verifier/verifier.go
kody-kimberl: If we did that, the notation-core-go functionality would be coupled to the way that we have specified for errors to be merged. The goal was to abstract the "result" from the individual errors. This was the consensus after a few different discussions and design changes in the notation-core-go PR. What is the downside of making these calls separately?

rgnote: I posted few comments on https://github.com/notaryproject/notation-core-go/pull/134 that are related to simplifying the OCSP status result as well.

> What is the downside of making these calls separately?

That increases the surface area of the package. Larger surface area means more maintenance burden.

verifier/verifier.go
```diff
@@ -63,6 +66,25 @@ func New(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore
 	}, nil
 }
 
+// NewWithRevocationClient creates a new verifier given trustPolicy, trustStore, pluginManager, and revocationClient
+func NewWithRevocationClient(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore, pluginManager plugin.Manager, revocationClient *http.Client) (notation.Verifier, error) {
+	if revocationClient == nil {
+		revocationClient = &http.Client{Timeout: 5 * time.Second}
+	}
```
priteshbandi, line 73: Why to default when you are taking revocationClient as input? Infact we should add this defualt in `New` method (revocation implementation is not optional). 
```suggestion
	if revocationClient == nil {
		return nil, errors.new('revocaton client is nil')
	}
```

verifier/verifier.go
```diff
@@ -63,6 +66,25 @@ func New(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore
...
+	result, err := New(trustPolicy, trustStore, pluginManager)
+	if err != nil {
+		return nil, err
+	}
+
+	v, ok := result.(*verifier)
+	if !ok {
+		// should never occur, but checking just in case
+		return nil, errors.New("unable to create verifier with client for revocation")
+	}
```
priteshbandi, line 83: Why do we need to do this? New should always return `notation.Verifier`

verifier/verifier.go
```diff
@@ -255,10 +277,15 @@ func (v *verifier) processSignature(ctx context.Context, sigBlob []byte, envelop
 	// check if we need to bypass the revocation check, since revocation can be
 	// skipped using a trust policy or a plugin may override the check
 	if outcome.VerificationLevel.Enforcement[trustpolicy.TypeRevocation] != trustpolicy.ActionSkip &&
-		!slices.Contains(pluginCapabilities, proto.CapabilityRevocationCheckVerifier) {
-		logger.Debugf("Validating revocation (not implemented)")
-		// TODO perform X509 revocation check (not in RC1)
-		// https://github.com/notaryproject/notation-go/issues/110
+		!slices.Contains(pluginCapabilities, proto.CapabilityRevocationCheckVerifier) &&
+		!slices.Contains(pluginCapabilities, proto.CapabilityTrustedIdentityVerifier) {
```
priteshbandi, line 281: Why do we care about `CapabilityTrustedIdentityVerifier` here?

verifier/verifier.go
```diff
@@ -518,6 +545,67 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
 	}
 }
 
+func verifyRevocation(outcome *notation.VerificationOutcome, client *http.Client, logger log.Logger) *notation.ValidationResult {
+	r, err := revocation.New(client)
+	if err != nil {
+		logger.Debug("no HTTP client provided for revocation checking")
```
priteshbandi, line 551: In future there can be other reasons so its better to get reason from `err`
```suggestion
		logger.Debug("error while calling revocation.New: %s , err.Error())
```

verifier/verifier.go
```diff
@@ -518,6 +545,67 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+		return &notation.ValidationResult{
+			Type:   trustpolicy.TypeRevocation,
+			Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeRevocation],
+			Error:  errors.New("unable to check revocation status, an HTTP client must be provided"),
```
priteshbandi, line 555: ditto

verifier/verifier.go
```diff
@@ -518,6 +545,67 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+		}
+	}
+
+	authenticSigningTime := outcome.EnvelopeContent.SignerInfo.SignedAttributes.SigningTime
+	// TODO use authenticSigningTime from signerInfo
+	// https://github.com/notaryproject/notation-core-go/issues/38
```
priteshbandi, line 561: IMO here it doesnt matter if its authentic signing time or a signing time which user added.

verifier/verifier.go
```diff
@@ -518,6 +545,67 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+	}
+
+	// Log all non-nil errors as debug
+	for i, certResult := range revocationResult {
+		for _, err := range certResult.ServerResults {
+			if err != nil {
+				logger.Debugf("error for certificate #%d in chain: %v", (i + 1), err)
+			}
+		}
+	}
```
priteshbandi, line 576: Can we combine this with line 579 to save iterating on all the revocationResult? IMO its okay to show only first failure and not show all the failures and save few cpu cycles.

verifier/verifier.go
```diff
@@ -518,6 +545,67 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+			currResult = revocation.Revoked
+			// will break out of loop after this
+		default:
+			// revocation.Unknown
+			currResult = revocation.Unknown
+			continue
+		}
+		if currResult == revocation.Revoked {
+			break
+		}
```
priteshbandi, line 593: Can we add break at line 585?

verifier/verifier.go
```diff
@@ -518,6 +545,67 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+	}
+
+	switch currResult {
+	case revocation.OK:
+		logger.Debug("no important errors encountered while checking revocation, status is OK")
+	case revocation.Revoked:
+		result.Error = errors.New("one or more certificates in the chain were determined to be revoked, status is Revoked")
```
priteshbandi, line 600:
```suggestion
		result.Error = errors.New("signing certificate(s) revoked")
```

verifier/verifier.go
```diff
@@ -518,6 +545,67 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+	default:
+		// revocation.Unknown
+		result.Error = errors.New("important error(s) encountered during revocation check, status is Unknown")
```
priteshbandi, line 603: If we are returning error its implicit that it is important.

```suggestion
		result.Error = errors.New("signing certificate(s) revocation status is unknown")
```

verifier/verifier_test.go
```diff
@@ -310,6 +316,91 @@ func assertNotationVerification(t *testing.T, scheme signature.SigningScheme) {
...
+				CertificateChain: certChain,
+			},
+		},
+		VerificationLevel: &trustpolicy.VerificationLevel{
+			Enforcement: map[trustpolicy.ValidationType]trustpolicy.ValidationAction{trustpolicy.TypeRevocation: trustpolicy.ActionEnforce},
+		},
+	}
+}
+
+func TestVerifyRevocation(t *testing.T) {
```
priteshbandi, line 335: 1. Can you please add a test where certificate is revoked but trust policy is configured to log revocation. validate that verification succeeds but proper error message is shown to user 
2. Can you please add a test where certificate is revoked but trust policy is configured to skip revocation. Validate that no revocation calls are made.

verifier/verifier.go
kody-kimberl: Exactly. It returns notation.Verifier, but the private revocationClient is not defined for this struct, only the local verifier struct includes this (which we know is returned, but the function declares the public interface). Thus, we need to cast from notation.Verifier to verifier. This has been discussed previously in https://github.com/notaryproject/notation-go/pull/295#discussion_r1151292505

verifier/verifier.go
kody-kimberl: No, that would have it break out of the switch, not the outer for loop. This will be revised when combining the loops as suggested in the above comment

verifier/verifier.go
rgnote: Can we include the certificate subject that was revoked. Otherwise, users will need to parse the signatures to debug issues.

kody-kimberl: The subject can be fairly long for an error message. I think that this may cause difficulties for understanding what went wrong (especially if there are multiple certificates. I believe a better alternative would be to add the subject of the cert to the debug log for each error instead in addition to the number of the cert in the chain.

rgnote: I don't think the error messages with subject names would be of unreasonable length. Notation already includes certificate subject names in various error messages.

verifier/verifier.go
kody-kimberl: Resolved in other thread: https://github.com/notaryproject/notation-go/pull/295#discussion_r1163419338

verifier/verifier.go
```diff
@@ -518,6 +536,69 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+			default:
+				// revocation_base.ResultUnknown
+				// overwrite result of OK, but not Revoked
+				currResult = revocation_base.ResultUnknown
+				// keep subject closest to leaf
+				if storedCertSubject == "" {
+					storedCertSubject = outcome.EnvelopeContent.SignerInfo.CertificateChain[i].Subject.String()
+				}
+			}
+		}
```
rgnote, line 586: This logic seems bit brittle (the need for comments to explain what this code is doing is an indicator). 

Since both `ResultRevoked` and `ResultUnknown` result in overall failure of revocation, I suggest doing something like below

```
	isErrored := false
	revocationStatus := revocation_base.ResultUnknown
	erroredCert := ""
	for i, certResult := range revocationResult {
		for _, err := range certResult.ServerResults {
			if err != nil {
				logger.Debugf("error for certificate #%d in chain with subject %v: %v", (i + 1), outcome.EnvelopeContent.SignerInfo.CertificateChain[i].Subject.String(), err)

		if certResult.Result == revocation_base.ResultRevoked || certResult.Result ==  revocation_base.ResultUnknown {
				isErrored = true
				revocationStatus = certResult.Result
				erroredCert = outcome.EnvelopeContent.SignerInfo.CertificateChain[i].Subject.String()
				break
			}
		if isErrored {
				break
			}
		}
```

The difference between the two approaches is that one gives precedence to `Revoked` error over `Unkown` error another doesn't. Given that both errors ultimately result in signature verification failing, I don't think we need to worry about giving precedence at all? What are you thoughts?

verifier/verifier.go
```diff
@@ -256,9 +269,14 @@ func (v *verifier) processSignature(ctx context.Context, sigBlob []byte, envelop
 	// skipped using a trust policy or a plugin may override the check
 	if outcome.VerificationLevel.Enforcement[trustpolicy.TypeRevocation] != trustpolicy.ActionSkip &&
 		!slices.Contains(pluginCapabilities, proto.CapabilityRevocationCheckVerifier) {
-		logger.Debugf("Validating revocation (not implemented)")
-		// TODO perform X509 revocation check (not in RC1)
-		// https://github.com/notaryproject/notation-go/issues/110
+
+		logger.Debug("Validating revocation")
```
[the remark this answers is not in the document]
kody-kimberl: This is matching the existing Debug logs for other validations (see lines 183, 227, 238, 250, 259, 289, and 296)

verifier/verifier.go
```diff
@@ -518,6 +536,69 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+
+	signingTime := outcome.EnvelopeContent.SignerInfo.SignedAttributes.SigningTime
+	revocationResult := r.Validate(outcome.EnvelopeContent.SignerInfo.CertificateChain, signingTime)
+
+	result := &notation.ValidationResult{
+		Type:   trustpolicy.TypeRevocation,
+		Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeRevocation],
+	}
+
+	currResult := revocation_base.ResultOK
```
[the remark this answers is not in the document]
kody-kimberl: Why would we default to Unknown? Should we not start with OK and convert to it to Unknown when a certificate has that result? It seems weird to specify Unknown as the default, then check to see if every cert is OK or NonRevokable to eventually turn it back. There should not be any cases where this default is returned without being overridden by an Unknown or Revoked result that is present

verifier/verifier.go
kody-kimberl: The reason for comments is a result of combining two separate processes into one loop in order to save some cycles per Pritesh's recommendation.

I still believe that there should be precedence for Revoked over Unknown. If even one certificate is revoked, then the chain should be considered revoked. I think this is important for users to be aware of. Let's take an example where cert 3 is Unknown and cert 5 is Revoked. If we just return at the first failure (in this case an Unknown), they may think that it is good to try again later (or may go into debug logs to find the error that occurred). However, if we prioritize revoked, then they will see easily that there is a certificate in the chain that is revoked (which may change how a user interprets or acts upon the response).

verifier/verifier.go
```diff
@@ -518,6 +536,69 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+	storedCertSubject := ""
+	for i, certResult := range revocationResult {
+		// Log all non-nil errors as debug
+		for _, err := range certResult.ServerResults {
+			if err != nil {
+				logger.Debugf("error for certificate #%d in chain with subject %v: %v", (i + 1), outcome.EnvelopeContent.SignerInfo.CertificateChain[i].Subject.String(), err)
```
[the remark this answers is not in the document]
kody-kimberl: Also, if we add the server to this, the message starts to get fairly long. Would it be better to split this into a few separate messages? Maybe something like:
```
logger.Debugf("error for certificate #%d in chain: %v", (i + 1), err)
logger.Debugf("\tcertificate subject: %v", outcome.EnvelopeContent.SignerInfo.CertificateChain[i].Subject.String())
logger.Debugf("\tserver URI: %v", <some new way to obtain the server URI>)
```

verifier/verifier.go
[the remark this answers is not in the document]
priteshbandi: >There should not be any cases where this default is returned without being overridden by an Unknown or Revoked result that is present

This can happen during code bug and thats the reason its recommended to start with non-happy case to avoid such issues.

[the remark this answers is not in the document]
kody-kimberl: But what I'm saying is how can this ever not be representative of the results? It goes through every result and checks the status, so if any are not OK it will reassign the currResult. I understand the hesitation, but I also think that starting with ResultUnknown makes it much more difficult to confirm that all of the results are OK if there is no error.

For example, the alternative to this would be going through the same logic but starting with a ResultUnknown. Then, we would need to count the number of OK results and compare to the total number of results in order to guarantee that they are all OK. I think this approach is much less understandable, but I can implement it if you think it is a better approach. I just don't see a scenario where this could be an issue. Can you give me a potential case where this Result would be incorrect because of the default?

verifier/verifier.go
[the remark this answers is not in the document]
kody-kimberl: Consensus was to add server URI string to ServerResult, and keep debug as one line since it is just a debug log.

verifier/verifier.go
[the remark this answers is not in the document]
kody-kimberl: Updated to make logic more clear and concise and default to ResultUnknown, since false negative is preferred over false positive in the case of a bug/unexpected error.

verifier/verifier.go
kody-kimberl: Consensus was that the Revoked over Unknown precedence makes sense. Updated to make logic more clear and concise.

verifier/verifier.go
```diff
@@ -518,6 +536,72 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
 	}
 }
 
+func verifyRevocation(outcome *notation.VerificationOutcome, client *http.Client, logger log.Logger) *notation.ValidationResult {
+	r, err := revocation.New(client)
```
shizhMSFT, line 540: Do we need to create a revocation instance each time? Is the instance reusable?

verifier/verifier.go
```diff
@@ -50,16 +54,25 @@ func NewFromConfig() (notation.Verifier, error) {
 
 // New creates a new verifier given trustPolicy, trustStore and pluginManager
 func New(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore, pluginManager plugin.Manager) (notation.Verifier, error) {
+	return NewWithRevocationClient(trustPolicy, trustStore, pluginManager, &http.Client{Timeout: 5 * time.Second})
+}
+
+// NewWithRevocationClient creates a new verifier given trustPolicy, trustStore, pluginManager, and revocationClient
+func NewWithRevocationClient(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore, pluginManager plugin.Manager, revocationClient *http.Client) (notation.Verifier, error) {
```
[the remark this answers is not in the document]
shizhMSFT: Instead of passing `revocationClient *http.Client`, how about `revocationClient revocation.Revocation`. Doing so also helps the unit tests.

verifier/verifier.go
```diff
@@ -518,6 +536,72 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+		}
+	}
+
+	result := &notation.ValidationResult{
+		Type:   trustpolicy.TypeRevocation,
+		Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeRevocation],
+	}
+
+	currResult := revocation_result.ResultUnknown
+	numOKResults := 0
```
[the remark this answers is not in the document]
kody-kimberl: The purpose of the counter is to ensure that if for some reason the loop ends before visiting a cert with a result of Unknown or Revoked that it will only be considered OK if the number of OK/NonRevokable results was the same as the number of results expected to be parsed. Essentially, it is just further bug protection to avoid false positives

verifier/verifier.go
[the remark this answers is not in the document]
kody-kimberl: I'm not sure that is the best approach. Since the creation of a revocation.Revocation object via New can result in an error, that means that an additional error check needs to be done everywhere that this would be switched. This would make the unit tests more complex, but more importantly, there would potentially be revocation related errors when calling the verify constructors (as it would need to call revocation.New with a client). I think that it makes sense to keep these errors confined to the revocation check (considering a use case like someone wanting to skip revocation, but getting a revocation related error anyways in a different spot)

verifier/verifier.go
kody-kimberl: The main rationale for this is if the client changes. Since this can accept a different client, a different revocation object would be needed. Changing the client for a revocation.Revocation object would require access to the revocation struct (which cannot be used outside that package), so we cannot reuse one.

verifier/verifier.go
```diff
@@ -518,6 +536,72 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+		logger.Debug("error while creating a revocation client, err: %s", err.Error())
+		return &notation.ValidationResult{
+			Type:   trustpolicy.TypeRevocation,
+			Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeRevocation],
+			Error:  fmt.Errorf("unable to check revocation status, err: %s", err.Error()),
+		}
+	}
+
+	signingTime := outcome.EnvelopeContent.SignerInfo.SignedAttributes.SigningTime
+	revocationResult, err := r.Validate(outcome.EnvelopeContent.SignerInfo.CertificateChain, signingTime)
```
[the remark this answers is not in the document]
kody-kimberl: I think the only time this can happen is if for some reason the SingingTime isn't specified. I think this check makes sense, but I'm not 100% certain on the desired alternative behavior. For now, I'll update signingTime to time.Now() if IsZero() is true. That way the cert that has been revoked would be considered revoked regardless of the time it is RevokedAt (unless RevokedAt has been set to a time in the future, in which case this would not be considered revoked since it is before).

verifier/verifier.go
```diff
@@ -518,6 +536,77 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+	finalResult := revocation_result.ResultUnknown
+	numOKResults := 0
+	var leafmostProblematicCertSubject string
+	for i, certResult := range revocationResult {
+		for _, serverResult := range certResult.ServerResults {
+			if serverResult != nil {
+				logger.Debugf("error for certificate #%d in chain with subject %v for server %q: %v", (i + 1), outcome.EnvelopeContent.SignerInfo.CertificateChain[i].Subject.String(), serverResult.Server, serverResult.Error)
+			}
+		}
+
```
priteshbandi, line 578: Can we remove this forloop because even if debug is not set we will iterate over all serverResult.

verifier/verifier.go
kody-kimberl: This is obsolete in new implementation as the instance is passed, resolving

verifier/verifier.go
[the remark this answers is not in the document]
kody-kimberl: After discussing with Pritesh, I think this error here is okay as they can pass their own revocation.Revocation implementation to avoid any issues. I have added a test to ensure that the default implementation used in the New constructor does not result in errors in the future. Resolving

verifier/verifier.go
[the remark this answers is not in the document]
patrickzheng200: > I think the only time this can happen is if for some reason the SingingTime isn't specified. I think this check makes sense, but I'm not 100% certain on the desired alternative behavior. For now, I'll update signingTime to time.Now() if IsZero() is true. That way the cert that has been revoked would be considered revoked regardless of the time it is RevokedAt (unless RevokedAt has been set to a time in the future, in which case this would not be considered revoked since it is before).

How about checking `IsZero()` in the notation-core-go code? something like:
```
...
if signingTime.IsZero() {xxx}
// Check if SigningTime was before the revocation
if opts.SigningTime.Before(resp.RevokedAt) {
	return errToServerResult(server, nil)
}
...
```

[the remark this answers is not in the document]
kody-kimberl: I think that may obscure the behavior in notation-core-go. IMO we should keep notation-core-go the same (to avoid assuming how this zero time should be handled) as users can simply pass a different time if they want it handled differently (similar to the reassignment of signingTime that is being done in the current implementation).

verifier/verifier.go
```diff
@@ -518,6 +540,78 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+			if revocationResult[i].Result == revocation_result.ResultRevoked {
+				revokedFound = true
+				revokedCertSubject = problematicCertSubject
+			}
+		}
+	}
+	if revokedFound {
+		problematicCertSubject = revokedCertSubject
+		finalResult = revocation_result.ResultRevoked
+	}
```
patrickzheng200, line 597: I'm thinking if we could make this part of code more elegant. Please correct me if I'm wrong, if the program could reach to line 589 `revokedFound = true`, looks like the program will always reach to line 595 `problematicCertSubject = revokedCertSubject`, because once `revokedFound` is set to true, it will stay to be true for the rest of the program. In this case, can we just set up `result.Error` and return `result` at the place where we set `revokedFound`, i.e., line 589?

verifier/verifier.go
```diff
@@ -50,16 +54,29 @@ func NewFromConfig() (notation.Verifier, error) {
...
 func New(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore, pluginManager plugin.Manager) (notation.Verifier, error) {
+	r, err := revocation.New(&http.Client{Timeout: 5 * time.Second})
+	if err != nil {
+		return nil, err
+	}
+	return NewWithOptions(trustPolicy, trustStore, pluginManager, r)
+}
+
+// NewWithOptions creates a new verifier given trustPolicy, trustStore, pluginManager, and revocationClient
+func NewWithOptions(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore, pluginManager plugin.Manager, revocationClient revocation.Revocation) (notation.Verifier, error) {
```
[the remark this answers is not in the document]
kody-kimberl: This was my thought as well when changing the name. I think that would make sense, but want to hear perspective of others before implementing

verifier/verifier.go
kody-kimberl: The issue with this is related to the subject. We want to use the leafmost revoked cert's subject. Thus, we are iterating backwards and do not break out (in case there is one closer to the leaf). This entire check for 594 is related to this subject for the error. This was the result of discussions about simplifying the logic with @rgnote and @priteshbandi

verifier/verifier.go
```diff
@@ -518,6 +540,78 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+			Type:   trustpolicy.TypeRevocation,
+			Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeRevocation],
+			Error:  fmt.Errorf("unable to check revocation status, revocation client cannot be nil"),
+		}
+	}
+
+	signingTime := outcome.EnvelopeContent.SignerInfo.SignedAttributes.SigningTime
+	if signingTime.IsZero() {
+		signingTime = time.Now()
+	}
```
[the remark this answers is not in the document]
kody-kimberl: Fair point. I'll move this check to core-go and add some documentation about it.

verifier/verifier.go
patrickzheng200: Oh, I see. My initial understanding is "whenever we find a revoked cert in a cert chain, then the cert chain is immediately considered revoked, and we don't need to check the remaining certs in the chain anymore."
If we always want the leaf-most revoked cert's subject to be returned, then yes, the current logic is necessary.

kody-kimberl: This was actually how I had it before. The issue with this is that the logic for differentiating between Unknown and Revoked becomes more complicated (more specifically, how to ensure that the leafmost cert subject for Unknown is used without being able to break). There is a trade-off for either approach, so the current one was decided as the most understandable to achieve both a precedence for Revoked over Unknown and using the subject for the leafmost cert for the appropriate result.

verifier/verifier.go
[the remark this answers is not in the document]
kody-kimberl: Added VerifierOptions in newest commit

priteshbandi (commented):
LGTM

verifier/verifier.go
```diff
@@ -50,16 +62,30 @@ func NewFromConfig() (notation.Verifier, error) {
...
+// NewWithOptions creates a new verifier given trustPolicy, trustStore, pluginManager, and revocationClient
+func NewWithOptions(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore, pluginManager plugin.Manager, opts VerifierOptions) (notation.Verifier, error) {
+	revocationClient := opts.RevocationClient
+	if revocationClient == nil {
+		var err error
+		revocationClient, err = revocation.New(&http.Client{Timeout: 5 * time.Second})
+		if err != nil {
+			return nil, err
+		}
+	}
```
priteshbandi, line 77: Do not defualt in NewWithOptions instead use New to defualt.
```
func New(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore, pluginManager plugin.Manager) (notation.Verifier, error) {
        revocationClient, err = revocation.New(&http.Client{Timeout: 5 * time.Second})
        if err != nil {
	        return nil, err
        }
	return NewWithOptions(trustPolicy, trustStore, pluginManager, VerifierOptions{RevocationClient: revocationClient})
}
```

verifier/verifier.go
```diff
@@ -518,6 +549,76 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+			Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeRevocation],
+			Error:  fmt.Errorf("unable to check revocation status, revocation client cannot be nil"),
+		}
+	}
+
+	authenticSigningTime, err := outcome.EnvelopeContent.SignerInfo.GetAuthenticSigningTime()
+	if err != nil {
+		logger.Debugf("not using authentic signing time due to error retrieving AuthenticSigningTime, err: %v", err)
+	}
+	revocationResult, err := r.Validate(outcome.EnvelopeContent.SignerInfo.CertificateChain, authenticSigningTime)
```
patrickzheng200, line 565: nit:
can we rename `revocationResult` to `revocationResults` to reflect it's an array?
```suggestion
	revocationResults, err := r.Validate(outcome.EnvelopeContent.SignerInfo.CertificateChain, authenticSigningTime)
```

verifier/verifier.go
```diff
@@ -518,6 +549,76 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+		problematicCertSubject = revokedCertSubject
+		finalResult = revocationresult.ResultRevoked
+	}
+	if numOKResults == len(revocationResult) {
+		finalResult = revocationresult.ResultOK
+	}
+
+	switch finalResult {
+	case revocationresult.ResultOK:
+		logger.Debug("no important errors encountered while checking revocation, status is OK")
```
patrickzheng200, line 611: I'm not sure what does "no `important` errors" mean here? Should it just be `no error`?
```suggestion
		logger.Debug("no error encountered while checking revocation, status is OK")
```

verifier/verifier.go
```diff
@@ -50,16 +62,30 @@ func NewFromConfig() (notation.Verifier, error) {
 
 // New creates a new verifier given trustPolicy, trustStore and pluginManager
 func New(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore, pluginManager plugin.Manager) (notation.Verifier, error) {
+	return NewWithOptions(trustPolicy, trustStore, pluginManager, VerifierOptions{})
+}
+
+// NewWithOptions creates a new verifier given trustPolicy, trustStore, pluginManager, and revocationClient
```
patrickzheng200, line 68: nit:
```suggestion
// NewWithOptions creates a new verifier given trustPolicy, trustStore, pluginManager, and VerifierOptions
```

verifier/verifier.go
kody-kimberl: Updated to "no verification impacting errors" rather than important in order to provide more clarity.

rgnote (dismissed):
LGTM

priteshbandi (approved):
LGTM

verifier/verifier.go
```diff
@@ -518,6 +550,76 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+			Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeRevocation],
+			Error:  fmt.Errorf("unable to check revocation status, err: %s", err.Error()),
+		}
+	}
+
+	result := &notation.ValidationResult{
+		Type:   trustpolicy.TypeRevocation,
+		Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeRevocation],
+	}
+
```
patrickzheng200, line 580: nit:
can remove extra line
```suggestion
```

patrickzheng200 (commented):
LGTM with one nit comment.

verifier/verifier.go
```diff
@@ -518,6 +550,76 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+	finalResult := revocationresult.ResultUnknown
+	numOKResults := 0
+	var problematicCertSubject string
+	revokedFound := false
+	var revokedCertSubject string
+	for i := len(certResults) - 1; i >= 0; i-- {
+		if certResults[i].ServerResults[0].Error != nil {
```
shizhMSFT, line 587: What if `len(ServerResults) == 0`?

verifier/verifier.go
```diff
@@ -518,6 +550,76 @@ func verifyAuthenticTimestamp(outcome *notation.VerificationOutcome) *notation.V
...
+		}
+
+		if certResults[i].Result == revocationresult.ResultOK || certResults[i].Result == revocationresult.ResultNonRevokable {
+			numOKResults++
+		} else {
+			finalResult = certResults[i].Result
+			problematicCertSubject = outcome.EnvelopeContent.SignerInfo.CertificateChain[i].Subject.String()
+			if certResults[i].Result == revocationresult.ResultRevoked {
+				revokedFound = true
+				revokedCertSubject = problematicCertSubject
```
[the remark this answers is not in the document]
kody-kimberl: This has been discussed a few times (most recently here: https://github.com/notaryproject/notation-go/pull/295#discussion_r1169417279)

TLDR: There are 2 problems with this:
 - Returning early means we wouldn't log every cert
 - If there is a revoked cert closer to the leaf, it wouldn't reflect that cert's subject

The current implementation goes from root to leaf without breaking to ensure everything gets logged, then stores both the revoked and unknown problematic subjects each time it is applicable (since this will ensure leafmost) and choses the correct one to use afterwards. This helps to avoid more complex/less understandable logic that would be needed to do this with iterating from leaf to root or returning early.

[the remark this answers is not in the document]
shizhMSFT: > Returning early means we wouldn't log every cert

This makes senses.

> If there is a revoked cert closer to the leaf, it wouldn't reflect that cert's subject

We can iterate the loop from the other end ⚡

shizhMSFT (approved):
LGTM

priteshbandi (approved):

PR #298, opened 2023-04-07
workflows: Add Gosec workflow

AdamKorcz:
Adds a workflow for Gosec, and removes false positive findings.

### DIFF
--- a/.github/workflows/gosec.yml
+++ b/.github/workflows/gosec.yml
@@ -1,0 +1,20 @@
+name: Run Gosec
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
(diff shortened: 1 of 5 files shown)


.github/workflows/gosec.yml
```diff
@@ -0,0 +1,20 @@
...
+    branches:
+      - main
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v2
```
shizhMSFT, line 16: nit:
```suggestion
        uses: actions/checkout@v3
```

PR #306, opened 2023-05-05
build: bump image-spec to v1.1.0-rc.3 and oras-go to v2.1.0

shizhMSFT:
Closes #305
Closes #307

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,11 +6,11 @@ require (
 	github.com/go-ldap/ldap/v3 v3.4.4
 	github.com/notaryproject/notation-core-go v1.0.0-rc.3
 	github.com/opencontainers/go-digest v1.0.0
-	github.com/opencontainers/image-spec v1.1.0-rc2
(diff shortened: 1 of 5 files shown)


sajayantony:
Should we remove the ability to create artifactManifest mediaType and only allow for read with a deprecation warning?

rchincha:
Also, under "internal/" so external imports don't work yet.

FeynmanZhou:
> Should we remove the ability to create artifactManifest mediaType and only allow for read with a deprecation warning?

Agree with  @sajayantony. This change is also supposed to be considered in Notation CLI.

shizhMSFT:
> Should we remove the ability to create artifactManifest mediaType and only allow for read with a deprecation warning?

This is partially addressed by #308. We might need more discussions on the warning and it can be done in another PR.

JeyJeyGao (approved):
LGTM

PR #316, opened 2023-05-24
feat: add validations for symlink

priteshbandi:
Added validations to make sure singingkeys.json, trustpolicy.json and config.json doesnt read's symlink

closes https://github.com/notaryproject/notation/issues/634

### DIFF
--- a/config/base.go
+++ b/config/base.go
@@ -2,6 +2,8 @@ package config
 
 import (
 	"encoding/json"
+	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 
@@ -28,7 +30,22 @@ func save(filePath string, cfg interface{}) error {
 
 // load reads file, parses json and stores in cfg struct
 func load(filePath string, cfg interface{}) error {
-	file, err := dir.ConfigFS().Open(filePath)
+	path, err := dir.ConfigFS().SysPath(filePath)
+	if err != nil {
+		return err
+	}
+
+	// throw error if path is a directory or is a symlink or does not exist.
+	fileInfo, err := os.Lstat(path)
+	if err != nil {
+		return err
+	}
+	mode := fileInfo.Mode()
+	if mode.IsDir() || mode&fs.ModeSymlink != 0 {
+		return fmt.Errorf("%q is not a regular file (symlinks are not supported)", path)
+	}
+
+	file, err := os.Open(path)
 	if err != nil {
 		return err
 	}
--- a/verifier/trustpolicy/trustpolicy.go
+++ b/verifier/trustpolicy/trustpolicy.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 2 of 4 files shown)


config/base.go
```diff
@@ -26,8 +28,23 @@ func save(filePath string, cfg interface{}) error {
 	return encoder.Encode(cfg)
 }
 
-// load reads file, parses json and stores in cfg struct
-func load(filePath string, cfg interface{}) error {
+// Load reads file, parses json and stores in cfg struct
+func Load(filePath string, cfg interface{}) error {
```
shizhMSFT, line 32: It is too public to be exposed. Consider putting this function into an `internal` package.

verifier/trustpolicy/trustpolicy.go
```diff
@@ -267,23 +267,19 @@ func (trustPolicyDoc *Document) GetApplicableTrustPolicy(artifactReference strin
 
 // LoadDocument loads a trust policy document from a local file system
 func LoadDocument() (*Document, error) {
-	jsonFile, err := dir.ConfigFS().Open(dir.PathTrustPolicy)
+	policyDocument := &Document{}
+
+	err := config.Load(dir.PathTrustPolicy, policyDocument)
 	if err != nil {
```
JeyJeyGao, line 273: nit:
```suggestion
	if err := config.Load(dir.PathTrustPolicy, policyDocument); err != nil {
```

verifier/trustpolicy/trustpolicy.go
```diff
@@ -267,23 +267,19 @@ func (trustPolicyDoc *Document) GetApplicableTrustPolicy(artifactReference strin
...
 		switch {
 		case errors.Is(err, os.ErrNotExist):
 			return nil, fmt.Errorf("trust policy is not present. To create a trust policy, see: %s", trustPolicyLink)
 		case errors.Is(err, os.ErrPermission):
 			return nil, fmt.Errorf("unable to read trust policy due to file permissions, please verify the permissions of %s", filepath.Join(dir.UserConfigDir, dir.PathTrustPolicy))
 		default:
-			return nil, err
+			return nil, fmt.Errorf("malformed trust policy. To create a trust policy, see: %s", trustPolicyLink)
```
shizhMSFT, line 279: This error message is not accurate. If the path is a symlink or other errors, the detail of the error is conveyed by "malformed trust policy".

config/base.go
```diff
@@ -26,8 +28,23 @@ func save(filePath string, cfg interface{}) error {
...
+	fileInfo, err := os.Lstat(path)
+	if err != nil {
+		return err
+	}
+	mode := fileInfo.Mode()
+	if mode.IsDir() || mode&fs.ModeSymlink != 0 {
+		return fmt.Errorf("%q is not a regular file (symlinks are not supported)", path)
+	}
+
 	file, err := dir.ConfigFS().Open(filePath)
```
shizhMSFT, line 48: Since you have `path`, you may directly call `os.Open(path)`.

config/base.go
Two-Hearts, line 32: Since `load` is a helper function, I don't think we should expose it to external users directly. Maybe we could move it into an internal package.

priteshbandi: There is no good way to do this without refactoring how confg.json and signingkeys.json are read/modified and thats the reason I am going to leave this method private and duplicate code in trustpolicy.go file

config/base.go
```diff
@@ -28,7 +30,22 @@ func save(filePath string, cfg interface{}) error {
 
 // load reads file, parses json and stores in cfg struct
 func load(filePath string, cfg interface{}) error {
-	file, err := dir.ConfigFS().Open(filePath)
+	path, err := dir.ConfigFS().SysPath(filePath)
+	if err != nil {
+		return err
+	}
+
+	// throw error if path is not a directory or is a symlink or does not exist.
```
Two-Hearts, line 38: nit:
```suggestion
	// throw error if path is a directory or is a symlink or does not exist.
```

verifier/trustpolicy/trustpolicy.go
```diff
@@ -267,18 +268,34 @@ func (trustPolicyDoc *Document) GetApplicableTrustPolicy(artifactReference strin
...
 func LoadDocument() (*Document, error) {
-	jsonFile, err := dir.ConfigFS().Open(dir.PathTrustPolicy)
+	path, err := dir.ConfigFS().SysPath(dir.PathTrustPolicy)
 	if err != nil {
-		switch {
-		case errors.Is(err, os.ErrNotExist):
+		return nil, err
+	}
+
+	// throw error if path is not a directory or is a symlink or does not exist.
```
Two-Hearts, line 276: nit:
```suggestion
	// throw error if path is a directory or is a symlink or does not exist.
```

Two-Hearts:
@priteshbandi it seems that your commits are not signed? Otherwise, LGTM.

priteshbandi:
> @priteshbandi it seems that your commits are not signed? Otherwise, LGTM.

Fixed the issue, the email address in git config was incorrect.

Two-Hearts (approved):
LGTM

rgnote (approved):
LGTM

shizhMSFT (approved):
LGTM but we definitely can refactor the open file sub-routine in a future PR.

JeyJeyGao (approved):
LGTM

PR #317, opened 2023-05-24
fix: add digest check for Sign

byronchien:
This PR adds digest check on Sign.

Signed-off-by: Byron Chien <byronc@ucla.edu>
Co-authored-by: Patrick Zheng <patrickzheng@microsoft.com>

### DIFF
--- a/notation.go
+++ b/notation.go
@@ -95,22 +95,21 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, signOpts
 		// artifactRef is a valid full reference
 		artifactRef = ref.Reference
 	}
-
 	targetDesc, err := repo.Resolve(ctx, artifactRef)
 	if err != nil {
 		return ocispec.Descriptor{}, fmt.Errorf("failed to resolve reference: %w", err)
 	}
 	// artifactRef is a tag or a digest, if it's a digest it has to match
 	// the resolved digest
-	if dgst, err := digest.Parse(artifactRef); err != nil {
-		// artifactRef is not a digest reference
+	if artifactRef != targetDesc.Digest.String() {
+		if _, err := digest.Parse(artifactRef); err == nil {
+			// artifactRef is a digest, but does not match the resolved digest
+			return ocispec.Descriptor{}, fmt.Errorf("user input digest %s does not match the resolved digest %s", artifactRef, targetDesc.Digest.String())
+		}
+		// artifactRef is a tag
 		logger.Warnf("Always sign the artifact using digest(`@sha256:...`) rather than a tag(`:%s`) because tags are mutable and a tag reference can point to a different artifact than the one signed", artifactRef)
 		logger.Infof("Resolved artifact tag `%s` to digest `%s` before signing", artifactRef, targetDesc.Digest.String())
-	} else if dgst.String() != targetDesc.Digest.String() {
-		// artifactRef is a digest, but does not match the resolved digest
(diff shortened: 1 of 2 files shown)


priteshbandi (approved):
LGTM

Two-Hearts:
After running through a test workflow from Notation CLI, I found this change is not quite correct.
`./notation.exe sign --allow-referrers-api <reg>/<repo>@sha256:b6ca290b6b4cdcca5b3db3ffa338ee0285c11744b4a6abaa9627746ee3291d8d -v`
The `notation-go` logic still sees the above digest as a `tag` and prints out the warning `Always sign the artifact using digest...`.

The reason causing this issue: `notation CLI` will always set `signOpts.ArtifactReference` as a `digest` and pass it to `notation-go`, related code: https://github.com/notaryproject/notation/blob/a695b60ed2b0d9a173724c23b5f2c0dd04558f3e/cmd/notation/sign.go#L133.
Thus, in `notation-go`, given the input is from `Notation CLI`, `ref, err := orasRegistry.ParseReference(artifactRef)` would always return `non-nil` error with EMPTY `ref` (this code is valid because `non-CLI` users might choose to pass a full reference to `notation-go`). 
Then, the new change in this PR `if ref.ValidateReferenceAsDigest() != nil` will always be evaluated to `true`, as long as the caller is `notation CLI`.
/cc: @priteshbandi @JeyJeyGao @shizhMSFT

notation.go
```diff
@@ -90,18 +90,22 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, signOpts
...
 	if err != nil {
 		return ocispec.Descriptor{}, fmt.Errorf("failed to resolve reference: %w", err)
 	}
-	if artifactRef != targetDesc.Digest.String() {
+	if ref.ValidateReferenceAsDigest() != nil {
 		// artifactRef is not a digest reference
 		logger.Warnf("Always sign the artifact using digest(`@sha256:...`) rather than a tag(`:%s`) because tags are mutable and a tag reference can point to a different artifact than the one signed", artifactRef)
 		logger.Infof("Resolved artifact tag `%s` to digest `%s` before signing", artifactRef, targetDesc.Digest.String())
+	} else if ref.Reference != targetDesc.Digest.String() {
+		return ocispec.Descriptor{}, fmt.Errorf("user input digest %s does not match the resolved digest %s", ref.Reference, targetDesc.Digest.String())
```
Two-Hearts, line 108: Proposing an alternative solution:
```
import "github.com/opencontainers/go-digest"
```
and under `Sign`, have something like this:
```
// at this point, ref can be empty and should not be used
// artifactRef is a tag or a digest, if it's a digest it has to match
// the resolved digest.
dgst, err := digest.Parse(artifactRef)
if err != nil {
    // artifactRef is not a digest reference
    logger.Warnf("Always sign the artifact using digest(`@sha256:...`) rather than a tag(`:%s`) because tags are mutable and a tag reference can point to a different artifact than the one signed", artifactRef)
    logger.Infof("Resolved artifact tag `%s` to digest `%s` before signing", artifactRef, targetDesc.Digest.String())
} else if dgst.String() != targetDesc.Digest.String() {
    return ocispec.Descriptor{}, fmt.Errorf("user input digest %s does not match the resolved digest %s", ref.Reference, targetDesc.Digest.String())
}
```

priteshbandi (approved):
LGTM

JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #335, opened 2023-07-12
Refactoring to thin out notation pkg

priteshbandi:
Refactoring to thin out notation pkg.

In this change, we are moving bunch of code from `notation` cli to `notation-go` pkg so that `notation` pkg is truely a cli wrapper around `notation-go`. This change is related to https://github.com/notaryproject/notation/pull/740


Two-Hearts (commented):
@priteshbandi Could you provide more details on the purpose of this PR? Thanks.

signingkeys/signingkeys.go
shizhMSFT, line 1: What's the reason to create a new package outside of the config package? The implementation of this package is strongly related to configs. If we want to have a new package, it should still be under the `config` folder like `config/signingkeys/signingkeys.go`.

priteshbandi:
Slate PR,

PR #345, opened 2023-08-29
fix: update error message from notation-go

Two-Hearts:
This PR tends to solve the following issues from Notation CLI which requires a few changes in notation-go library:
[#699](https://github.com/notaryproject/notation/issues/699), [#700](https://github.com/notaryproject/notation/issues/700), [#701](https://github.com/notaryproject/notation/issues/701).

This PR includes:
<del>Updated `verificationOutcomes` in the return of `notation.Verify` to include verification failed reasons of each signature, so that Notation CLI could display them to the user without having to enable the `-v` or `-d` flag.

1. Updated `err` returned from `notation.Verify` as a joined error. (based on code review, `verificationOutcomes` related logic is not changed in this PR.)
3. The error message updates related to trust store, based on discussions from the issues mentioned above.

The updated error messages is displayed in the PR of Notation CLI: https://github.com/notaryproject/notation/pull/771.


verifier/truststore/errors.go
```diff
@@ -0,0 +1,51 @@
...
+type ErrorNonExistence struct {
+	WrappedError error
+}
+
+func (e ErrorNonExistence) Error() string {
+	if e.WrappedError != nil {
+		return e.WrappedError.Error()
+	}
+	return "unable to find specified trust store or certificate"
+}
```
shizhMSFT, line 51: This error type should really be replaced with `ErrorTrustStore` and `ErrorNonExistence`.

shizhMSFT: Meanwhile, consider wrapping [fs.ErrNotExist](https://pkg.go.dev/io/fs#pkg-variables).

notation.go
```diff
@@ -193,10 +193,13 @@ type ValidationResult struct {
...
 
-// VerificationOutcome encapsulates a signature blob's descriptor, its content,
-// the verification level and results for each verification type that was
-// performed.
+// VerificationOutcome encapsulates a signature manifest's descriptor,
+// its envelope blob, its content, the verification level and results for each
+// verification type that was performed.
 type VerificationOutcome struct {
+	// SignatureManifestDescriptor
+	SignatureManifestDescriptor *ocispec.Descriptor
```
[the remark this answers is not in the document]
shizhMSFT: Instead, consider embedding the digest of the signature manifest descriptor in `VerificationOutcome.Error`.

verifier/truststore/errors.go
```diff
@@ -0,0 +1,51 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package truststore
+
+// ErrorTrustStore is used when accessing specified trust store failed
+type ErrorTrustStore struct {
```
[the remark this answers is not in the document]
Two-Hearts: Added `Unwrap`.

verifier/truststore/errors.go
```diff
@@ -0,0 +1,51 @@
...
+	WrappedError error
+}
```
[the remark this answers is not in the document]
Two-Hearts: Oh, I see. In `errors.go` under `notation` package, we already have all the types defined as `ErrorXXX` though. Should we keep the style so that they are consistent?

verifier/truststore/truststore.go
```diff
@@ -64,49 +64,49 @@ type x509TrustStore struct {
...
 	if err != nil {
-		return nil, err
+		return nil, ErrorTrustStore{InnerError: fmt.Errorf("failed to get path of trust store %s with type %s: %w", namedStore, storeType, err)}
 	}
 	// throw error if path is not a directory or is a symlink or does not exist.
 	fileInfo, err := os.Lstat(path)
 	if err != nil {
 		if os.IsNotExist(err) {
-			return nil, fmt.Errorf("%q does not exist", path)
+			return nil, ErrorTrustStore{InnerError: fs.ErrNotExist, Msg: fmt.Sprintf("the trust store %q of type %q doesn't exist", namedStore, storeType)}
```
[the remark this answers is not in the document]
Two-Hearts: @JeyJeyGao Nice suggestion, but here's why I need the original method. In your suggestion, when upstream prints this error message, the content of `fs.ErrNotExist` will be printed out. In the original method, only the user customized error message will be printed out (if the error can indeed be unwrapped to a `fs.ErrNotExist`).

JeyJeyGao (approved):
LGTM

verifier/truststore/truststore.go
shizhMSFT, line 80: nit: `err` contains `fs.ErrNotExist`
```suggestion
			return nil, ErrorTrustStore{InnerError: err, Msg: fmt.Sprintf("the trust store %q of type %q doesn't exist", namedStore, storeType)}
```

notation.go
```diff
@@ -347,12 +347,12 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, ve
 		return ocispec.Descriptor{}, nil, ErrorSignatureRetrievalFailed{Msg: fmt.Sprintf("user input digest %s does not match the resolved digest %s", ref.Reference, artifactDescriptor.Digest.String())}
 	}
 
+	var verificationSucceeded bool
 	var verificationOutcomes []*VerificationOutcome
+	var verificationFailedErr error = ErrorVerificationFailed{}
```
shizhMSFT, line 352: It seems `verificationFailedErr` is initialized and returned but never assigned.

notation.go
```diff
@@ -380,16 +380,15 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, ve
 					logger.Error("Got nil outcome. Expecting non-nil outcome on verification failure")
 					return err
 				}
-
-				if _, ok := outcome.Error.(ErrorUserMetadataVerificationFailed); ok {
-					verificationFailedErr = outcome.Error
-				}
```
shizhMSFT, line 386: This part is removed and thus this change may consider as a breaking change. Is this intended?

notation.go
```diff
@@ -380,16 +380,15 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, ve
...
-
+				outcome.Error = fmt.Errorf("failed to verify signature with digest %v, %w", sigManifestDesc.Digest, outcome.Error)
+				verificationOutcomes = append(verificationOutcomes, outcome)
```
shizhMSFT, line 384: The `Verify()` function states

> return the successful signature verification outcome.

However, this PR introduces a breaking change on the function behavior. It should be avoided.

shizhMSFT: In order not to break, you may consider [`errors.Join()`](https://pkg.go.dev/errors#Join) to combine multiple errors into a single one.

notation.go
Two-Hearts: Yes, this is also the behavior in v1.0.0. This is the general error on verification failure. When printing its error message, the output is `signature verification failed`.

notation.go
Two-Hearts: Nice catch! This is not intended to be deleted.

Two-Hearts: After the update, we could remove this part now, and it's not a breaking change.

notation.go
```diff
@@ -380,22 +381,24 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, ve
...
+			// succeeded outcome
+			verificationOutcomes = []*VerificationOutcome{outcome}
 			logger.Debugf("Signature verification succeeded for artifact %v with signature digest %v", artifactDescriptor.Digest, sigManifestDesc.Digest)
 
 			// early break on success
 			return errDoneVerification
 		}
 
+		// at this point, the verification process is failed
+		verificationFailedErr = errors.Join(verificationFailedErrorArray...)
```
shizhMSFT, line 400: It seems not right as it only fails one referrers page. The verification process will attempt the next page.

Two-Hearts: Nice catch! Updated.

verifier/truststore/errors.go
```diff
@@ -0,0 +1,54 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package truststore
+
+// ErrorTrustStore is used when accessing specified trust store failed
+type ErrorTrustStore struct {
```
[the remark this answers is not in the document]
Two-Hearts: I updated the new error types, they are under `[verifier/truststore/errors.go](https://github.com/notaryproject/notation-go/pull/345/files#diff-7f18bd96a3672bb001f5df982e3d300c7b1b35b51423c181a96834a9f07a0470)`.
Kept the original naming (errors in notation-go/errors.go) to avoid any breaking change.

shizhMSFT (dismissed):
LGTM with suggestions

notation.go
```diff
@@ -416,8 +416,9 @@ func Verify(ctx context.Context, verifier Verifier, repo registry.Repository, ve
 	}
 
 	// Verification Failed
-	if len(verificationOutcomes) == 0 {
+	if !verificationSucceeded {
 		logger.Debugf("Signature verification failed for all the signatures associated with artifact %v", artifactDescriptor.Digest)
+		verificationFailedErr = errors.Join(verificationFailedErrorArray...)
```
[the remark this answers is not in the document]
Two-Hearts: Even better, we could remove line 352, and return `errors.Join(verificationFailedErrorArray...)` directly.

shizhMSFT (dismissed):
LGTM

priteshbandi:
Couple of observations:
1. We should't return `verificationOutcomes` for all the signatures we have evaluated; In endless data attack and system will crash with OOM error.
2.  There are some precondition errors such as ts/tp misconfiguration for which we should shortcircuit and not evaluate signatures. More discussion in https://github.com/notaryproject/notation/pull/771#issuecomment-1731705600

Two-Hearts:
> Couple of observations:
> 
> 1. IMO we should't return `verificationOutcomes` for all the signatures we have evaluated; In endless data attack and system will crash with OOM error. Before adding adding more functionality based on `verificationOutcomes`, we should disuses how we want to mitigate the aforementioned issue.

@priteshbandi `verificationOutcomes` related logic is NOT changed in this PR. Sorry that I forgot to update this PR's description, but you could verify it in the code. All this PR doing is updating trust store error messages. 
(btw, we have the `maxSignatureAttempts`. By default, this value is set to `100` from the notation CLI side. So endless attack is not a concern here.)

Two-Hearts:
Hi @priteshbandi, regarding your concern on returning errors for all the signatures, this won't bring in endless attack: https://github.com/notaryproject/notation-go/blob/main/notation.go#L361, because we use `MaxSignatureAttempts` to limit the maximum number of signatures processed by notation. The number of errors returned will always be smaller or equal to `MaxSignatureAttempts`. As a library, returning these errors is an enhancement to notation-go. The caller, such as Notation CLI, can decide whether to display these errors.
(It's backward compatible because we are NOT updating `verificationOutcomes`, instead we use the new Golang feature [errors.Join](https://pkg.go.dev/errors#Join) to do it.)
/cc: @shizhMSFT

JeyJeyGao (dismissed):
LGTM

verifier/truststore/errors.go
```diff
@@ -0,0 +1,54 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package truststore
+
+// TrustStoreError is used when accessing specified trust store failed
+type TrustStoreError struct {
```
priteshbandi, line 17: We have two errors of similar time and I do see need of this kind of errors in future were we want user friendly error message but still want to preserve orignal error. Should we create an generic WrappedError struct?

verifier/truststore/truststore.go
```diff
@@ -64,49 +64,49 @@ type x509TrustStore struct {
 // GetCertificates returns certificates under storeType/namedStore
 func (trustStore *x509TrustStore) GetCertificates(ctx context.Context, storeType Type, namedStore string) ([]*x509.Certificate, error) {
 	if !isValidStoreType(storeType) {
-		return nil, fmt.Errorf("unsupported store type: %s", storeType)
+		return nil, TrustStoreError{Msg: fmt.Sprintf("unsupported trust store type: %s", storeType)}
 	}
 	if !file.IsValidFileName(namedStore) {
-		return nil, errors.New("named store name needs to follow [a-zA-Z0-9_.-]+ format")
+		return nil, TrustStoreError{Msg: fmt.Sprintf("named store name needs to follow [a-zA-Z0-9_.-]+ format: %s is invalid", namedStore)}
```
priteshbandi, line 70:
```suggestion
		return nil, TrustStoreError{Msg: fmt.Sprintf("named store name needs to follow [a-zA-Z0-9_.-]+ format, %s is invalid", namedStore)}
```

verifier/truststore/truststore.go
```diff
@@ -64,49 +64,49 @@ type x509TrustStore struct {
...
 	}
 	path, err := trustStore.trustStorefs.SysPath(dir.X509TrustStoreDir(string(storeType), namedStore))
 	if err != nil {
-		return nil, err
+		return nil, TrustStoreError{InnerError: fmt.Errorf("failed to get path of trust store %s with type %s: %w", namedStore, storeType, err)}
```
priteshbandi, line 74: This error message is not actionable or helpful to user. should we just say:
```suggestion
		return nil, TrustStoreError{InnerError: err, Msg: fmt.Sprintf("Something went wrong while trying to read trust store %q of type %q", namedStore, storeType)}
```

verifier/truststore/truststore.go
```diff
@@ -64,49 +64,49 @@ type x509TrustStore struct {
...
 	}
 	// throw error if path is not a directory or is a symlink or does not exist.
 	fileInfo, err := os.Lstat(path)
 	if err != nil {
 		if os.IsNotExist(err) {
-			return nil, fmt.Errorf("%q does not exist", path)
+			return nil, TrustStoreError{InnerError: err, Msg: fmt.Sprintf("the trust store %q of type %q doesn't exist", namedStore, storeType)}
```
priteshbandi, line 80:
```suggestion
			return nil, TrustStoreError{InnerError: err, Msg: fmt.Sprintf("the trust store %q of type %q does not exists", namedStore, storeType)}
```

verifier/truststore/truststore.go
```diff
@@ -64,49 +64,49 @@ type x509TrustStore struct {
...
 		}
-		return nil, err
+		return nil, TrustStoreError{InnerError: fmt.Errorf("failed to access the trust store %q: %w", path, err)}
```
priteshbandi, line 82: Why are we not creating TrustStoreError like this

`
return nil, TrustStoreError{InnerError: err, Msg: fmt.Stringf("failed to access the trust store %q of type %q", namedStore, storeType )}
`

Also path is location to trust store.

verifier/truststore/truststore.go
```diff
@@ -64,49 +64,49 @@ type x509TrustStore struct {
...
 	}
 	mode := fileInfo.Mode()
 	if !mode.IsDir() || mode&fs.ModeSymlink != 0 {
-		return nil, fmt.Errorf("%q is not a regular directory (symlinks are not supported)", path)
+		return nil, TrustStoreError{Msg: fmt.Sprintf("trust store %q is not a regular directory (symlinks are not supported)", path)}
```
priteshbandi, line 86: to be more specific in error message lets add type and name .

verifier/truststore/truststore.go
```diff
@@ -64,49 +64,49 @@ type x509TrustStore struct {
...
-		return nil, err
+		return nil, TrustStoreError{InnerError: fmt.Errorf("failed to access the trust store %q: %w", path, err)}
 	}
 
 	var certificates []*x509.Certificate
 	for _, file := range files {
 		joinedPath := filepath.Join(path, file.Name())
 		if file.IsDir() || file.Type()&fs.ModeSymlink != 0 {
-			return nil, fmt.Errorf("%q is not a regular file (directories or symlinks are not supported)", joinedPath)
+			return nil, CertificateError{Msg: fmt.Sprintf("trusted certificate %q is not a regular file (directories or symlinks are not supported)", joinedPath)}
```
priteshbandi, line 97: Can we update message to be:
"Certificate file %s in truststore %s of type % is not a regular file...."

verifier/truststore/truststore.go
yizha1: @priteshbandi is this too broad by saying "Something wrong"? Users may get confused without any reason, ideally, for an error message we should have what happened, what is the reason and suggested actions. In this case, do you mean, the reason is not clear, or reason is clear, but we don't have suggested actions?

verifier/truststore/errors.go
Two-Hearts: @priteshbandi Not sure if I understand your comment, what do you mean by `two errors of similar time`? 
The original error is preserved. For example when `TrustStoreError` is created, it would be `TrustStoreError{InnerError: fmt.Errorf("failed to access the trust store %q: %w", path, err)}` , note, the original error is wrapped. If the caller wants the inner error, they can just call `TrustStoreError`'s `Unwrap()` method.

verifier/truststore/truststore.go
Two-Hearts: @priteshbandi Note the detailed error is wrapped in the error. Therefore, the real reason of why failed to get path of trust store is shown to the user.
In addition, due to the error type `TrustStoreError`, caller such as Notation CLI is able to handle this type and show actionable instructions to the user, for example, `Use command 'notation cert add' to create and add trusted certificates to the trust store.` In other words, different callers might have different actionable step based on notation-go's error message.

verifier/truststore/truststore.go
Two-Hearts: That's a good idea. Updating the code.

verifier/truststore/errors.go
Two-Hearts: Oh, and one reason to separate `TrustStoreError` and `CertificateError` is that they might have different actionable next steps. So the caller such as Notation CLI needs a way to tell apart these two types of errors.

verifier/truststore/truststore.go
Two-Hearts: Updated.

verifier/truststore/truststore.go
Two-Hearts: Updated.

verifier/truststore/truststore.go
Two-Hearts: Updated.

verifier/truststore/truststore.go
Two-Hearts: Updated.

shizhMSFT (approved):
LGTM

JeyJeyGao (approved):
LGTM

priteshbandi (approved):
LGTM

PR #347, opened 2023-08-31
bump: update oras-go to v2.3.0

### LINKED ISSUE #346: oras-go updated v2.3.0
### Is your feature request related to a problem?

We need to update to oras-go v2.3.0 and apply new feature `oras.PackManifest` 

### What solution do you propose?

update to oras-go v2.3.0

### What alternatives have you considered?

No

### Any additional context?

_No response_

JeyJeyGao:
- Update oras-go to v2.3.0.
- Replace oras.Pack() with oras.PackManifest() as it is deprecated in v2.3.0.
- Generate an empty config blob manually, as oras.PackManifest() does not generate the config blob with the notation artifact type as the media type.

Resolves #346 

### DIFF
--- a/registry/repository.go
+++ b/registry/repository.go
@@ -35,6 +35,19 @@ const (
 	maxManifestSizeLimit = 4 * 1024 * 1024  // 4 MiB
 )
 
+var (
+	// notationEmptyConfigDesc is the descriptor of an empty notation manifest
+	// config
+	// reference: https://github.com/notaryproject/specifications/blob/v1.0.0/specs/signature-specification.md#storage
+	notationEmptyConfigDesc = ocispec.Descriptor{
+		MediaType: ArtifactTypeNotation,
+		Digest:    ocispec.DescriptorEmptyJSON.Digest,
+		Size:      ocispec.DescriptorEmptyJSON.Size,
+	}
+	// notationEmptyConfigData is the data of an empty notation manifest config
+	notationEmptyConfigData = ocispec.DescriptorEmptyJSON.Data
+)
+
 // RepositoryOptions provides user options when creating a Repository
 // it is kept for future extensibility
 type RepositoryOptions struct{}
@@ -192,14 +205,14 @@ func (c *repositoryClient) getSignatureBlobDesc(ctx context.Context, sigManifest
 func (c *repositoryClient) uploadSignatureManifest(ctx context.Context, subject, blobDesc ocispec.Descriptor, annotations map[string]string) (ocispec.Descriptor, error) {
 	configDesc, err := pushNotationManifestConfig(ctx, c.GraphTarget)
 	if err != nil {
-		return ocispec.Descriptor{}, err
+		return ocispec.Descriptor{}, fmt.Errorf("failed to push notation manifest config: %w", err)
 	}
 
 	opts := oras.PackManifestOptions{
 		Subject:             &subject,
 		ManifestAnnotations: annotations,
 		Layers:              []ocispec.Descriptor{blobDesc},
-		ConfigDescriptor:    configDesc,
+		ConfigDescriptor:    &configDesc,
 	}
 
 	return oras.PackManifest(ctx, c.GraphTarget, oras.PackManifestVersion1_1_RC4, "", opts)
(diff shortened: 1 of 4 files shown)


Two-Hearts (dismissed):
LGTM

registry/repository.go
```diff
@@ -187,13 +190,44 @@ func (c *repositoryClient) getSignatureBlobDesc(ctx context.Context, sigManifest
...
+	}
+
+	return oras.PackManifest(ctx, c.GraphTarget, oras.PackManifestVersion1_1_RC4, "", opts)
+}
+
+// pushNotationManifestConfig pushes an empty notation manifest config, if it
+// doesn't exist.
+func pushNotationManifestConfig(ctx context.Context, pusher content.Pusher) (*ocispec.Descriptor, error) {
+	// generate a empty config descriptor for notation manifest
+	configContent := []byte("{}")
```
qweeah, line 212: how about using [DescriptorEmptyJSON .Data](https://pkg.go.dev/github.com/opencontainers/image-spec@v1.1.0-rc4/specs-go/v1#pkg-variables)

shizhMSFT (dismissed):
LGTM

registry/repository.go
```diff
@@ -187,13 +190,44 @@ func (c *repositoryClient) getSignatureBlobDesc(ctx context.Context, sigManifest
 
 // uploadSignatureManifest uploads the signature manifest to the registry
 func (c *repositoryClient) uploadSignatureManifest(ctx context.Context, subject, blobDesc ocispec.Descriptor, annotations map[string]string) (ocispec.Descriptor, error) {
-	opts := oras.PackOptions{
+	configDesc, err := pushNotationManifestConfig(ctx, c.GraphTarget)
+	if err != nil {
+		return ocispec.Descriptor{}, err
```
Wwwsylvia, line 195: Do we want to return a more informative error message here?

Wwwsylvia (changes requested):

registry/repository.go
JeyJeyGao: Thanks! Updated.

registry/repository.go
```diff
@@ -187,13 +190,44 @@ func (c *repositoryClient) getSignatureBlobDesc(ctx context.Context, sigManifest
...
+
+	return oras.PackManifest(ctx, c.GraphTarget, oras.PackManifestVersion1_1_RC4, "", opts)
+}
+
+// pushNotationManifestConfig pushes an empty notation manifest config, if it
+// doesn't exist.
+func pushNotationManifestConfig(ctx context.Context, pusher content.Pusher) (*ocispec.Descriptor, error) {
+	// generate a empty config descriptor for notation manifest
+	configContent := []byte("{}")
+	desc := content.NewDescriptorFromBytes(ArtifactTypeNotation, configContent)
```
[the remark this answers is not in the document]
JeyJeyGao: updated to use [DescriptorEmptyJSON.Data](https://pkg.go.dev/github.com/opencontainers/image-spec@v1.1.0-rc4/specs-go/v1#pkg-variables)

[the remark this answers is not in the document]
JeyJeyGao: updated to add `var` as the global variable

registry/repository.go
```diff
@@ -187,13 +190,44 @@ func (c *repositoryClient) getSignatureBlobDesc(ctx context.Context, sigManifest
...
+		Layers:              []ocispec.Descriptor{blobDesc},
+		ConfigDescriptor:    configDesc,
+	}
+
+	return oras.PackManifest(ctx, c.GraphTarget, oras.PackManifestVersion1_1_RC4, "", opts)
+}
+
+// pushNotationManifestConfig pushes an empty notation manifest config, if it
+// doesn't exist.
+func pushNotationManifestConfig(ctx context.Context, pusher content.Pusher) (*ocispec.Descriptor, error) {
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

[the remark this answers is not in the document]
JeyJeyGao: Updated.

registry/repository.go
JeyJeyGao: Updated.

registry/repository.go
```diff
@@ -187,13 +190,44 @@ func (c *repositoryClient) getSignatureBlobDesc(ctx context.Context, sigManifest
...
+	// generate a empty config descriptor for notation manifest
+	configContent := []byte("{}")
+	desc := content.NewDescriptorFromBytes(ArtifactTypeNotation, configContent)
+
+	// check if the config exists
+	if ros, ok := pusher.(content.ReadOnlyStorage); ok {
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

shizhMSFT (dismissed):
LGTM

registry/repository.go
```diff
@@ -187,13 +203,38 @@ func (c *repositoryClient) getSignatureBlobDesc(ctx context.Context, sigManifest
...
+		return ocispec.Descriptor{}, fmt.Errorf("unable to verify existence: %s: %s. Details: %w", notationEmptyConfigDesc.Digest.String(), notationEmptyConfigDesc.MediaType, err)
+	}
+	if exists {
+		return notationEmptyConfigDesc, nil
+	}
+
+	// push the config
+	if err := pusher.Push(ctx, notationEmptyConfigDesc, bytes.NewReader(notationEmptyConfigData)); err != nil && !errors.Is(err, errdef.ErrAlreadyExists) {
+		return ocispec.Descriptor{}, fmt.Errorf("unable to push: %s: %s. Details: %w", notationEmptyConfigDesc.Digest.String(), notationEmptyConfigDesc.MediaType, err)
+	}
```
Two-Hearts, line 236: What if `err` is `errdef.ErrAlreadyExists`? We are ignoring this error.
Since we already checked existence above, can I say `!errors.Is(err, errdef.ErrAlreadyExists)` is always true? If yes, then, how about,
```suggestion
	if err := pusher.Push(ctx, notationEmptyConfigDesc, bytes.NewReader(notationEmptyConfigData)); err != nil {
		return ocispec.Descriptor{}, fmt.Errorf("unable to push: %s: %s. Details: %w", notationEmptyConfigDesc.Digest.String(), notationEmptyConfigDesc.MediaType, err)
	}
```

Wwwsylvia: We do want to ignore the `errdef.ErrAlreadyExists` error as it is not helpful for the user, it's fine as long as the config is there. It is also possible that `Exists` returns `false` at the time we checks it, but another process uploads a same config before we do `Push`.

Wwwsylvia (approved):
LGTM

registry/repository.go
Two-Hearts: Then we should leave a comment in the code stating this. Currently, the function could return `nil` even though the error is `non-nil`.

JeyJeyGao: Added comments.

Wwwsylvia (approved):
LGTM

Two-Hearts (approved):
LGTM

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

PR #354, opened 2023-09-28
Verifier: Add fuzzer from cncf-fuzzing

AdamKorcz:
Moves the fuzzer over to notation-go [from cncf-fuzzing](https://github.com/cncf/cncf-fuzzing/blob/main/projects/notary/fuzz_verification.go).

I will instruct OSS-Fuzz to run it from here, once it has been merged.


shizhMSFT:
@AdamKorcz Thanks for the PR. Could you fix the check as it is failed?

PR #368, opened 2023-12-23
chore: start using plugin-framework package

priteshbandi:
chore: start using plugin-framework package.
## https://github.com/notaryproject/notation-go/pull/372 supersedes this PR (which I have converted to draft)

_**Note:** this PR has breaking changes but they are required in order to add support for plugin as library._

~This PR depends on https://github.com/notaryproject/notation-plugin-framework-go/pull/5~


plugin/plugin.go
```diff
@@ -31,53 +31,19 @@ import (
 	"github.com/notaryproject/notation-go/internal/slices"
 	"github.com/notaryproject/notation-go/log"
 	"github.com/notaryproject/notation-go/plugin/proto"
+	"github.com/notaryproject/notation-plugin-framework-go/plugin"
 )
 
 var executor commander = &execCommander{} // for unit test
 
-// GenericPlugin is the base requirement to be an plugin.
-type GenericPlugin interface {
```
shizhMSFT, line 39: We shall not remove any interfaces, types, or functions as `notation-go` is stable at `v1`.

Instead, consider this
```go
type GenericPlugin = pluginframework.GenericPlugin
```

If the breaking change is not avoidable, you might need to consider `notation-go/v2` but it would be another proposal to be discussed.

shizhMSFT: The same technique is applied when golang introduce the `io/fs` package as well as deprecate `io/ioutil`.

Examples can be found at https://pkg.go.dev/os#FileInfo
```go
type FileInfo = fs.FileInfo
```

shizhMSFT (changes requested):
Please consider starting moving to plugin framework package only after it reaches a stable version (i.e. `v1.0.0`).

plugin/plugin.go
priteshbandi: I can try to add backward comparability support for interfaces but for other types and functions there would be breaking change. IMO the maintenance effort and code-change required to maintain backward comparability is not be worth it.  
Lets discuss this in next community meeting, as suggested by you one way is to `notation-go/v2`

---

Please consider starting moving to plugin framework package only after it reaches a stable version (i.e. v1.0.0).
> I was thinking of doing v1 for plugin framework package at the same time when we do next release for notation-go.

PR #369, opened 2024-01-05
feat: plugin install iteration 2

Two-Hearts:
This PR is the 2nd iteration of plugin install.
Installation SHOULD pass in any of the following scenario:
1. Installing a single valid plugin executable file on file system. An example can be `notation-com.example.plugin` on `Unix` and `notation-com.example.plugin.exe` on `Windows`.
2. Installing from a dir which contains `one and only one` executable file with file name in the `notation-{plugin-name}` format.
3. On `Unix`, install from a dir which contains `no executable file`, however, there's `one and only one` candidate file with file name in the `notation-{plugin-name}` format. (Notation would set it to executable and try install it with a warning alert to the user). On `Windows`, install would fail in this case since '.exe' file extension is required.

Corresponding tests are added as well.

### DIFF
--- a/plugin/manager.go
+++ b/plugin/manager.go
@@ -247,8 +247,7 @@ func parsePluginFromDir(ctx context.Context, path string) (string, string, error
 		}
 		// only take regular files
 		if info.Mode().IsRegular() {
-			candidatePluginName, err := parsePluginName(d.Name())
-			if err != nil {
+			if candidatePluginName, err = parsePluginName(d.Name()); err != nil {
 				// file name does not follow the notation-{plugin-name} format,
 				// continue
 				return nil
--- a/plugin/manager.go
+++ b/plugin/manager.go
@@ -246,7 +246,8 @@ func parsePluginFromDir(path string) (string, string, error) {
 		}
 		// only take regular files
 		if info.Mode().IsRegular() {
-			if pluginName, err = parsePluginName(d.Name()); err != nil {
+			pName, err := parsePluginName(d.Name())
+			if err != nil {
 				// file name does not follow the notation-{plugin-name} format,
 				// continue
 				return nil
@@ -264,13 +265,13 @@ func parsePluginFromDir(path string) (string, string, error) {
 			}
 			foundPluginExecutableFile = true
 			pluginExecutableFile = p
+			pluginName = pName
 		}
 		return nil
 	}); err != nil {
 		return "", "", err
 	}
 	if !foundPluginExecutableFile {
-		fmt.Println("no plugin executable file", filesWithValidNameFormat)
 		// if no executable file was found, but there's one and only one
 		// potential candidate, try install the candidate
 		if len(filesWithValidNameFormat) == 1 {
--- a/plugin/manager.go
+++ b/plugin/manager.go
@@ -127,7 +127,7 @@ func (m *CLIManager) Install(ctx context.Context, installOpts CLIInstallOptions)
 		// input is not a dir, check if it's a single plugin executable file
 		installFromNonDir = true
 		pluginExecutableFile = installOpts.PluginPath
-		pluginName, err = ParsePluginName(filepath.Base(pluginExecutableFile))
+		pluginName, err = parsePluginName(filepath.Base(pluginExecutableFile))
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to read plugin name from file path %s: %w", pluginExecutableFile, err)
 		}
@@ -140,7 +140,7 @@ func (m *CLIManager) Install(ctx context.Context, installOpts CLIInstallOptions)
 		}
 	}
 	// validate and get new plugin metadata
-	if err := validatePluginFileExtensionAgainstOS(filepath.Base(pluginExecutableFile), pluginName); err != nil {
+	if err := validatePluginFilenameAgainstOS(filepath.Base(pluginExecutableFile), pluginName); err != nil {
 		return nil, nil, err
 	}
 	newPlugin, err := NewCLIPlugin(ctx, pluginName, pluginExecutableFile)
@@ -246,7 +246,7 @@ func parsePluginFromDir(path string) (string, string, error) {
 		}
 		// only take regular files
 		if info.Mode().IsRegular() {
-			if pluginName, err = ParsePluginName(d.Name()); err != nil {
+			if pluginName, err = parsePluginName(d.Name()); err != nil {
 				// file name does not follow the notation-{plugin-name} format,
 				// continue
 				return nil
--- a/plugin/manager.go
+++ b/plugin/manager.go
@@ -270,6 +270,7 @@ func parsePluginFromDir(path string) (string, string, error) {
 		return "", "", err
 	}
 	if !foundPluginExecutableFile {
+		fmt.Println("no plugin executable file", filesWithValidNameFormat)
 		// if no executable file was found, but there's one and only one
 		// potential candidate, try install the candidate
 		if len(filesWithValidNameFormat) == 1 {
--- a/plugin/manager.go
+++ b/plugin/manager.go
@@ -279,7 +279,7 @@ func parsePluginFromDir(ctx context.Context, path string) (string, string, error
 			if err := setExecutable(candidate); err != nil {
 				return "", "", fmt.Errorf("no plugin executable file was found: %w", err)
 			}
-			logger.Warnf("Found candidate plugin executable file %q without executable permission. Setting user executable bit and try install.", filepath.Base(candidate))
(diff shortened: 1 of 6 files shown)


plugin/manager_windows.go
```diff
@@ -36,3 +38,20 @@ func isExecutableFile(filePath string) (bool, error) {
...
 	return strings.EqualFold(filepath.Ext(filepath.Base(filePath)), ".exe"), nil
 }
+
+// parsePluginName checks if fileName is a valid plugin file name
+// and gets plugin name from it based on spec: https://github.com/notaryproject/specifications/blob/main/specs/plugin-extensibility.md#installation
+func parsePluginName(fileName string) (string, error) {
+	fname := file.TrimFileExtension(fileName)
+	pluginName, found := strings.CutPrefix(fname, proto.Prefix)
+	if !found || pluginName == "" {
+		return "", fmt.Errorf("invalid plugin executable file name. Plugin file name requires format notation-{plugin-name}, but got %s", fname)
```
shizhMSFT, line 48:
```suggestion
		return "", fmt.Errorf("invalid plugin executable file name. Plugin file name requires format notation-{plugin-name}.exe, but got %s", fileName)
```

plugin/manager_windows.go
```diff
@@ -36,3 +38,20 @@ func isExecutableFile(filePath string) (bool, error) {
 	}
 	return strings.EqualFold(filepath.Ext(filepath.Base(filePath)), ".exe"), nil
 }
+
+// parsePluginName checks if fileName is a valid plugin file name
+// and gets plugin name from it based on spec: https://github.com/notaryproject/specifications/blob/main/specs/plugin-extensibility.md#installation
+func parsePluginName(fileName string) (string, error) {
+	fname := file.TrimFileExtension(fileName)
+	pluginName, found := strings.CutPrefix(fname, proto.Prefix)
+	if !found || pluginName == "" {
```
shizhMSFT, line 47: Should check filename has `.exe` suffix (case insensitive) in the method. Otherwise, line 250 of `manager.go` may have incorrect logic state on `notation-foo.bar` on Windows.

plugin/manager.go
```diff
@@ -263,12 +265,23 @@ func parsePluginFromDir(path string) (string, string, error) {
...
 	}
 	if !foundPluginExecutableFile {
+		// if no executable file was found, but there's one and only one
+		// potential candidate, try install the candidate
+		if len(filesWithValidNameFormat) == 1 {
+			candidate := filesWithValidNameFormat[0]
+			if err := setExecutable(candidate); err != nil {
+				return "", "", fmt.Errorf("no plugin executable file was found: %w", err)
+			}
+			logger.Warnf("Found candidate plugin executable file %q without executable permission. Setting user executable bit and try install.", filepath.Base(candidate))
```
shizhMSFT, line 282: nit:
```suggestion
			logger.Warnf("Found candidate plugin executable file %q without executable permission. Setting user executable bit and trying to install.", filepath.Base(candidate))
```

plugin/manager_windows.go
```diff
@@ -42,16 +42,19 @@ func isExecutableFile(filePath string) (bool, error) {
 // parsePluginName checks if fileName is a valid plugin file name
 // and gets plugin name from it based on spec: https://github.com/notaryproject/specifications/blob/main/specs/plugin-extensibility.md#installation
 func parsePluginName(fileName string) (string, error) {
-	fname := file.TrimFileExtension(fileName)
+	fname := fileName
+	if strings.EqualFold(filepath.Ext(fileName), ".exe") {
```
shizhMSFT, line 46: If it returns `false`, should it fail?

Two-Hearts: @shizhMSFT This function is getting the plugin name from file name only. Checking whether the plugin file is executable is done in the next step of the workflow.

shizhMSFT (dismissed):
LGTM

plugin/manager.go
```diff
@@ -119,30 +120,27 @@ func (m *CLIManager) Install(ctx context.Context, installOpts CLIInstallOptions)
...
+			return nil, nil, fmt.Errorf("failed to read plugin name from input file: %w", err)
 		}
 		isExec, err := isExecutableFile(pluginExecutableFile)
 		if err != nil {
-			return nil, nil, fmt.Errorf("failed to check if file %s is executable: %w", pluginExecutableFile, err)
+			return nil, nil, fmt.Errorf("failed to check if input file is executable: %w", err)
 		}
 		if !isExec {
-			return nil, nil, fmt.Errorf("file %s is not executable", pluginExecutableFile)
+			return nil, nil, errors.New("input file is not executable")
```
rgnote, line 140: Why not show the file path for better debuggability?

plugin/plugin.go
```diff
@@ -225,17 +223,6 @@ func (c execCommander) Output(ctx context.Context, name string, command proto.Co
 	return stdout.Bytes(), nil, nil
 }
 
-// ParsePluginName checks if fileName is a valid plugin file name
-// and gets plugin name from it based on spec: https://github.com/notaryproject/specifications/blob/main/specs/plugin-extensibility.md#installation
-func ParsePluginName(fileName string) (string, error) {
```
rgnote, line 230: Isn't removing this a breaking change? Should we mark this as obsolete instead?

Two-Hearts: @rgnote This won't be a breaking change as it's in the `main` branch ONLY. In other words, should we release the next version `v1.1.0`, removing it would become a breaking change.

plugin/manager.go
Two-Hearts: @rgnote The added log is at line 123 of `plugin/manager.go`

plugin/plugin.go
rgnote: Right. Won't we release this code at some point? We will, right. Unless the next release is a major version upgrade, we should not let this into main branch.

plugin/manager.go
rgnote: We can include the file name

plugin/plugin.go
Two-Hearts: > Right. Won't we release this code at some point? We will, right. Unless the next release is a major version upgrade, we should not let this into main branch.

Please note, this code is not in any Notation release yet, therefore, this change is not a breaking change. /cc: @shizhMSFT

plugin/manager.go
Two-Hearts: Added the file name to the error message.

plugin/plugin.go
rgnote: > it won't be a breaking change when we release v1.1.0, because to v1.1.0, it's a new feature

Can you please elaborate more on this? If this change is included in v1.1.0, then any consumers of `ParsePluginName` API will break, right?

Two-Hearts: > > it won't be a breaking change when we release v1.1.0, because to v1.1.0, it's a new feature
> 
> Can you please elaborate more on this? If this change is included in v1.1.0, then any consumers of `ParsePluginName` API will break, right?

We don't have any consumer of `ParsePluginName` because it's never been released by Notation: https://github.com/notaryproject/notation-go/releases. Therefore, users will start to use this function only after we release `v1.1.0`.

rgnote: > We don't have any consumer of ParsePluginName yet because it's never been released by Notation: https://github.com/notaryproject/notation-go/releases.

Got it. Thanks.

rgnote (approved):
LGTM

plugin/manager.go
```diff
@@ -263,12 +268,23 @@ func parsePluginFromDir(path string) (string, string, error) {
...
+		// potential candidate, try install the candidate
+		if len(filesWithValidNameFormat) == 1 {
+			candidate := filesWithValidNameFormat[0]
+			if err := setExecutable(candidate); err != nil {
+				return "", "", fmt.Errorf("no plugin executable file was found: %w", err)
+			}
+			logger.Warnf("Found candidate plugin executable file %q without executable permission. Setting user executable bit and trying to install.", filepath.Base(candidate))
+			return candidate, candidatePluginName, nil
+		}
 		return "", "", errors.New("no plugin executable file was found")
```
JeyJeyGao, line 288: What happens if there are two files without executable permission: notation-plugin-a and notation-plugin-b? In this case, what error should be returned?

shizhMSFT (approved):
LGTM

plugin/manager.go
Two-Hearts: @JeyJeyGao In this case, the install would fail with error: `failed to read plugin from input directory: no plugin executable file was found`

Two-Hearts: The corresponding test is covered at here: https://github.com/notaryproject/notation-go/blob/3d8621dbf47107d04186ebf956b0b837783834c7/plugin/manager_test.go#L504

JeyJeyGao (approved):
LGTM

PR #379, opened 2024-01-26
feat: add support for signing blob

priteshbandi:
This PR adds support for signing blobs.

* This PR is ready for review**

### DIFF
--- a/notation.go
+++ b/notation.go
@@ -22,8 +22,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"mime"
 	"strings"
 	"time"
+
+	orasRegistry "oras.land/oras-go/v2/registry"
 
 	"github.com/notaryproject/notation-core-go/signature"
 	"github.com/notaryproject/notation-go/internal/envelope"
@@ -32,7 +36,6 @@ import (
 	"github.com/notaryproject/notation-go/verifier/trustpolicy"
 	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
-	orasRegistry "oras.land/oras-go/v2/registry"
 )
 
 var errDoneVerification = errors.New("done verification")
@@ -65,6 +68,12 @@ type Signer interface {
 	Sign(ctx context.Context, desc ocispec.Descriptor, opts SignerSignOptions) ([]byte, *signature.SignerInfo, error)
 }
 
+type BlobSigner interface {
+	// SignBlob signs the artifact described by its descriptor,
+	// and returns the signature and SignerInfo.
+	SignBlob(ctx context.Context, reader io.Reader, opts SignBlobOptions) ([]byte, *signature.SignerInfo, error)
+}
+
 // signerAnnotation facilitates return of manifest annotations by signers
 type signerAnnotation interface {
 	// PluginAnnotations returns signature manifest annotations returned from
@@ -85,21 +94,25 @@ type SignOptions struct {
 	UserMetadata map[string]string
 }
 
+type SignBlobOptions struct {
+	SignerSignOptions
+
+	ContentMediaType string
+
+	// UserMetadata contains key-value pairs that are added to the signature
+	// payload
+	UserMetadata map[string]string
+}
+
 // Sign signs the artifact and push the signature to the Repository.
-// The descriptor of the sign content is returned upon sucessful signing.
+// The descriptor of the sign content is returned upon successful signing.
 func Sign(ctx context.Context, signer Signer, repo registry.Repository, signOpts SignOptions) (ocispec.Descriptor, error) {
-	// sanity check
-	if signer == nil {
-		return ocispec.Descriptor{}, errors.New("signer cannot be nil")
+	// sanity checks
+	if err := validate(signer, signOpts.SignerSignOptions); err != nil {
+		return ocispec.Descriptor{}, err
 	}
 	if repo == nil {
 		return ocispec.Descriptor{}, errors.New("repo cannot be nil")
-	}
-	if signOpts.ExpiryDuration < 0 {
-		return ocispec.Descriptor{}, fmt.Errorf("expiry duration cannot be a negative value")
-	}
-	if signOpts.ExpiryDuration%time.Second != 0 {
-		return ocispec.Descriptor{}, fmt.Errorf("expiry duration supports minimum granularity of seconds")
 	}
 
 	logger := log.GetLogger(ctx)
@@ -150,6 +163,44 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, signOpts
 	}
 
 	return targetDesc, nil
+}
+
+func SignBlob(ctx context.Context, signer BlobSigner, reader io.Reader, signOpts SignBlobOptions) ([]byte, error) {
+	// sanity checks
+	if err := validate(signer, signOpts.SignerSignOptions); err != nil {
+		return nil, err
+	}
+	if reader == nil {
+		return nil, errors.New("reader cannot be nil")
+	}
+	if signOpts.ContentMediaType != "" {
+		if _, _, err := mime.ParseMediaType(signOpts.ContentMediaType); err != nil {
+			return nil, fmt.Errorf("invalid content media-type '%s': %v", signOpts.ContentMediaType, err)
+		}
+	}
+
+	sig, _, err := signer.SignBlob(ctx, reader, signOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	return sig, nil
+}
+
+func validate(signer any, signOpts SignerSignOptions) error {
+	if signer == nil {
+		return errors.New("signer cannot be nil")
+	}
+	if signOpts.ExpiryDuration < 0 {
+		return fmt.Errorf("expiry duration cannot be a negative value")
+	}
+	if signOpts.ExpiryDuration%time.Second != 0 {
+		return fmt.Errorf("expiry duration supports minimum granularity of seconds")
+	}
+	if _, _, err := mime.ParseMediaType(signOpts.SignatureMediaType); err != nil {
+		return fmt.Errorf("invalid signature media-type '%s': %v", signOpts.SignatureMediaType, err)
+	}
+	return nil
 }
 
 func addUserMetadataToDescriptor(ctx context.Context, desc ocispec.Descriptor, userMetadata map[string]string) (ocispec.Descriptor, error) {
--- a/notation.go
+++ b/notation.go
@@ -73,13 +73,14 @@ type Signer interface {
 // SignBlobOptions contains parameters for notation.SignBlob.
 type SignBlobOptions struct {
 	SignerSignOptions
-
+	// ContentMediaType is the media-type of the blob being signed.
 	ContentMediaType string
 	// UserMetadata contains key-value pairs that are added to the signature
 	// payload
 	UserMetadata map[string]string
 }
 
+// BlobDescriptorGenerator creates descriptor using the digest Algorithm.
 type BlobDescriptorGenerator func(digest.Algorithm) (ocispec.Descriptor, error)
 
 // BlobSigner is a generic interface for signing arbitrary data.
@@ -115,7 +116,7 @@ type SignOptions struct {
 // The descriptor of the sign content is returned upon successful signing.
 func Sign(ctx context.Context, signer Signer, repo registry.Repository, signOpts SignOptions) (ocispec.Descriptor, error) {
 	// sanity check
-	if err := validate(signer, signOpts.SignerSignOptions); err != nil {
+	if err := validateSignArguments(signer, signOpts.SignerSignOptions); err != nil {
 		return ocispec.Descriptor{}, err
 	}
 	if repo == nil {
@@ -175,7 +176,7 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, signOpts
 // SignBlob signs the arbitrary data and returns the signature
 func SignBlob(ctx context.Context, signer BlobSigner, blobReader io.Reader, signBlobOpts SignBlobOptions) ([]byte, *signature.SignerInfo, error) {
 	// sanity checks
-	if err := validate(signer, signBlobOpts.SignerSignOptions); err != nil {
+	if err := validateSignArguments(signer, signBlobOpts.SignerSignOptions); err != nil {
 		return nil, nil, err
 	}
 
@@ -191,7 +192,7 @@ func SignBlob(ctx context.Context, signer BlobSigner, blobReader io.Reader, sign
 		return nil, nil, fmt.Errorf("invalid content media-type '%s': %v", signBlobOpts.ContentMediaType, err)
 	}
 
-	getDescFun := func(hashAlgo digest.Algorithm) (ocispec.Descriptor, error) {
+	getDescFunc := func(hashAlgo digest.Algorithm) (ocispec.Descriptor, error) {
 		h := hashAlgo.Hash()
 		bytes, err := io.Copy(hashAlgo.Hash(), blobReader)
 		if err != nil {
@@ -206,10 +207,10 @@ func SignBlob(ctx context.Context, signer BlobSigner, blobReader io.Reader, sign
 		return addUserMetadataToDescriptor(ctx, targetDesc, signBlobOpts.UserMetadata)
 	}
 
-	return signer.SignBlob(ctx, getDescFun, signBlobOpts.SignerSignOptions)
+	return signer.SignBlob(ctx, getDescFunc, signBlobOpts.SignerSignOptions)
 }
 
-func validate(signer any, signOpts SignerSignOptions) error {
+func validateSignArguments(signer any, signOpts SignerSignOptions) error {
 	if signer == nil {
 		return errors.New("signer cannot be nil")
 	}
--- a/notation.go
+++ b/notation.go
@@ -517,14 +517,14 @@ func generateAnnotations(signerInfo *signature.SignerInfo, annotations map[strin
 
 func getDescriptorFunc(ctx context.Context, reader io.Reader, contentMediaType string, userMetadata map[string]string) BlobDescriptorGenerator {
 	return func(hashAlgo digest.Algorithm) (ocispec.Descriptor, error) {
-		h := hashAlgo.Hash()
-		bytes, err := io.Copy(hashAlgo.Hash(), reader)
+		digester := hashAlgo.Digester()
+		bytes, err := io.Copy(digester.Hash(), reader)
 		if err != nil {
 			return ocispec.Descriptor{}, err
 		}
 		targetDesc := ocispec.Descriptor{
 			MediaType: contentMediaType,
-			Digest:    digest.NewDigest(hashAlgo, h),
+			Digest:    digester.Digest(),
 			Size:      bytes,
 		}
 		return addUserMetadataToDescriptor(ctx, targetDesc, userMetadata)
--- a/notation.go
+++ b/notation.go
@@ -192,21 +192,7 @@ func SignBlob(ctx context.Context, signer BlobSigner, blobReader io.Reader, sign
 		return nil, nil, fmt.Errorf("invalid content media-type '%s': %v", signBlobOpts.ContentMediaType, err)
 	}
 
-	getDescFunc := func(hashAlgo digest.Algorithm) (ocispec.Descriptor, error) {
-		h := hashAlgo.Hash()
-		bytes, err := io.Copy(hashAlgo.Hash(), blobReader)
-		if err != nil {
-			return ocispec.Descriptor{}, err
-		}
-
-		targetDesc := ocispec.Descriptor{
-			MediaType: signBlobOpts.ContentMediaType,
-			Digest:    digest.NewDigest(hashAlgo, h),
-			Size:      bytes,
-		}
-		return addUserMetadataToDescriptor(ctx, targetDesc, signBlobOpts.UserMetadata)
-	}
-
+	getDescFunc := getDescriptorFunc(ctx, blobReader, signBlobOpts.ContentMediaType, signBlobOpts.UserMetadata)
 	return signer.SignBlob(ctx, getDescFunc, signBlobOpts.SignerSignOptions)
 }
 
@@ -315,7 +301,7 @@ func (outcome *VerificationOutcome) UserMetadata() (map[string]string, error) {
 
 // VerifierVerifyOptions contains parameters for Verifier.Verify.
 type VerifierVerifyOptions struct {
-	// ArtifactReference is the reference of the artifact that is been
+	// ArtifactReference is the reference of the artifact that is being
 	// verified against to. It must be a full reference.
 	ArtifactReference string
 
@@ -349,7 +335,7 @@ type verifySkipper interface {
 
 // VerifyOptions contains parameters for notation.Verify.
 type VerifyOptions struct {
-	// ArtifactReference is the reference of the artifact that is been
+	// ArtifactReference is the reference of the artifact that is being
 	// verified against to.
 	ArtifactReference string
 
@@ -527,4 +513,20 @@ func generateAnnotations(signerInfo *signature.SignerInfo, annotations map[strin
 	}
 	annotations[ocispec.AnnotationCreated] = signingTime.Format(time.RFC3339)
 	return annotations, nil
+}
+
+func getDescriptorFunc(ctx context.Context, reader io.Reader, contentMediaType string, userMetadata map[string]string) BlobDescriptorGenerator {
+	return func(hashAlgo digest.Algorithm) (ocispec.Descriptor, error) {
+		h := hashAlgo.Hash()
+		bytes, err := io.Copy(hashAlgo.Hash(), reader)
+		if err != nil {
+			return ocispec.Descriptor{}, err
+		}
+		targetDesc := ocispec.Descriptor{
+			MediaType: contentMediaType,
+			Digest:    digest.NewDigest(hashAlgo, h),
+			Size:      bytes,
+		}
+		return addUserMetadataToDescriptor(ctx, targetDesc, userMetadata)
+	}
 }
--- a/notation.go
+++ b/notation.go
@@ -30,6 +30,8 @@ import (
 	orasRegistry "oras.land/oras-go/v2/registry"
 
 	"github.com/notaryproject/notation-core-go/signature"
+	"github.com/notaryproject/notation-core-go/signature/cose"
+	"github.com/notaryproject/notation-core-go/signature/jws"
 	"github.com/notaryproject/notation-go/internal/envelope"
 	"github.com/notaryproject/notation-go/log"
 	"github.com/notaryproject/notation-go/registry"
@@ -44,7 +46,7 @@ var reservedAnnotationPrefixes = [...]string{"io.cncf.notary"}
 // SignerSignOptions contains parameters for Signer.Sign.
 type SignerSignOptions struct {
 	// SignatureMediaType is the envelope type of the signature.
-	// Currently both `application/jose+json` and `application/cose` are
+	// Currently, both `application/jose+json` and `application/cose` are
 	// supported.
 	SignatureMediaType string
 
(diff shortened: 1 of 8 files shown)


notation.go
```diff
@@ -65,6 +68,12 @@ type Signer interface {
 	Sign(ctx context.Context, desc ocispec.Descriptor, opts SignerSignOptions) ([]byte, *signature.SignerInfo, error)
 }
 
+type BlobSigner interface {
```
JeyJeyGao, line 71: Please add type document for it.

notation.go
```diff
@@ -85,22 +94,26 @@ type SignOptions struct {
 	UserMetadata map[string]string
 }
 
+type SignBlobOptions struct {
```
JeyJeyGao, line 97: Need doc for it.

notation.go
```diff
@@ -85,22 +94,26 @@ type SignOptions struct {
...
+	SignerSignOptions
+
+	ContentMediaType string
```
JeyJeyGao, line 100: Need doc.

notation.go
```diff
@@ -152,6 +165,44 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, signOpts
 	return targetDesc, nil
 }
 
+func SignBlob(ctx context.Context, signer BlobSigner, reader io.Reader, signOpts SignBlobOptions) ([]byte, error) {
```
JeyJeyGao, line 168: Need function doc.

notation.go
```diff
@@ -152,6 +165,44 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, signOpts
...
+	if signer == nil {
+		return errors.New("signer cannot be nil")
+	}
+	if signOpts.ExpiryDuration < 0 {
+		return fmt.Errorf("expiry duration cannot be a negative value")
+	}
+	if signOpts.ExpiryDuration%time.Second != 0 {
+		return fmt.Errorf("expiry duration supports minimum granularity of seconds")
+	}
+	if _, _, err := mime.ParseMediaType(signOpts.SignatureMediaType); err != nil {
```
JeyJeyGao, line 200: Do we need to validate the `SignatureMediaType` is equal to `jws.MediaTypeEnvelope` or `cose.MediaTypeEnvelope` directly?

notation.go
Two-Hearts, line 71: I'm thinking if we could reuse the existing `Signer` interface, `SignOptions` struct and func `notation.Sign`, so the design could be:
```
// SignOptions contains parameters for notation.Sign.
type SignOptions struct {
	SignerSignOptions

        // BlobSign, if set to true, specifies a blob signing
        BlobSign bool

        // BlobSignDescriptor is the blob descriptor to be signed.
        // It is only used when 'BlobSign' is true.
        BlobSignDescriptor *ocispec.Descriptor

	// ArtifactReference sets the reference of the artifact that needs to be
	// signed. It can be a tag, a digest or a full reference.
        // It is only used when 'BlobSign' is false.
	ArtifactReference string

	// UserMetadata contains key-value pairs that are added to the signature
	// payload
	UserMetadata map[string]string
}
```
In the above design, we are adding new fields into existing struct without any breaking change. Callers just need to set the flag to choose between signing a blob and an oci image. 
Callers are also allowed to provide their own descriptor to get signed. For example, Notation CLI will create such a descriptor out of the `blob path` and pass it to notation-go.

shizhMSFT: I think we can drop the `BlobSigner` interface and reuse the `Signer` interface as it is generic enough to cover all signing scenarios. Meanwhile, I think we can keep the `notation.SignBlob()` method and `notation.SignBlobOptions` type as utilities.

shizhMSFT: Since the [blob sign spec](https://github.com/notaryproject/notation/pull/811) is still under review, it would be better revisit this after the spec is merged.

priteshbandi: The problem with using [Signer Interface](https://github.com/notaryproject/notation-go/blob/main/notation.go#L62-L66) because the `Sign(ctx context.Context, desc ocispec.Descriptor, opts SignerSignOptions) ([]byte, *signature.SignerInfo, error)` method takes descriptor as argument but in case of blob we would need to create descriptor from `byte[]`. To create descriptor we need to know the digest algo and for that we need to call plugin for [`DescribeKey`](https://github.com/notaryproject/notation-go/blob/main/signer/plugin.go#L94-L97) and since we dont have access to plugin in `notation.go` we cannot use SignerInterface.

Two-Hearts: @priteshbandi Just sharing my thoughts here:
1. User specifies the <blob_path> + <hash_alg> to Notation CLI, where <hash_alg> is used to hash the blob content, supported values are `SHA256`, `SHA384` and `SHA512`.
2. Notation CLI constructs the blob descriptor based on user inputs. An example desc would be:
```
{
    "mediaType": "application/octet-stream",
    "digest": digest.FromBytes(hash(blob_content)),
    "size": 1024,
    "annotations": { // user defined metadata
        "hash-algorithm": SHA512
    }
}
```
Since the `hash-algorithm` is also in the payload, it's get signed as well and integrity is guaranteed.
4. Notation CLI uses the same `Signer interface` and call `notation-go SignBlob` method to sign the descriptor.
5. On verification, the <hash-algorithm> in the payload is used to hash the content then compute the digest again and compare with the value in the target descriptor.

(Similar design was seen in a Timestamp countersignature, RFC 3161.)

Two-Hearts: As we discussed offline, just re-posting my two cents here regarding the hash algorithm:
I think we don't have to use the `signing algorithm` to hash the blob content when computing the `digest` field of the target descriptor. This wasn't the case when we sign an `oci image` either. When we sign an `oci image`, we resolve the descriptor from the reference without knowing any information of the signing key. Then we send the descriptor along with other signed attributes to the `Signer`, who uses the `signing algorithm` to hash the payload again before signing.
Similarly, for `blob signing` we could use [digest.FromReader](https://pkg.go.dev/github.com/opencontainers/go-digest@v1.0.0#FromReader) with [Canonical](https://pkg.go.dev/github.com/opencontainers/go-digest@v1.0.0#Canonical) hash algorithm, essentially SHA256. This user experience aligns with what we have today. 

In terms of hash collision, SHA256 is a collision-resistant hash. Also, there is the `size` field, which is hashed by the `signing algorithm` as a part of the final payload before generating the signature. Therefore, it is even harder to find such a collision under SHA256 AND with the same size.

priteshbandi: I think we don't have to use the signing algorithm to hash the blob content when computing the digest field of the target descriptor. This wasn't the case when we sign an oci image either. When we sign an oci image, we resolve the descriptor from the reference without knowing any information of the signing key. Then we send the descriptor along with other signed attributes to the Signer, who uses the signing algorithm to hash the payload again before signing.

> In my opinion, we shouldn't draw a comparison between image signing and blob signing. In the case of OCI images, the descriptor itself is the content being signed (registry decides if they want to represent an artifact using sha256 or sha512 hash). However, with blobs, we first need to create the descriptor, which is then used for signing.  


Similarly, for blob signing we could use [digest.FromReader](https://pkg.go.dev/github.com/opencontainers/go-digest@v1.0.0#FromReader) with [Canonical](https://pkg.go.dev/github.com/opencontainers/go-digest@v1.0.0#Canonical) hash algorithm, essentially SHA256. This user experience aligns with what we have today. The design becomes pretty simple with much lower engineering cost. (I'm seeing duplicate codes in the implementation of the new BlobSigner interface as the current Signer interface is already very generic.)

> Yes but as per [signature-specification.md#algorithm-selection](https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#algorithm-selection), it is recommended to use a larger hashing algorithm for higher key sizes. In my opinion, we should adhere to this practice as it prevents confusion (as its similar to Image signing), and users with larger keys would naturally anticipate enhanced signature strength. Additionally, the process of hashing with one algorithm for descriptor generation and then employing another algorithm for signature generation seems counterintuitive.

Two-Hearts: @priteshbandi Just want to clarify, the your proposed signing process actually hashes 3 times, they are:
1. Using the signature algorithm associated hash func to compute the hash of the blob content. https://github.com/notaryproject/notation-go/blob/1263e20b887f51ca7e37eb2009fe54c798385c09/signer/plugin.go#L253
2. Using `digest.FromBytes(<hash computed from step 1>)`, this step uses the [Canonical hash](https://github.com/opencontainers/go-digest/blob/ea51bea511f75cfa3ef6098cc253c5c3609b037a/digest.go#L93) to compute the digest field of the target descriptor. https://github.com/notaryproject/notation-go/blob/1263e20b887f51ca7e37eb2009fe54c798385c09/signer/plugin.go#L267
3. The final step before signature generation, the `Signer` uses the signing algorithm to hash the payload again.

My suggestion is removing step 1 and change step 2 to just `digest.FromReader`.

priteshbandi: > @priteshbandi Just want to clarify, your proposed signing process actually hashes 3 times, they are:
Nop, I am advocating about using the same hashing algorithm throughout the signing operation. In this case it would be 1) during content digest calculation to generate descriptor 2)  during the actual signature generation process, the Signer uses the signing algorithm to hash the payload again.
> I dont think we need 1st step.

>This is quite common actually, see RFC 5652 CMS as an example: https://www.rfc-editor.org/rfc/rfc5652#section-5.3. Under Signed-data type, there's the digestAlgorithm and signatureAlgorithm defined separately.
Indeed, the specification permits all these permutations, but I don't see a compelling reason  to not maintain consistency in the signing algorithm throughout the entire signing operation. Having a consistent hashing algo will not only improve user experience as they wont need to worry about passing additional attribute (or introducing another defualt) but will also keep it consistent with container image signing spec. While I acknowledge that this may introduce some code complexity, in my opinion, it's a worthwhile tradeoff

Two-Hearts: > Nop, I am advocating about using the same hashing algorithm throughout the signing operation. In this case it would be 1) during content digest calculation to generate descriptor 2)  during the actual signature generation process, the Signer uses the signing algorithm to hash the payload again.
> I dont think we need 1st step.

But your current implementation in this PR does not align with what you are stating above:  
the current implementation in this PR does hash 3 times. `digest.FromBytes` in the `getDescriptor` function, which uses SHA256 as hash func.

priteshbandi: But your current implementation in this PR does not align with what you are stating above: the current implementation in this PR does hash 3 times. `digest.FromBytes` in the `getDescriptor` function, which uses SHA256 as hash func.

Ohh, that's a bug in code :D; Also, this PR was raised to build consensus on implementation approach and is not ready for detailed review.

Two-Hearts: Got it. Please fix the bug. Ideally, we should only need to hash 2 times instead of 3.

priteshbandi: Done.

shizhMSFT: > Since the [blob sign spec](https://github.com/notaryproject/specifications/pull/283) is still under review, it would be better revisit this after the spec is merged.

By following my above comment as well as @sajayantony 's comment in [the community meeting (2/5/2024)](https://hackmd.io/_vrqBGAOSUC_VWvFzWruZw?both#Feb-5-2024), let's follow up this conversion in the spec PR.

shizhMSFT (commented):
Blob signing introduced by https://github.com/notaryproject/specifications/pull/283 is slightly different from the OCI artifact signing. In summary, blob signing has the same signing / verification core as OCI artifact signing where the digest algorithm in the notary-project specific descriptor is restricted to the signing / verification key.

It is natural to share the same `Signer` interface with both OCI artifact and blob signing and add a new option `IsBlob` in `SignerSignOptions` (and `VerifierVerifyOptions` for `Verifier`) since the signing core is the notary-project specific descriptor. However, this solution is vulnerable for those existing implementations (if any) don't support blob signing / verification. Precisely, the `IsBlob` option may be dropped by those implementations unintentionally where a blob may be verified as an artifact although `IsBlob` is set to `true`, which produces unexpected results. Therefore, it is better to split it to two different interfaces as this PR does. In `notation-go/v2`, maybe we can combine the signing cores into a single one.

To elaborate a bit, we need the following new stuffs for signing
```go
type BlobSigner interface {
    SignBlob(ctx context.Context, desc ocispec.Descriptor, opts SignerSignOptions) ([]byte, *signature.SignerInfo, error)
    BlobDigestAlgorithm(ctx context.Context) (digest.Algorithm, error)
}

type SignBlobOptions struct {
    SignerSignOptions

    ContentMediaType string
    UserMetadata map[string]string
}

func SignBlob(ctx context.Context, signer BlobSigner, reader io.Reader, signOpts SignBlobOptions) ([]byte, error)
```
The `BlobSigner.SignBlob` does not take a `reader` since `notation-go`, as a library, it should provide flexibility where the digest of a huge blob is generated by other component efficiently. The `BlobSigner.SignBlob` implementation can later check if the digest algorithm matches the signing key or not and reject if not.

The `SignBlob` is more like a utility function of `BlobSigner.SignBlob` where it takes a reader to compute the digest using the algorithm specified by `BlobSigner.BlobDigestAlgorithm()` backed by the signing key.

/cc @Two-Hearts

priteshbandi:
> Blob signing introduced by [notaryproject/specifications#283](https://github.com/notaryproject/specifications/pull/283) is slightly different from the OCI artifact signing. In summary, blob signing has the same signing / verification core as OCI artifact signing where the digest algorithm in the notary-project specific descriptor is restricted to the signing / verification key.
> 
> It is natural to share the same `Signer` interface with both OCI artifact and blob signing and add a new option `IsBlob` in `SignerSignOptions` (and `VerifierVerifyOptions` for `Verifier`) since the signing core is the notary-project specific descriptor. However, this solution is vulnerable for those existing implementations (if any) don't support blob signing / verification. Precisely, the `IsBlob` option may be dropped by those implementations unintentionally where a blob may be verified as an artifact although `IsBlob` is set to `true`, which produces unexpected results. Therefore, it is better to split it to two different interfaces as this PR does. In `notation-go/v2`, maybe we can combine the signing cores into a single one.
> 
> To elaborate a bit, we need the following new stuffs for signing
> 
> ```go
> type BlobSigner interface {
>     SignBlob(ctx context.Context, desc ocispec.Descriptor, opts SignerSignOptions) ([]byte, *signature.SignerInfo, error)
>     BlobDigestAlgorithm(ctx context.Context) (digest.Algorithm, error)
> }
> 
> type SignBlobOptions struct {
>     SignerSignOptions
> 
>     ContentMediaType string
>     UserMetadata map[string]string
> }
> 
> func SignBlob(ctx context.Context, signer BlobSigner, reader io.Reader, signOpts SignBlobOptions) ([]byte, error)
> ```
> 
> The `BlobSigner.SignBlob` does not take a `reader` since `notation-go`, as a library, it should provide flexibility where the digest of a huge blob is generated by other component efficiently. The `BlobSigner.SignBlob` implementation can later check if the digest algorithm matches the signing key or not and reject if not.
> 
> The `SignBlob` is more like a utility function of `BlobSigner.SignBlob` where it takes a reader to compute the digest using the algorithm specified by `BlobSigner.BlobDigestAlgorithm()` backed by the signing key.
> 
> /cc @Two-Hearts

As discussed over slack I will be updating notation blob signing functions to following:
```
type SignBlobOptions struct {
	SignerSignOptions
	ContentMediaType string
	UserMetadata map[string]string
}

type BlobDescriptorGenerator func(digest.Algorithm) (ocispec.Descriptor, error)

type BlobSigner interface {
	SignBlob(ctx context.Context, genDesc BlobDescriptorGenerator, opts SignerSignOptions) ([]byte, *signature.SignerInfo, error)
}

Reasoning: This approach reduces the number of calls to the plugin's DescribeKey to just one, compared to the two calls required in the previous approach. Additionally, in earlier approach we did explore caching the DescribeKey response based on keyId, but it limits the seamless migration of signing keys behind the KeyId.

File: notation.go [hunk 6]
```diff
@@ -56,15 +61,36 @@ type SignerSignOptions struct {
...
+type SignBlobOptions struct {
+	SignerSignOptions
+
+	ContentMediaType string
+	// UserMetadata contains key-value pairs that are added to the signature
+	// payload
+	UserMetadata map[string]string
+}
+
+type BlobDescriptorGenerator func(digest.Algorithm) (ocispec.Descriptor, error)
```
shizhMSFT, line 83: Request for documentation

notation.go
```diff
@@ -152,6 +172,64 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, signOpts
...
+			Digest:    digest.NewDigest(hashAlgo, h),
+			Size:      bytes,
+		}
+		return addUserMetadataToDescriptor(ctx, targetDesc, signBlobOpts.UserMetadata)
+	}
+
+	return signer.SignBlob(ctx, getDescFun, signBlobOpts.SignerSignOptions)
+}
+
+func validate(signer any, signOpts SignerSignOptions) error {
```
shizhMSFT, line 212: Is it better to name it `validateSignArguments`? I guess you need a similar name for verify.

notation.go
```diff
@@ -152,6 +172,64 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, signOpts
...
+	if _, _, err := mime.ParseMediaType(signBlobOpts.ContentMediaType); err != nil {
+		return nil, nil, fmt.Errorf("invalid content media-type '%s': %v", signBlobOpts.ContentMediaType, err)
+	}
+
+	getDescFun := func(hashAlgo digest.Algorithm) (ocispec.Descriptor, error) {
+		h := hashAlgo.Hash()
+		bytes, err := io.Copy(hashAlgo.Hash(), blobReader)
+		if err != nil {
+			return ocispec.Descriptor{}, err
+		}
```
shizhMSFT, line 199: Consider [Algorithm.FromReader()](https://pkg.go.dev/github.com/opencontainers/go-digest#Algorithm.FromReader)

signer/plugin.go
```diff
@@ -29,88 +30,133 @@ import (
...
+	crypto.SHA256: digest.SHA256,
+	crypto.SHA384: digest.SHA384,
+	crypto.SHA512: digest.SHA512,
+}
+
 // NewFromPlugin creates a notation.Signer that signs artifacts and generates
 // signatures by delegating the one or more operations to the named plugin,
 // as defined in https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#signing-interfaces.
-func NewFromPlugin(plugin plugin.SignPlugin, keyID string, pluginConfig map[string]string) (notation.Signer, error) {
+func NewFromPlugin(plugin plugin.SignPlugin, keyID string, pluginConfig map[string]string) (*PluginSigner, error) {
```
shizhMSFT, line 55: Changing the function signature is a breaking change. Please consider a new constructor method.

signer/signer.go
```diff
@@ -36,24 +36,24 @@ import (
...
-// genericSigner implements notation.Signer and embeds signature.Signer
-type genericSigner struct {
+// GenericSigner implements notation.Signer and embeds signature.Signer
+type GenericSigner struct {
 	signature.Signer
 }
 
 // New returns a builtinSigner given key and cert chain
-func New(key crypto.PrivateKey, certChain []*x509.Certificate) (notation.Signer, error) {
+func New(key crypto.PrivateKey, certChain []*x509.Certificate) (*GenericSigner, error) {
```
shizhMSFT, line 45: Changing the function signature is a breaking change. Please consider a new constructor method.

signer/plugin.go
```diff
@@ -29,88 +30,133 @@ import (
...
 		return nil, nil, err
 	}
 
-	logger.Debugf("Using plugin %v with capabilities %v to sign artifact %v in signature media type %v", metadata.Name, metadata.Capabilities, desc.Digest, opts.SignatureMediaType)
-	if metadata.HasCapability(plugin.CapabilitySignatureGenerator) {
-		return s.generateSignature(ctx, desc, opts, metadata)
-	} else if metadata.HasCapability(plugin.CapabilityEnvelopeGenerator) {
+	logger.Debugf("Using plugin %v with capabilities %v to sign oci artifact %v in signature media type %v", metadata.Name, metadata.Capabilities, desc.Digest, opts.SignatureMediaType)
+	if metadata.HasCapability(proto.CapabilitySignatureGenerator) {
+		logger.Debug("Invoking plugin's describe-key command")
```
shizhMSFT, line 89: Should this debug log inside of the `getKeySpec` method?

signer/plugin.go
```diff
@@ -29,88 +30,133 @@ import (
...
+func (s *PluginSigner) generateSignature(ctx context.Context, desc ocispec.Descriptor, opts notation.SignerSignOptions, ks signature.KeySpec, metadata *plugin.GetMetadataResponse) ([]byte, *signature.SignerInfo, error) {
+	logger := log.GetLogger(ctx)
+	logger.Debug("Generating signature by plugin")
+	genericSigner := GenericSigner{
 		Signer: &pluginPrimitiveSigner{
 			ctx:          ctx,
 			plugin:       s.plugin,
 			keyID:        s.keyID,
-			pluginConfig: config,
+			pluginConfig: s.mergeConfig(opts.PluginConfig),
```
shizhMSFT, line 159: The config has been merged before calling `generateSignature`. Is it possible to reuse `mergedConfig`?

signer/plugin.go
```diff
@@ -163,25 +209,25 @@ func (s *pluginSigner) generateSignatureEnvelope(ctx context.Context, desc ocisp
 		return nil, nil, err
 	}
 
-	content := envContent.Payload.Content
+	cnt := envContent.Payload.Content
```
shizhMSFT, line 212: Is there any reason to use this non-descriptive variable name `cnt`?

signer/signer.go
```diff
@@ -146,3 +146,26 @@ func (s *genericSigner) Sign(ctx context.Context, desc ocispec.Descriptor, opts
...
+
+	ks, err := s.KeySpec()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	digestAlg, ok := algorithms[ks.SignatureAlgorithm().Hash()]
+	if !ok {
+		return nil, nil, fmt.Errorf("unknown hashing algo %v", ks.SignatureAlgorithm().Hash())
+	}
```
shizhMSFT, line 163: It seems we can refactor this part as it has been used by both `plugin.go` and `signer.go`.

notation.go
priteshbandi: We cannot use `[Algorithm.FromReader()](https://pkg.go.dev/github.com/opencontainers/go-digest#Algorithm.FromReader)` because we also need content length.

signer/signer.go
priteshbandi: we are moving from interface to a particular implementation. Shouldn't this be backward compatible ?

signer/plugin.go
priteshbandi: tracking it in https://github.com/notaryproject/notation-go/pull/379#discussion_r1512772475

signer/signer.go
shizhMSFT: But the obtained `KeySpec` instances are all of `signature.KeySpec` type. We can refactor from this point.

signer/plugin.go
```diff
@@ -29,88 +30,136 @@ import (
...
+		return s.generateSignature(ctx, desc, opts, ks, metadata, mergedConfig)
+	} else if metadata.HasCapability(proto.CapabilityEnvelopeGenerator) {
 		return s.generateSignatureEnvelope(ctx, desc, opts)
 	}
 	return nil, nil, fmt.Errorf("plugin does not have signing capabilities")
 }
 
-func (s *pluginSigner) generateSignature(ctx context.Context, desc ocispec.Descriptor, opts notation.SignerSignOptions, metadata *plugin.GetMetadataResponse) ([]byte, *signature.SignerInfo, error) {
+// SignBlob signs the arbitrary data and returns the marshalled envelope.
+func (s *PluginSigner) SignBlob(ctx context.Context, blobGenFunc notation.BlobDescriptorGenerator, opts notation.SignerSignOptions) ([]byte, *signature.SignerInfo, error) {
```
rgnote, line 108:
```suggestion
func (s *PluginSigner) SignBlob(ctx context.Context, descGenFunc notation.BlobDescriptorGenerator, opts notation.SignerSignOptions) ([]byte, *signature.SignerInfo, error) {
```

signer/signer.go
```diff
@@ -36,24 +36,34 @@ import (
...
 	signature.Signer
 }
 
 // New returns a builtinSigner given key and cert chain
 func New(key crypto.PrivateKey, certChain []*x509.Certificate) (notation.Signer, error) {
+	return NewGenericSigner(key, certChain)
+}
+
+// NewGenericSigner returns a builtinSigner given key and cert chain
+func NewGenericSigner(key crypto.PrivateKey, certChain []*x509.Certificate) (*GenericSigner, error) {
```
rgnote, line 50: Do we need two interfaces for creating GenericSigner from key/cert? Wouldn't that create confusion?

signer/signer.go
```diff
@@ -36,24 +36,34 @@ import (
 // signingAgent is the unprotected header field used by signature.
 const signingAgent = "Notation/1.0.0"
 
-// genericSigner implements notation.Signer and embeds signature.Signer
-type genericSigner struct {
+// GenericSigner implements notation.Signer and embeds signature.Signer
+type GenericSigner struct {
```
rgnote, line 40: What is the reason for making this public?

signer/signer.go
```diff
@@ -146,3 +156,30 @@ func (s *genericSigner) Sign(ctx context.Context, desc ocispec.Descriptor, opts
 	// TODO: re-enable timestamping https://github.com/notaryproject/notation-go/issues/78
 	return sig, &envContent.SignerInfo, nil
 }
+
+// SignBlob signs the descriptor returned by blobGen and returns the marshalled envelope
+func (s *GenericSigner) SignBlob(ctx context.Context, blobGenFunc notation.BlobDescriptorGenerator, opts notation.SignerSignOptions) ([]byte, *signature.SignerInfo, error) {
```
rgnote, line 161:
```suggestion
func (s *GenericSigner) SignBlob(ctx context.Context, descGenFunc notation.BlobDescriptorGenerator, opts notation.SignerSignOptions) ([]byte, *signature.SignerInfo, error) {
```

signer/signer.go
```diff
@@ -146,3 +156,30 @@ func (s *genericSigner) Sign(ctx context.Context, desc ocispec.Descriptor, opts
...
+
+	desc, err := getDescriptor(ks, blobGenFunc)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return s.Sign(ctx, desc, opts)
+}
+
+func getDescriptor(ks signature.KeySpec, blobGenFunc notation.BlobDescriptorGenerator) (ocispec.Descriptor, error) {
```
rgnote, line 178:
```suggestion
func getDescriptor(ks signature.KeySpec, descGenFunc notation.BlobDescriptorGenerator) (ocispec.Descriptor, error) {
```

signer/signer.go
priteshbandi: GenericSigner satisfies both blob signer and container signer interface so instead of returning notatin.Signer now we need to return `GenericSigner`

signer/signer.go
priteshbandi: Yes this is done for backward comparability, we need to return `GenericSigner` and we cannot update existing function.

rgnote (approved):
LGTM

signer/signer.go
rgnote: Should we mark `New` as deprecated so that we encourage users to use NewGenericSigner method?

priteshbandi: Yep, I can do that in new PR

signer/plugin.go
```diff
@@ -29,88 +30,136 @@ import (
...
+var algorithms = map[crypto.Hash]digest.Algorithm{
+	crypto.SHA256: digest.SHA256,
+	crypto.SHA384: digest.SHA384,
+	crypto.SHA512: digest.SHA512,
+}
+
 // NewFromPlugin creates a notation.Signer that signs artifacts and generates
 // signatures by delegating the one or more operations to the named plugin,
 // as defined in https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#signing-interfaces.
 func NewFromPlugin(plugin plugin.SignPlugin, keyID string, pluginConfig map[string]string) (notation.Signer, error) {
```
shizhMSFT, line 55: Do you want to mark this method as deprecated?

signer/signer.go
```diff
@@ -36,24 +36,34 @@ import (
 // signingAgent is the unprotected header field used by signature.
 const signingAgent = "Notation/1.0.0"
 
-// genericSigner implements notation.Signer and embeds signature.Signer
-type genericSigner struct {
+// GenericSigner implements notation.Signer and embeds signature.Signer
+type GenericSigner struct {
 	signature.Signer
 }
```
shizhMSFT, line 42: Since `GenericSigner` is exposed, do you want to hide `signature.Signer` from public use?

shizhMSFT (approved):
LGTM

rgnote (approved):
LGTM

PR #394, opened 2024-04-02
feat: Implements VerifyBlob functionality

priteshbandi:
This PR implements VerifyBlob functionality as per https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md

*This PRis ready for review*


dir/path.go
```diff
@@ -57,8 +57,14 @@ const (
...
-	// PathTrustPolicy is the trust policy file relative path.
-	PathTrustPolicy = "trustpolicy.json"
+	// PathTrustPolicy is the OCI trust policy file relative path.
+	// Deprecated: PathTrustPolicy exists for historical compatibility and should not be used.
+	// To get OCI trust policy path, use PathOCITrustPolicy.
+	PathTrustPolicy = PathOCITrustPolicy
+	// PathOCITrustPolicy is the OCI trust policy file relative path.
+	PathOCITrustPolicy = "trustpolicy.json"
+	// PathBlobTrustPolicy is the Blob trust policy file relative path.
+	PathBlobTrustPolicy = "trustpolicy.json"
```
Two-Hearts, line 67: The name for blob trust policy should be `trustpolicy.blob.json`.

dir/path.go
```diff
@@ -57,8 +57,14 @@ const (
...
 	// PathSigningKeys is the signingkeys file relative path.
 	PathSigningKeys = "signingkeys.json"
-	// PathTrustPolicy is the trust policy file relative path.
-	PathTrustPolicy = "trustpolicy.json"
+	// PathTrustPolicy is the OCI trust policy file relative path.
+	// Deprecated: PathTrustPolicy exists for historical compatibility and should not be used.
+	// To get OCI trust policy path, use PathOCITrustPolicy.
+	PathTrustPolicy = PathOCITrustPolicy
+	// PathOCITrustPolicy is the OCI trust policy file relative path.
+	PathOCITrustPolicy = "trustpolicy.json"
```
Two-Hearts, line 65: According to our spec: https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#trust-policy,  `PathOCITrustPolicy` should have two valid values, they are `trustpolicy.json` and `trustpolicy.oci.json`.

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,135 @@
...
+	err := getDocument(dir.PathBlobTrustPolicy, &doc)
+	return &doc, err
+}
+
+// Validate validates a policy document according to its version's rule set.
+// if any rule is violated, returns an error
+func (policyDoc *BlobDocument) Validate() error {
+	// sanity check
+	if policyDoc == nil {
+		return errors.New("trust policy document cannot be nil")
```
Two-Hearts, line 62: This applies to other error messages in this PR, can we add the word `blob` in the error message? This makes the error message very clear to the user that the blob trust policy is malformed. Something like:
```suggestion
		return errors.New("blob trust policy document cannot be nil")
```

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	// Validate Version
+	if policyDoc.Version == "" {
+		return errors.New("trust policy has empty version, version must be specified")
```
Two-Hearts, line 67:
```suggestion
		return errors.New("blob trust policy has empty version, version must be specified")
```

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	if !slices.Contains(supportedPolicyVersions, policyDoc.Version) {
+		return fmt.Errorf("trust policy document uses unsupported version %q", policyDoc.Version)
```
Two-Hearts, line 70:
```suggestion
		return fmt.Errorf("blob trust policy document uses unsupported version %q", policyDoc.Version)
```

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	// Validate the policy according to 1.0 rules
+	if len(policyDoc.BlobTrustPolicies) == 0 {
+		return errors.New("trust policy document can not have zero trust policy statements")
```
Two-Hearts, line 75:
```suggestion
		return errors.New("blob trust policy document can not have zero trust policy statements")
```

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	policyStatementNameCount := make(map[string]int)
+	foundGlobalPolicy := false
+	for _, statement := range policyDoc.BlobTrustPolicies {
+		policyStatementNameCount[statement.Name]++
+		if err := validatePolicyCore(statement.Name, statement.SignatureVerification, statement.TrustStores, statement.TrustedIdentities); err != nil {
+			return err
```
Two-Hearts, line 83:
```suggestion
			return fmt.Errorf("blob trust policy: %w", err)
```

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+
+		if statement.GlobalPolicy {
+			if foundGlobalPolicy {
+				return errors.New("multiple trust policy statements have globalPolicy set to true. Only statement should be marked as global policy")
```
Two-Hearts, line 88:
```suggestion
				return errors.New("multiple blob trust policy statements have globalPolicy set to true. Only one statement can be marked as global policy")
```

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,135 @@
...
+	if !slices.Contains(supportedPolicyVersions, policyDoc.Version) {
+		return fmt.Errorf("trust policy document uses unsupported version %q", policyDoc.Version)
+	}
+
+	// Validate the policy according to 1.0 rules
+	if len(policyDoc.BlobTrustPolicies) == 0 {
+		return errors.New("trust policy document can not have zero trust policy statements")
+	}
+
+	policyStatementNameCount := make(map[string]int)
```
Two-Hearts, line 78: Consider using a `set` to check duplication. notation-go has its built in `set` here: https://github.com/notaryproject/notation-go/blob/main/internal/container/set.go

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,135 @@
...
+			foundGlobalPolicy = true
+		}
+	}
+
+	// Verify unique policy statement names across the policy document
+	for key := range policyStatementNameCount {
+		if policyStatementNameCount[key] > 1 {
+			return fmt.Errorf("multiple trust policy statements use the same name %q, statement names must be unique", key)
+		}
+	}
```
Two-Hearts, line 99: If using the `set` as mentioned above, we won't need this check.

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return nil
+}
+
+// GetApplicableTrustPolicy returns a pointer to the deep copied TrustPolicy
+// statement that applies to the given registry scope. If no applicable trust
+// policy is found, returns an error
+// see https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#selecting-a-trust-policy-based-on-artifact-uri
```
Two-Hearts, line 107: The function doc needs to be updated accordingly.

verifier/trustpolicy/trustpolicy.go
```diff
@@ -136,185 +134,46 @@ var (
 
 var supportedPolicyVersions = []string{"1.0"}
```
Two-Hearts, line 135: nit:
`supportedPolicyVersions` might not be a shared variable between `blob` and `oci` trust policy documents in the future. Their versions may develop separately.

verifier/trustpolicy/blob.go
Two-Hearts, line 1: Please add unit tests to this file.

dir/path.go
priteshbandi: Yep, added a fallback logic

verifier/trustpolicy/trustpolicy.go
priteshbandi: Yep I agree and that's the reason its private variable. When it diverges we will create new ones variables and validations.

verifier/trustpolicy/blob.go
priteshbandi: good point, done

verifier/verifier.go
```diff
@@ -46,99 +46,216 @@ import (
...
+
+	payload := &envelope.Payload{}
+	err = json.Unmarshal(outcome.EnvelopeContent.Payload.Content, payload)
+	if err != nil {
+		logger.Error("Failed to unmarshal the payload content in the signature blob to envelope.Payload")
+		outcome.Error = err
+		return outcome, err
+	}
+
+	desc, err := descGenFunc(payload.TargetArtifact.Digest.Algorithm())
```
Two-Hearts, line 220: hmmm, eventually, we are going to compare `desc` with `payload.TargetArtifact`, I think we shouldn't use `payload.TargetArtifact.Digest.Algorithm` to compute the digest of `desc`? According to our spec https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#steps, `step 3.vi.b`:
```
Additionally for Blob artifacts, calculate the digest of the blob using the hashing algorithm deduced from signing certificate's public key (see [Algorithm Selection](https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#algorithm-selection)) and make sure the digests match.
```

verifier/verifier.go
```diff
@@ -46,99 +46,216 @@ import (
...
 )
 
-// verifier implements notation.Verifier and notation.verifySkipper
-type verifier struct {
-	trustPolicyDoc   *trustpolicy.Document
-	trustStore       truststore.X509TrustStore
-	pluginManager    plugin.Manager
-	revocationClient revocation.Revocation
+// Verifier implements notation.Verifier, notation.BlobVerifier and notation.verifySkipper
+type Verifier struct {
```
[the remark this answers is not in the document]
priteshbandi: To avoid creating duplicate verifiers for OCI and blob verification purposes, we can leverage the fact that the underlying struct already implements both notation.BlobVerifier and notation.Verifier interfaces.

verifier/verifier.go
```diff
@@ -46,99 +46,216 @@ import (
...
+}
+
+// NewVerifier creates a new Verifier given ociTrustPolicy, trustStore and pluginManager
+func NewVerifier(ociTrustPolicy *trustpolicy.OCIDocument, blobTrustPolicy *trustpolicy.BlobDocument, trustStore truststore.X509TrustStore, pluginManager plugin.Manager) (*Verifier, error) {
+	return NewVerifierWithOptions(ociTrustPolicy, blobTrustPolicy, trustStore, pluginManager, VerifierOptions{})
+}
+
+// NewVerifierWithOptions creates a new Verifier given ociTrustPolicy, blobTrustPolicy,
+// trustStore, pluginManager, and verifierOptions
+func NewVerifierWithOptions(ociTrustPolicy *trustpolicy.OCIDocument, blobTrustPolicy *trustpolicy.BlobDocument, trustStore truststore.X509TrustStore, pluginManager plugin.Manager, verifierOptions VerifierOptions) (*Verifier, error) {
```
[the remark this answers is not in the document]
priteshbandi: We need this method to be public to allow users to create Verifier from code instead of files.

notation.go
```diff
@@ -352,6 +377,48 @@ type VerifyOptions struct {
...
+func VerifyBlob(ctx context.Context, blobVerifier BlobVerifier, blobReader io.Reader, signature []byte, verifyBlobOpts VerifyBlobOptions) (ocispec.Descriptor, *VerificationOutcome, error) {
+	if blobVerifier == nil {
+		return ocispec.Descriptor{}, nil, errors.New("blobVerifier cannot be nil")
+	}
+
+	if blobReader == nil {
+		return ocispec.Descriptor{}, nil, errors.New("blobReader cannot be nil")
+	}
+
+	if signature == nil || len(signature) == 0 {
```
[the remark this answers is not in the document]
priteshbandi: nil and []byte{} are different

verifier/verifier.go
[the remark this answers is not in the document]
priteshbandi: Not really, we aren't exporting any field inside the struct so user cannot modify any field.

verifier/verifier.go
priteshbandi: Good catch, fixed

verifier/verifier.go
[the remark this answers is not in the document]
Two-Hearts: I see. Then keeping it as `unexported` should be fine? Is there any blocker that requires the struct must be exported? 

I'm thinking if we should provide clear constructors for OCI and blob verifiers respectively. By 'clear' I mean unmixed constructors, user should know exactly what they want and create. The `NewVerifier` constructor in this PR still looks a bit error prone to me (it's possible that user chooses to only pass in an OCI trust policy, but later decides to use it as a blob verifier interface, then their verification would have an issue). Maybe others have different opinions? /cc: @shizhMSFT @JeyJeyGao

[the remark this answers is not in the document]
priteshbandi: A public method that returns unexported/private struct seems odd.

We are providing both kind of constroctors i.e. oci only, blob only and oci&blob only, each has there own usecase. If user want oci only then they should use oci only constroctor.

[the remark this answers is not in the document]
priteshbandi: tracking it in https://github.com/notaryproject/notation-go/pull/394#discussion_r1558875797

verifier/verifier.go
```diff
@@ -46,99 +46,216 @@ import (
...
+// New creates a new Verifier given ociTrustPolicy, trustStore and pluginManager
+// Deprecated: New function exists for historical compatibility and should not be used.
+// To create Verifier, use NewVerifier function.
+func New(ociTrustPolicy *trustpolicy.OCIDocument, trustStore truststore.X509TrustStore, pluginManager plugin.Manager) (notation.Verifier, error) {
+	return NewVerifier(ociTrustPolicy, nil, trustStore, pluginManager)
 }
 
 // SkipVerify validates whether the verification level is skip.
-func (v *verifier) SkipVerify(ctx context.Context, opts notation.VerifierVerifyOptions) (bool, *trustpolicy.VerificationLevel, error) {
+func (v *Verifier) SkipVerify(ctx context.Context, opts notation.VerifierVerifyOptions) (bool, *trustpolicy.VerificationLevel, error) {
```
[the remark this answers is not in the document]
priteshbandi: https://github.com/notaryproject/notation-go/pull/394#discussion_r1558875797

verifier/verifier.go
```diff
@@ -46,99 +46,216 @@ import (
...
 type VerifierOptions struct {
 	// RevocationClient is an implementation of revocation.Revocation to use for
 	// verifying revocation
 	RevocationClient revocation.Revocation
 }
 
-// NewFromConfig returns a verifier based on local file system
-func NewFromConfig() (notation.Verifier, error) {
+// NewOCIVerifierFromConfig returns a OCI Verifier based on local file system
+func NewOCIVerifierFromConfig() (*Verifier, error) {
```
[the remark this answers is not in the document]
priteshbandi: tracking it on https://github.com/notaryproject/notation-go/pull/394#discussion_r1558875797

verifier/verifier.go
```diff
@@ -46,99 +46,216 @@ import (
...
 
-	return New(policyDocument, x509TrustStore, plugin.NewCLIManager(dir.PluginFS()))
+	return NewVerifier(policyDocument, nil, x509TrustStore, plugin.NewCLIManager(dir.PluginFS()))
 }
 
-// New creates a new verifier given trustPolicy, trustStore and pluginManager
-func New(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore, pluginManager plugin.Manager) (notation.Verifier, error) {
-	return NewWithOptions(trustPolicy, trustStore, pluginManager, VerifierOptions{})
+// NewBlobVerifierFromConfig returns a Blob Verifier based on local file system
+func NewBlobVerifierFromConfig() (*Verifier, error) {
```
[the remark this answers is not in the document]
priteshbandi: Tracking it in https://github.com/notaryproject/notation-go/pull/394#discussion_r1558875797

verifier/verifier.go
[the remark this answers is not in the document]
Two-Hearts: > A public method that returns unexported/private struct seems odd.

We are not returning the unexported/private struct `verifier` here.
For OCI, the constructor would be 
```
func NewOciVerifierWithOptions(ociTrustPolicy *trustpolicy.OCIDocument, trustStore truststore.X509TrustStore, pluginManager plugin.Manager, opts VerifierOptions) (notation.Verifier, error)   // user cannot pass a blob trust policy to an OCI verifier.

func NewOciVerifierFromConfig() (notation.Verifier, error)
```

For Blob, the constructor would be 
```
func NewBlobVerifierWithOptions(blobTrustPolicy *trustpolicy.BlobDocument, trustStore truststore.X509TrustStore, pluginManager plugin.Manager, opts VerifierOptions) (notation.BlobVerifier , error)    // user cannot pass an oci trust policy to a blob verifier.

func NewBlobVerifierFromConfig() (notation.BlobVerifier , error)
```

We shouldn't provide a constructor of `oci + blob` by returning the concrete struct. It's error prone, because semantically, due to Golang duck typing, it's possible that the caller only passed in an OCI trust policy but use it as a `notatoin.BlobVerifier ` later, this is a semantic error and Notation is not checking it. This kind of error is implicit and hard to figure out for the caller during runtime.

verifier/verifier.go
```diff
@@ -43,102 +44,235 @@ import (
...
+
 	return false, verificationLevel, nil
 }
 
-// Verify verifies the signature blob `signature` against the target OCI
+// VerifyBlob verifies the signature of given blob , and returns the outcome upon
+// successful verification.
+func (v *Verifier) VerifyBlob(ctx context.Context, descGenFunc notation.BlobDescriptorGenerator, signature []byte, opts notation.BlobVerifierVerifyOptions) (*notation.VerificationOutcome, error) {
+	logger := log.GetLogger(ctx)
+	logger.Debugf("Verify signature %v of media type %v", opts.SignatureMediaType)
```
rgnote, line 195: First `%v` is missing

verifier/verifier.go
```diff
@@ -635,7 +770,7 @@ func verifyRevocation(outcome *notation.VerificationOutcome, r revocation.Revoca
 	return result
 }
 
-func executePlugin(ctx context.Context, installedPlugin pluginframework.VerifyPlugin, trustPolicy *trustpolicy.TrustPolicy, capabilitiesToVerify []pluginframework.Capability, envelopeContent *signature.EnvelopeContent, pluginConfig map[string]string) (*pluginframework.VerifySignatureResponse, error) {
+func executePlugin(ctx context.Context, installedPlugin pluginframework.VerifyPlugin, capabilitiesToVerify []pluginframework.Capability, envelopeContent *signature.EnvelopeContent, identities []string, pluginConfig map[string]string) (*pluginframework.VerifySignatureResponse, error) {
```
rgnote, line 773:
```suggestion
func executePlugin(ctx context.Context, installedPlugin pluginframework.VerifyPlugin, capabilitiesToVerify []pluginframework.Capability, envelopeContent *signature.EnvelopeContent, trustedIdentities []string, pluginConfig map[string]string) (*pluginframework.VerifySignatureResponse, error) {
```

verifier/verifier.go
```diff
@@ -676,36 +811,36 @@ func executePlugin(ctx context.Context, installedPlugin pluginframework.VerifyPl
...
+		Signature:       sig,
 		TrustPolicy:     policy,
 		PluginConfig:    pluginConfig,
 	}
 	return installedPlugin.VerifySignature(ctx, req)
 }
 
-func verifyX509TrustedIdentities(certs []*x509.Certificate, trustPolicy *trustpolicy.TrustPolicy) error {
-	if slices.Contains(trustPolicy.TrustedIdentities, trustpolicyInternal.Wildcard) {
+func verifyX509TrustedIdentities(name string, identities []string, certs []*x509.Certificate) error {
```
rgnote, line 827:
```suggestion
func verifyX509TrustedIdentities(policyName string, trsutedIdentities []string, certs []*x509.Certificate) error {
```

verifier/verifier.go
```diff
@@ -155,7 +289,7 @@ func (v *verifier) Verify(ctx context.Context, desc ocispec.Descriptor, signatur
 		logger.Debug("Skipping signature verification")
 		return outcome, nil
 	}
-	err = v.processSignature(ctx, signature, envelopeMediaType, trustPolicy, pluginConfig, outcome)
+	err = v.processSignature(ctx, signature, envelopeMediaType, trustPolicy.Name, trustPolicy.TrustedIdentities, trustPolicy.TrustStores, pluginConfig, outcome)
```
rgnote, line 292: Why not simply pass the `trustPolicy` struct to keep the argument count lower?

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/notaryproject/notation-go/internal/slices"
+)
+
+// BlobDocument represents a trustpolicy.blob.json document
+type BlobDocument struct {
+	// Version of the policy document
+	Version string `json:"version"`
+
+	// BlobTrustPolicies include each policy statement
+	BlobTrustPolicies []BlobTrustPolicy `json:"trustPolicies"`
```
rgnote, line 31: Let's remove the Blob prefix. `BlobDocument.BlobTrustPolicies` is stuttering.

verifier/trustpolicy/trustpolicy.go
```diff
@@ -136,185 +134,16 @@ var (
...
```
rgnote, line 135: Let's split this into two separate variables and move them to blob.go, oci.go files, so that we don't accidentally introduce regression w.r.t. versioning.

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	policyNames := set.New[string]()
+	foundGlobalPolicy := false
+	for _, statement := range policyDoc.BlobTrustPolicies {
+		// Verify unique policy statement names across the policy document
+		if policyNames.Contains(statement.Name) {
+			return fmt.Errorf("multiple blob trust policy statements use the same name %q, statement names must be unique", statement.Name)
+		}
+		policyNames.Add(statement.Name)
```
rgnote, line 87: You should validate the `statement.Name` exists before making use of it for uniqueness check.

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	return nil
+}
+
+// GetApplicableTrustPolicy returns a pointer to the deep copied TrustPolicy
+// see https://github.com/notaryproject/notaryproject/blob/v1.1.0/specs/trust-store-trust-policy.md#blob-trust-policy
+func (policyDoc *BlobDocument) GetApplicableTrustPolicy(policyName string) (*BlobTrustPolicy, error) {
+	for _, policyStatement := range policyDoc.BlobTrustPolicies {
+		if policyName == "" {
```
rgnote, line 108: How can we make sure it is the user's intent to apply the global policy and not unintentionally picking it because they made an error in passing the policyName? At notation CLI layer, picking the global policy if PolicyName is empty makes sense. But at notation-go layer, should we make sure the customer is intentional in picking global policy? I can see a user shooting themselves in the foot by picking the global policy because their code has a bug where it is not properly initializing the policyName variable.

verifier/trustpolicy/oci.go
```diff
@@ -0,0 +1,253 @@
...
+
+	policyNames := set.New[string]()
+	for _, statement := range policyDoc.TrustPolicies {
+		// Verify unique policy statement names across the policy document
+		if policyNames.Contains(statement.Name) {
+			return fmt.Errorf("multiple oci trust policy statements use the same name %q, statement names must be unique", statement.Name)
+		}
+		policyNames.Add(statement.Name)
+
+		if err := validatePolicyCore(statement.Name, statement.SignatureVerification, statement.TrustStores, statement.TrustedIdentities); err != nil {
```
rgnote, line 120: Let's do the core validation first before validating unique policy names.

verifier/trustpolicy/trustpolicy.go
```diff
@@ -389,30 +218,91 @@ func (signatureVerification *SignatureVerification) GetVerificationLevel() (*Ver
...
+	}
+	return nil
 }
 
 // validateTrustStore validates if the policy statement is following the
-// Notary Project spec rules for truststores
-func validateTrustStore(statement TrustPolicy) error {
-	for _, trustStore := range statement.TrustStores {
+// Notary Project spec rules for truststore
+func validateTrustStore(name string, trustStores []string) error {
```
rgnote, line 295:
```suggestion
func validateTrustStore(policyName string, trustStores []string) error {
```

verifier/trustpolicy/trustpolicy.go
```diff
@@ -421,83 +311,50 @@ func validateTrustStore(statement TrustPolicy) error {
 
 // validateTrustedIdentities validates if the policy statement is following the
 // Notary Project spec rules for trusted identities
-func validateTrustedIdentities(statement TrustPolicy) error {
-	// If there is a wildcard in trusted identies, there shouldn't be any other
+func validateTrustedIdentities(name string, tis []string) error {
```
rgnote, line 314: ### ```suggestion
func validateTrustedIdentities(policyName string, trustedIdentities []string) error {
```

File: verifier/verifier.go (hunk 14 above)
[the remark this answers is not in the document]
Comment (JeyJeyGao) [reply verifier/verifier.go:58]: I think a public method that returns an unexported struct as an interface is a common practice to hide the internal data structure.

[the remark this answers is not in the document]
Comment (JeyJeyGao) [reply verifier/verifier.go:58]: It is better to keep it as a private struct. We should export them as little as possible.

File: verifier/verifier.go [hunk 31]
```diff
@@ -43,102 +44,235 @@ import (
...
 	}
 	// load trust store
 	x509TrustStore := truststore.NewX509TrustStore(dir.ConfigFS())
 
-	return New(policyDocument, x509TrustStore, plugin.NewCLIManager(dir.PluginFS()))
+	return NewVerifier(policyDocument, nil, x509TrustStore, plugin.NewCLIManager(dir.PluginFS()))
 }
 
-// New creates a new verifier given trustPolicy, trustStore and pluginManager
-func New(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore, pluginManager plugin.Manager) (notation.Verifier, error) {
```
JeyJeyGao, line 79: Should we keep the New() function to avoid breaking change?

verifier/verifier.go
```diff
@@ -43,102 +44,235 @@ import (
...
+	// load trust store
+	x509TrustStore := truststore.NewX509TrustStore(dir.ConfigFS())
+
+	return NewVerifier(nil, policyDocument, x509TrustStore, plugin.NewCLIManager(dir.PluginFS()))
 }
 
-// NewWithOptions creates a new verifier given trustPolicy, trustStore,
+// NewWithOptions creates a new Verifier given ociTrustPolicy, trustStore,
 // pluginManager, and VerifierOptions
-func NewWithOptions(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore, pluginManager plugin.Manager, opts VerifierOptions) (notation.Verifier, error) {
```
JeyJeyGao, line 85: We are introducing breaking change.

verifier/verifier.go
```diff
@@ -43,102 +44,235 @@ import (
...
+}
+
+// NewVerifier creates a new Verifier given ociTrustPolicy, trustStore and pluginManager
+func NewVerifier(ociTrustPolicy *trustpolicy.OCIDocument, blobTrustPolicy *trustpolicy.BlobDocument, trustStore truststore.X509TrustStore, pluginManager plugin.Manager) (*Verifier, error) {
+	return NewVerifierWithOptions(ociTrustPolicy, blobTrustPolicy, trustStore, pluginManager, VerifierOptions{})
+}
+
+// NewVerifierWithOptions creates a new Verifier given ociTrustPolicy, blobTrustPolicy,
+// trustStore, pluginManager, and verifierOptions
+func NewVerifierWithOptions(ociTrustPolicy *trustpolicy.OCIDocument, blobTrustPolicy *trustpolicy.BlobDocument, trustStore truststore.X509TrustStore, pluginManager plugin.Manager, verifierOptions VerifierOptions) (*Verifier, error) {
```
JeyJeyGao, line 115: Should we return an interface rather than the struct?

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,133 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/notaryproject/notation-go/dir"
+	set "github.com/notaryproject/notation-go/internal/container"
+	"github.com/notaryproject/notation-go/internal/slices"
+)
+
+// BlobDocument represents a trustpolicy.blob.json document
+type BlobDocument struct {
```
JeyJeyGao, line 26: Do we need to consider how to support a multi-version trust policy? If we have trust policy v2.0, and the struct fields differ from v1.0 because our verifier uses the v1.0 struct directly, how can we ensure that both v1.0 and v2.0 work without causing breaking changes?

notation.go
```diff
@@ -530,3 +603,19 @@ func getDescriptorFunc(ctx context.Context, reader io.Reader, contentMediaType s
 		return addUserMetadataToDescriptor(ctx, targetDesc, userMetadata)
 	}
 }
+
+func validateContentType(contentMediaType string) error {
+	if contentMediaType != "" {
```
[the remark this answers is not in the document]
priteshbandi: It should be part of `notation blob sign`  implementation. For notation-go mediatype is a required parameter.

notation.go
```diff
@@ -328,6 +327,32 @@ type Verifier interface {
 	Verify(ctx context.Context, desc ocispec.Descriptor, signature []byte, opts VerifierVerifyOptions) (*VerificationOutcome, error)
 }
 
+// BlobVerifierVerifyOptions contains parameters for Verifier.Verify.
```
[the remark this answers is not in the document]
priteshbandi: Good catch, fixed

notation.go
```diff
@@ -328,6 +327,32 @@ type Verifier interface {
 	Verify(ctx context.Context, desc ocispec.Descriptor, signature []byte, opts VerifierVerifyOptions) (*VerificationOutcome, error)
```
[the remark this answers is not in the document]
priteshbandi: For this pull request, we'll keep the old approach as it's already a significant change. We can discuss deprecating or renaming it in a future pull request. 
https://github.com/notaryproject/notation-go/issues/410

verifier/trustpolicy/blob.go
priteshbandi: good catch, moved `policyNames.Add` to end of for loop.

verifier/verifier.go
priteshbandi: We do have new function `func New(ociTrustPolicy *trustpolicy.OCIDocument, trustStore truststore.X509TrustStore, pluginManager plugin.Manager) (notation.Verifier, error) {`

verifier/verifier.go
priteshbandi: Document is alias for OCIDocument so should be backward compatible. What am I missing?

verifier/verifier.go
```diff
@@ -43,102 +44,235 @@ import (
...
+// the NewVerifierWithOptions constructor
 type VerifierOptions struct {
 	// RevocationClient is an implementation of revocation.Revocation to use for
 	// verifying revocation
 	RevocationClient revocation.Revocation
 }
 
-// NewFromConfig returns a verifier based on local file system
-func NewFromConfig() (notation.Verifier, error) {
+// NewOCIVerifierFromConfig returns a OCI Verifier based on local file system
```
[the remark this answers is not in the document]
priteshbandi: There is no `OCIVerifier` struct, we are reusing `verifier` struct for both blob and oci artifact verification. The difference is if verifier was created for oci, you cannot verify blob, vice versa.

verifier/verifier.go
priteshbandi: because trust policy for blob is different from oci and we want to reuse method.

verifier/verifier.go
priteshbandi: We cannot because verifier implements both Verifier or BlobVerifier interfaces.

verifier/trustpolicy/blob.go
priteshbandi: That's a good point, we already have this issue for OCI trust policy so IMO we should address it outside of this PR.

verifier/trustpolicy/blob.go
priteshbandi: Done, made changes for blob policy.

verifier/trustpolicy/blob.go
Two-Hearts: A common model for trust policy is desired here. If we have multiple trust policy versions, all of them can be transformed to the common model internally. I agree that this can be done in a separate PR in the future.

verifier/verifier.go
```diff
@@ -43,90 +44,229 @@ import (
...
-func NewWithOptions(trustPolicy *trustpolicy.Document, trustStore truststore.X509TrustStore, pluginManager plugin.Manager, opts VerifierOptions) (notation.Verifier, error) {
-	revocationClient := opts.RevocationClient
+// Deprecated: NewWithOptions function exists for historical compatibility and should not be used.
+// To create verifier, use NewVerifierWithOptions function.
+func NewWithOptions(ociTrustPolicy *trustpolicy.OCIDocument, trustStore truststore.X509TrustStore, pluginManager plugin.Manager, opts VerifierOptions) (notation.Verifier, error) {
+	return NewVerifierWithOptions(ociTrustPolicy, nil, trustStore, pluginManager, opts)
+}
+
+// NewVerifier creates a new verifier given ociTrustPolicy, trustStore and pluginManager
+func NewVerifier(ociTrustPolicy *trustpolicy.OCIDocument, blobTrustPolicy *trustpolicy.BlobDocument, trustStore truststore.X509TrustStore, pluginManager plugin.Manager) (*verifier, error) {
```
rgnote, line 109: Mark as deprecated?

verifier/verifier.go
```diff
@@ -43,90 +44,229 @@ import (
...
+	return NewVerifierWithOptions(ociTrustPolicy, blobTrustPolicy, trustStore, pluginManager, VerifierOptions{})
+}
+
+// NewVerifierWithOptions creates a new verifier given ociTrustPolicy, blobTrustPolicy,
+// trustStore, pluginManager, and verifierOptions
+func NewVerifierWithOptions(ociTrustPolicy *trustpolicy.OCIDocument, blobTrustPolicy *trustpolicy.BlobDocument, trustStore truststore.X509TrustStore, pluginManager plugin.Manager, verifierOptions VerifierOptions) (*verifier, error) {
```
rgnote, line 115: Mark as deprecated?

verifier/verifier.go
priteshbandi: Its the new recommended one

verifier/verifier.go
priteshbandi: Its the new recommended one

verifier/verifier.go
```diff
@@ -43,90 +44,229 @@ import (
...
+		errMsg := fmt.Sprintf("failed to generate descriptor for given artifact. Error: %s", err)
+		logger.Error(errMsg)
+		descErr := errors.New(errMsg)
+		outcome.Error = descErr
+		return outcome, descErr
+	}
+
+	if desc.Digest != payload.TargetArtifact.Digest || desc.Size != payload.TargetArtifact.Size ||
+		(desc.MediaType != "" && desc.MediaType != payload.TargetArtifact.MediaType) {
+		logger.Infof("payload.TargetArtifact in signature: %+v", payload.TargetArtifact)
```
rgnote, line 254:
```suggestion
		logger.Infof("payload present in the signature: %+v", payload.TargetArtifact)
```

verifier/verifier.go
```diff
@@ -43,90 +44,229 @@ import (
...
+		logger.Infof("Target blob that want to be verify: %+v", desc)
```
rgnote, line 255:
```suggestion
		logger.Infof("payload derived from the blob: %+v", desc)
```

verifier/verifier.go
```diff
@@ -43,90 +44,229 @@ import (
...
+		outcome.Error = errors.New("signature integrity check failed")
```
rgnote, line 256:
```suggestion
		outcome.Error = errors.New("integrity check failed. signature does not match the given blob")
```

rgnote (dismissed):
LGTM

verifier/verifier.go
JeyJeyGao: Sure. Resolved.

verifier/trustpolicy/blob.go
priteshbandi: Tracking it in https://github.com/notaryproject/notation-go/issues/415

verifier/verifier.go
```diff
@@ -43,90 +44,229 @@ import (
...
 }
 
-// Verify verifies the signature blob `signature` against the target OCI
+// VerifyBlob verifies the signature of given blob , and returns the outcome upon
+// successful verification.
+func (v *verifier) VerifyBlob(ctx context.Context, descGenFunc notation.BlobDescriptorGenerator, signature []byte, opts notation.BlobVerifierVerifyOptions) (*notation.VerificationOutcome, error) {
+	logger := log.GetLogger(ctx)
+	logger.Debugf("Verify signature of media type %v", opts.SignatureMediaType)
+
+	var trustPolicy *trustpolicy.BlobTrustPolicy
```
Two-Hearts, line 197: There should be a sanity check on `*trustpolicy.BlobTrustPolicy` for `VerifyBlob`. If `v.blobTrustPolicyDoc` is nil, the verification should fail immediately. This can happen because a user can use new an `OCI verifier`, but call `VerifyBlob` on it.

verifier/verifier.go
```diff
@@ -138,7 +278,7 @@ func (v *verifier) Verify(ctx context.Context, desc ocispec.Descriptor, signatur
 	logger := log.GetLogger(ctx)
 
 	logger.Debugf("Verify signature against artifact %v referenced as %s in signature media type %v", desc.Digest, artifactRef, envelopeMediaType)
-	trustPolicy, err := v.trustPolicyDoc.GetApplicableTrustPolicy(artifactRef)
+	trustPolicy, err := v.ociTrustPolicyDoc.GetApplicableTrustPolicy(artifactRef)
```
Two-Hearts, line 281: Same here, check if `v.ociTrustPolicyDoc == nil` and fail fast.

verifier/verifier.go
```diff
@@ -254,7 +394,19 @@ func (v *verifier) processSignature(ctx context.Context, sigBlob []byte, envelop
 
 	// verify x509 trust store based authenticity
 	logger.Debug("Validating cert chain")
-	authenticityResult := verifyAuthenticity(ctx, trustPolicy, v.trustStore, outcome)
+	sigScheme := outcome.EnvelopeContent.SignerInfo.SignedAttributes.SigningScheme
+	trustCerts, err := loadX509TrustStores(ctx, sigScheme, policyName, trustStores, v.trustStore)
```
Two-Hearts, line 398:
```suggestion
	trustCerts, err := loadX509TrustStores(ctx, outcome.EnvelopeContent.SignerInfo.SignedAttributes.SigningScheme, policyName, trustStores, v.trustStore)
```

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,144 @@
...
+// GetGlobalTrustPolicy returns a pointer to the deep copy of the TrustPolicy that is marked as global policy
+// see https://github.com/notaryproject/notaryproject/blob/v1.1.0/specs/trust-store-trust-policy.md#blob-trust-policy
+func (policyDoc *BlobDocument) GetGlobalTrustPolicy() (*BlobTrustPolicy, error) {
+	for _, policyStatement := range policyDoc.TrustPolicies {
+		if policyStatement.GlobalPolicy {
+			return (&policyStatement).clone(), nil
+		}
+	}
+
+	return nil, fmt.Errorf("no global blob trust policy")
```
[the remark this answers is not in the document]
priteshbandi: Tracking it in https://github.com/notaryproject/notation-go/pull/394#discussion_r1650274908

notation.go
```diff
@@ -358,6 +383,55 @@ type VerifyOptions struct {
...
+func VerifyBlob(ctx context.Context, blobVerifier BlobVerifier, blobReader io.Reader, signature []byte, verifyBlobOpts VerifyBlobOptions) (ocispec.Descriptor, *VerificationOutcome, error) {
+	if blobVerifier == nil {
+		return ocispec.Descriptor{}, nil, errors.New("blobVerifier cannot be nil")
+	}
+
+	if blobReader == nil {
+		return ocispec.Descriptor{}, nil, errors.New("blobReader cannot be nil")
+	}
+
+	if signature == nil || len(signature) == 0 {
```
JeyJeyGao, line 409:
```suggestion
	if len(signature) == 0 {
```

https://pkg.go.dev/builtin#len
> if v is nil, len(v) is zero.

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,144 @@
...
+
+	// Validate the policy according to 1.0 rules
+	if len(policyDoc.TrustPolicies) == 0 {
+		return errors.New("blob trust policy document can not have zero trust policy statements")
+	}
+
+	policyNames := set.New[string]()
+	var foundGlobalPolicy bool
+	for _, statement := range policyDoc.TrustPolicies {
+		// Verify unique policy statement names across the policy document
```
JeyJeyGao, line 86: Do we need to check that the name is empty or not.

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,144 @@
...
+			return fmt.Errorf("multiple blob trust policy statements use the same name %q, statement names must be unique", statement.Name)
+		}
+
+		if err := validatePolicyCore(statement.Name, statement.SignatureVerification, statement.TrustStores, statement.TrustedIdentities); err != nil {
+			return fmt.Errorf("blob trust policy: %w", err)
+		}
+
+		if statement.GlobalPolicy {
+			if foundGlobalPolicy {
+				return errors.New("multiple blob trust policy statements have globalPolicy set to true. Only one trust policy statement should be marked as global policy")
```
Two-Hearts, line 97: nit:
```suggestion
				return errors.New("multiple blob trust policy statements have globalPolicy set to true. Only one trust policy statement can be marked as global policy")
```

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,144 @@
...
+		return  nil, errors.New("policy name cannot be empty")
+	}
+	for _, policyStatement := range policyDoc.TrustPolicies {
+		// exact match
+		if policyStatement.Name == policyName {
+			return (&policyStatement).clone(), nil
+		}
+	}
+
+	return nil, fmt.Errorf("no applicable blob trust policy. Applicability for a given blob is determined by policy name")
```
Two-Hearts, line 120: We should explicitly show which trust policy was searched and not found.
```suggestion
	return nil, fmt.Errorf("no applicable blob trust policy with name %q", policyName)
```

Two-Hearts (commented):
Overall LGTM, with a few suggestions left.

verifier/trustpolicy/blob.go
priteshbandi: validatePolicyCore method call at line 91 already does that

verifier/trustpolicy/blob.go
```diff
@@ -0,0 +1,144 @@
...
+		if policyNames.Contains(statement.Name) {
+			return fmt.Errorf("multiple blob trust policy statements use the same name %q, statement names must be unique", statement.Name)
+		}
+
+		if err := validatePolicyCore(statement.Name, statement.SignatureVerification, statement.TrustStores, statement.TrustedIdentities); err != nil {
+			return fmt.Errorf("blob trust policy: %w", err)
+		}
+
+		if statement.GlobalPolicy {
+			if foundGlobalPolicy {
```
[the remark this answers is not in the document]
priteshbandi: Good catch; fixed.

Two-Hearts (approved):
LGTM

Two-Hearts (approved):
LGTM

verifier/trustpolicy/oci.go
```diff
@@ -0,0 +1,257 @@
...
+// To load OCI Document, use LoadOCIDocument function.
+var LoadDocument = LoadOCIDocument
+
+var supportedOCIPolicyVersions = []string{"1.0"}
+
+// LoadOCIDocument retrieves a trust policy document from the local file system.
+// It attempts to read from dir.PathOCITrustPolicy first; if not found, it tries dir.PathTrustPolicy.
+// If both dir.PathOCITrustPolicy and dir.PathTrustPolicy exist, dir.PathOCITrustPolicy will be read.
+func LoadOCIDocument() (*OCIDocument, error) {
+
```
JeyJeyGao, line 76: nit: could you remove the empty line.

JeyJeyGao (approved):
LGTM with a nit

verifier/verifier.go
```diff
@@ -43,90 +44,232 @@ import (
...
-		trustStore:       trustStore,
-		pluginManager:    pluginManager,
-		revocationClient: revocationClient,
+		ociTrustPolicyDoc:  ociTrustPolicy,
+		blobTrustPolicyDoc: blobTrustPolicy,
+		trustStore:         trustStore,
+		pluginManager:      pluginManager,
+		revocationClient:   revocationClient,
 	}, nil
+
```
JeyJeyGao, line 152: Also here.

verifier/trustpolicy/oci.go
priteshbandi: Will do in next revision

verifier/verifier.go
priteshbandi: Will do in next revision

PR #402, opened 2024-05-11
fix: error message for dangling reference index

### LINKED ISSUE #400: Push successful but logged an error
### What is not working as expected?

```
$ notation sign --key  wabbit-networks-ca-issued $image -v
INFO Using the referrers tag schema
INFO Reference v1 resolved to manifest descriptor: {MediaType:application/vnd.docker.distribution.manifest.v2+json Digest:sha256:43837354f3f9a79b20b430b8ce4b6cc43c52dbabed9e88775b53697fe560c329 Size:1782 URLs:[] Annotations:map[] Data:[] Platform:<nil> ArtifactType:}
Warning: Always sign the artifact using digest(@sha256:...) rather than a tag(:v1) because tags are mutable and a tag reference can point to a different artifact than the one signed.
ERRO Failed to push the signature
Warning: Removal of outdated referrers index from remote registry failed. Garbage collection may be required.
Successfully signed wabbitregistry.azurecr.io/net-monitor@sha256:43837354f3f9a79b20b430b8ce4b6cc43c52dbabed9e88775b53697fe560c329
```

ERRO Failed to push the signature
......
Successfully signed 

Which is confusing.

### What did you expect to happen?

For this case: 
Warning: Removal of outdated referrers index from remote registry failed. Garbage collection may be required.

We should not log an error.

### How can we reproduce it?

re-sign an artifact in a registry that doesn't support deleting blob with digest

### Describe your environment

Linux

### What is the version of your notation-go Library?

1.1.0

JeyJeyGao:
Test example:
```
notation sign docker.io/xxx/hello:v1 --plugin azure-kv --id https://xxx.vault.azure.net/certificates/xxx/xxx --plugin-config credential_type=azurecli -v
INFO The referrers tag schema is always attempted 
INFO Reference v1 resolved to manifest descriptor: {MediaType:application/vnd.docker.distribution.manifest.v2+json Digest:sha256:f54a58bc1aac5ea1a25d796ae155dc228b3f0e11d046ae276b39c4bf2f13d8c4 Size:525 URLs:[] Annotations:map[] Data:[] Platform:<nil> ArtifactType:} 
Warning: Always sign the artifact using digest(@sha256:...) rather than a tag(:v1) because tags are mutable and a tag reference can point to a different artifact than the one signed.
Warning: Removal of outdated referrers index from remote registry failed. Garbage collection may be required.
Successfully signed docker.io/mxd950630/hello@sha256:f54a58bc1aac5ea1a25d796ae155dc228b3f0e11d046ae276b39c4bf2f13d8c4
```

Resolves #400 


notation.go
```diff
@@ -41,6 +41,9 @@ import (
 )
 
 var errDoneVerification = errors.New("done verification")
+
+const errReferrersTagSchemaDelete = "failed to delete dangling referrers index"
```
Two-Hearts, line 45: I'm curious if there's a corresponding error type defined in `oras-go`? If yes, we should use that one.

Two-Hearts (commented):
Just a note, this PR depends on PR #405.

notation.go
JeyJeyGao: Updated.

notation.go
```diff
@@ -166,7 +168,12 @@ func Sign(ctx context.Context, signer Signer, repo registry.Repository, signOpts
 	logger.Debugf("Pushing signature of artifact descriptor: %+v, signature media type: %v", targetDesc, signOpts.SignatureMediaType)
 	_, _, err = repo.PushSignature(ctx, signOpts.SignatureMediaType, sig, targetDesc, annotations)
 	if err != nil {
-		logger.Error("Failed to push the signature")
+		var referrerError *remote.ReferrersError
+		if errors.As(err, &referrerError) && referrerError.IsReferrersIndexDelete() {
+			logger.Warn("The signature has been attached to the artifact, but it failed to delete the dangling referrers index.")
+		} else {
+			logger.Error("Failed to push the signature")
```
Two-Hearts, line 175: nit:
We need to align our log format. I'm seeing some log ends with a `.`, others don't. Which format is desired here?
And we'd like all logs to begin with a capital letter.

JeyJeyGao: Removed `.`

Two-Hearts (approved):
LGTM

priteshbandi (approved):

PR #423, opened 2024-07-15
feat: CRL

JeyJeyGao:


verifier/verifier.go
```diff
@@ -1019,7 +1087,27 @@ func verifyTimestamp(ctx context.Context, policyName string, trustStores []strin
...
+			return errors.New("revocation client does not support OCSP")
+		}
+	case trustpolicy.RevocationModeCRL:
+		logger.Debug("Revocation mode is set to CRL")
+		if r, ok := r.(revocation.CRLRevocation); ok {
+			certResults, err = r.ValidateCRLOnly(tsaCertChain, time.Time{})
+		} else {
+			return errors.New("revocation client does not support CRL")
+		}
+	}
```
junczhu, line 1110: Do we need a default case here?

JeyJeyGao: Yes, we need default. This PR is for PoC only. `RevocationMode` will no be supported in v1.3.0

PR #424, opened 2024-07-16
fix: fix usage of SignerInfo.AuthenticSigningTime

Two-Hearts:
This PR updates the use of `SignerInfo.AuthenticSigningTime` due to upgrade of notation-core-go.

### DIFF
--- a/verifier/verifier.go
+++ b/verifier/verifier.go
@@ -721,11 +721,19 @@ func verifyRevocation(outcome *notation.VerificationOutcome, r revocation.Revoca
 		}
 	}
 
-	authenticSigningTime, err := outcome.EnvelopeContent.SignerInfo.AuthenticSigningTime()
-	if err != nil {
-		logger.Debugf("Not using authentic signing time due to error retrieving AuthenticSigningTime, err: %v", err)
-		authenticSigningTime = time.Time{}
+	var authenticSigningTime time.Time
+	var err error
+	if outcome.EnvelopeContent.SignerInfo.SignedAttributes.SigningScheme == signature.SigningSchemeX509SigningAuthority {
+		authenticSigningTime, err = outcome.EnvelopeContent.SignerInfo.AuthenticSigningTime()
+		if err != nil {
+			return &notation.ValidationResult{
+				Type:   trustpolicy.TypeRevocation,
+				Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeRevocation],
+				Error:  fmt.Errorf("unable to check revocation status, err: %s", err.Error()),
+			}
+		}
 	}
+
 	certResults, err := r.Validate(outcome.EnvelopeContent.SignerInfo.CertificateChain, authenticSigningTime)
 	if err != nil {
 		logger.Debug("Error while checking revocation status, err: %s", err.Error())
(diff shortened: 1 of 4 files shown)


verifier/verifier.go
```diff
@@ -721,11 +721,19 @@ func verifyRevocation(outcome *notation.VerificationOutcome, r revocation.Revoca
 		}
 	}
 
-	authenticSigningTime, err := outcome.EnvelopeContent.SignerInfo.AuthenticSigningTime()
-	if err != nil {
-		logger.Debugf("Not using authentic signing time due to error retrieving AuthenticSigningTime, err: %v", err)
-		authenticSigningTime = time.Time{}
+	var authenticSigningTime time.Time
+	var err error
+	if outcome.EnvelopeContent.SignerInfo.SignedAttributes.SigningScheme == signature.SigningSchemeX509SigningAuthority {
```
shizhMSFT, line 726: nit:
```suggestion
	if outcome.EnvelopeContent.SignerInfo.SignedAttributes.SigningScheme == signature.SigningSchemeX509SigningAuthority {
		var err error
```

verifier/verifier.go
```diff
@@ -721,11 +721,19 @@ func verifyRevocation(outcome *notation.VerificationOutcome, r revocation.Revoca
...
+		authenticSigningTime, err = outcome.EnvelopeContent.SignerInfo.AuthenticSigningTime()
+		if err != nil {
+			return &notation.ValidationResult{
+				Type:   trustpolicy.TypeRevocation,
+				Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeRevocation],
+				Error:  fmt.Errorf("unable to check revocation status, err: %s", err.Error()),
+			}
+		}
```
priteshbandi, line 734: This error looks wrong; for signing authority there is no revocation check for stimestamp certs as there is no timestamp signature. 
Also authentic signing time must have been checked before `verifyRevocation` call, so in what case this error will be thrown? [ref](https://github.com/notaryproject/notation-go/blob/b52583166f2b98c75899356aaf4964e98c261c48/verifier/verifier.go#L697)

Two-Hearts: @priteshbandi The original logic is calling [`SignerInfo.AuthenticSigningTime`](https://github.com/notaryproject/notation-core-go/blob/f45197cbd53b9cbaf867cfe5152e3b8cf4ea9766/signature/types.go#L207). Since we upgraded notation-core-go, we need to update its usage in notation-go.

Since it's under `notary.x509.signingAuthority`, the only scenario that would trigger an error is `signerInfo.SignedAttributes.SigningTime.IsZero() == true`, i.e., there's no `signingTime` in the signature envelope. So yes, since we should already checked this before revocation, this error should never be triggered.

>This error looks wrong; for signing authority there is no revocation check for stimestamp certs as there is no timestamp signature.

Yes, this change is not related to timestamp as it's a concept under `notary.x509`.

shizhMSFT (approved):
LGTM

JeyJeyGao (approved):
LGTM

verifier/verifier.go
priteshbandi: Since we know that this wold never happen, can we remove this check?

Two-Hearts: Updated.

JeyJeyGao (approved):
LGTM

priteshbandi (approved):
LGTM

PR #429, opened 2024-07-31
refactor!: update revocation based on notation-core-go

Two-Hearts:
This PR makes refactoring of revocation with the updated notation-core-go.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,4 +25,4 @@ require (
 	golang.org/x/sync v0.6.0 // indirect
 )
 
-replace github.com/notaryproject/notation-core-go => github.com/Two-Hearts/notation-core-go v0.0.0-20240731063103-74357d901612
+replace github.com/notaryproject/notation-core-go => github.com/Two-Hearts/notation-core-go v0.0.0-20240801005628-17703be86fe1
--- a/go.mod
+++ b/go.mod
@@ -25,4 +25,4 @@ require (
 	golang.org/x/sync v0.6.0 // indirect
 )
 
-replace github.com/notaryproject/notation-core-go => github.com/Two-Hearts/notation-core-go v0.0.0-20240730042226-e19d36b91b8d
+replace github.com/notaryproject/notation-core-go => github.com/Two-Hearts/notation-core-go v0.0.0-20240730073616-5d5043a8f812
--- a/go.mod
+++ b/go.mod
@@ -25,4 +25,4 @@ require (
 	golang.org/x/sync v0.6.0 // indirect
 )
 
-replace github.com/notaryproject/notation-core-go => github.com/Two-Hearts/notation-core-go v0.0.0-20240801005628-17703be86fe1
+replace github.com/notaryproject/notation-core-go => github.com/Two-Hearts/notation-core-go v0.0.0-20240806060916-8a09715b8a0b
--- a/go.mod
+++ b/go.mod
@@ -25,4 +25,4 @@ require (
 	golang.org/x/sync v0.6.0 // indirect
 )
 
-replace github.com/notaryproject/notation-core-go => github.com/Two-Hearts/notation-core-go v0.0.0-20240730073616-5d5043a8f812
+replace github.com/notaryproject/notation-core-go => github.com/Two-Hearts/notation-core-go v0.0.0-20240731063103-74357d901612
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.21
 
 require (
 	github.com/go-ldap/ldap/v3 v3.4.8
-	github.com/notaryproject/notation-core-go v1.1.0-beta.1.0.20240808015012-004b86dbf289
+	github.com/notaryproject/notation-core-go v1.1.0-rc.1
 	github.com/notaryproject/notation-plugin-framework-go v1.0.0
 	github.com/notaryproject/tspclient-go v0.2.0
 	github.com/opencontainers/go-digest v1.0.0
--- a/go.mod
+++ b/go.mod
@@ -25,4 +25,4 @@ require (
 	golang.org/x/sync v0.6.0 // indirect
 )
 
-replace github.com/notaryproject/notation-core-go => github.com/Two-Hearts/notation-core-go v0.0.0-20240806060916-8a09715b8a0b
+replace github.com/notaryproject/notation-core-go => github.com/Two-Hearts/notation-core-go v0.0.0-20240807034940-02ab652946b1
--- a/go.mod
+++ b/go.mod
@@ -24,3 +24,5 @@ require (
 	github.com/x448/float16 v0.8.4 // indirect
 	golang.org/x/sync v0.6.0 // indirect
 )
+
+replace github.com/notaryproject/notation-core-go => github.com/Two-Hearts/notation-core-go v0.0.0-20240730042226-e19d36b91b8d
--- a/go.mod
+++ b/go.mod
@@ -4,9 +4,9 @@ go 1.21
 
 require (
 	github.com/go-ldap/ldap/v3 v3.4.8
-	github.com/notaryproject/notation-core-go v1.1.0-beta.1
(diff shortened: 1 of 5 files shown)


verifier/verifier.go
```diff
@@ -120,68 +130,89 @@ func NewVerifier(ociTrustPolicy *trustpolicy.OCIDocument, blobTrustPolicy *trust
...
+// New creates a new verifier given ociTrustPolicy, trustStore and pluginManager.
+//
 // Deprecated: New function exists for historical compatibility and should not be used.
 // To create verifier, use NewVerifier function.
 func New(ociTrustPolicy *trustpolicy.OCIDocument, trustStore truststore.X509TrustStore, pluginManager plugin.Manager) (notation.Verifier, error) {
 	return NewVerifier(ociTrustPolicy, nil, trustStore, pluginManager)
 }
 
+// setCodeSigningRevocation sets code signing revocation object of v
+func (v *verifier) setCodeSigningRevocation(verifierOptions VerifierOptions) error {
```
JeyJeyGao, line 191:
```suggestion
func (v *verifier) setRevocation(verifierOptions VerifierOptions) error {
```

Is it better to move `revocationTimestampingValidator ` to this function, so all revocation client related things are in one place.

go.mod
```diff
@@ -24,3 +24,5 @@ require (
 	github.com/x448/float16 v0.8.4 // indirect
 	golang.org/x/sync v0.6.0 // indirect
 )
+
+replace github.com/notaryproject/notation-core-go => github.com/Two-Hearts/notation-core-go v0.0.0-20240801005628-17703be86fe1
```
shizhMSFT, line 28: This comment is a reminder to remove the temp replacement

verifier/verifier.go
shizhMSFT, line 191: nit: code style: it is better to make this function static

verifier/verifier.go
```diff
@@ -1019,7 +1061,10 @@ func verifyTimestamp(ctx context.Context, policyName string, trustStores []strin
 
 	// 5. Perform the timestamping certificate chain revocation check
 	logger.Debug("Checking timestamping certificate chain revocation...")
-	certResults, err := r.Validate(tsaCertChain, time.Time{})
+	certResults, err := r.ValidateContext(ctx, revocation.ValidateContextOptions{
+		CertChain:   tsaCertChain,
+		SigningTime: time.Time{},
```
shizhMSFT, line 1066: This line seems redundant.

verifier/verifier.go
Two-Hearts: Updated.

Two-Hearts: Updated.

verifier/verifier.go
```diff
@@ -1082,3 +1043,43 @@ func verifyTimestamp(ctx context.Context, policyName string, trustStores []strin
 	// success
 	return nil
 }
+
+// setRevocation sets revocation validators of verifier
+func setRevocation(verifier *verifier, verifierOptions VerifierOptions) error {
```
shizhMSFT, line 1048: Previously, I was thinking of something like
```go
func getRevocationTimestampingValidator(verifierOptions VerifierOptions) (Validator, error) {
```

However, with your current code structure, it makes more sense if you change it back to
```go
func (v *verifier) setRevocation(verifierOptions VerifierOptions) error {
```

Two-Hearts: Updated.

go.mod
Two-Hearts: Removed.

JeyJeyGao (dismissed):
LGTM

shizhMSFT (approved):
LGTM

verifier/verifier.go
```diff
@@ -509,6 +543,59 @@ func (v *verifier) processSignature(ctx context.Context, sigBlob []byte, envelop
...
+			Type:   trustpolicy.TypeRevocation,
+			Action: outcome.VerificationLevel.Enforcement[trustpolicy.TypeRevocation],
+			Error:  fmt.Errorf("unable to check revocation status, code signing revocation validator cannot be nil"),
+		}
+	}
+
+	var authenticSigningTime time.Time
+	if outcome.EnvelopeContent.SignerInfo.SignedAttributes.SigningScheme == signature.SigningSchemeX509SigningAuthority {
+		authenticSigningTime, _ = outcome.EnvelopeContent.SignerInfo.AuthenticSigningTime()
+	}
```
rgnote, line 560: This is for my understanding I think. Why are we considering the authentic signing time only in the case of signing authority but not in notary.x509 case?

Two-Hearts: This PR is not changing the original logic at this part. But I think I could explain more. This is according to the Notary Project spec:  https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#standard-attributes
It says:
```
Authentic Signing Time (critical):  ... This claim is REQUIRED and only valid when signing scheme is notary.x509.signingAuthority. ....
```

JeyJeyGao (approved):
LGTM

verifier/verifier.go
rgnote: I understand that AuthenticSigningTime only applies to `notary.x509.signingAuthority`. But, the equivalent of it in the case of `notary.x509` is the RFC#3161 TSA signature. My question is why are we not taking the TSA signature into account (for the case of `notary.x509` signatures) for revocations?

yizha1: @rgnote Regarding revocations, we've decided to exclude the TSA signature consideration for now. The main reason is that revocation timing is not deterministic, which may not accurately reflect the actual date of invalidation/compromising.

Two-Hearts: `AuthenticSigningTime` and `RFC#3161 TSA signature` are not equivalent though. `RFC#3161 TSA signature` does not provide information about the `signing time`, it's just the time point when the TSA sees the signature. However, `AuthenticSigningTime` is the exact `signing time`.
Another difference is that `AuthenticSigningTime` is a signed attribute. However, `TSA signature` is an unsigned attribute, if we want to use it, we MUST verify the signature before consuming it. Note: Since the `Authentic Timestamp` step can be set to `logged` and `timestamp verification` can be disabled by trust policy, we cannot directly consume it during `Revocation` step.
Therefore, in this version, we tend to not use it during `Revocation`, in this way, whenever the cert status is `revoked`, we just say the cert chain is revoked (this is more strict, so there isn't security concern).

rgnote: > AuthenticSigningTime and RFC#3161 TSA signature are not equivalent though. RFC#3161 TSA signature does not provide information about the signing time, it's just the time point when the TSA sees the signature. However, AuthenticSigningTime is the exact signing time.

RFC#3161 addresses this by rejecting timestamps if they are way too far from the actual signing time. Also, AuthenticSigningTime doesn't necessarily be the exact signing time. It can be set to any timestamp by a Signing Authority (due to a bug for example). For all practical purposes, TSA signature and AuthenticSigingTime are equivalent. The only difference is "who" is attesting those timestamps.


However, thanks for the explanation though. I acknowledge the complexity of using TSA signatures in conjunction with  revocation effective time. However, I don't see this being documented anywhere. I'm worried about users revoking their certs with effective time and assuming that Notation would correctly pass (or fail) signature verification for notary.x509 signatures. If notation is not going to support it, we should atleast call this out in our documentation and also track this as an issue for future consideration. Can we please track this as an issue that Notation does not consider revocation effective time for notary.x509 signatures due to X,Y and Z reasons.  Thanks.

yizha1: @rgnote Thanks. I created a new issue https://github.com/notaryproject/notation-go/issues/436 for tracking.
We already had a document issue, and I added a comment to call out the revocation scenarios. https://github.com/notaryproject/notaryproject.dev/issues/404. Hope it works for you.

rgnote (approved):
LGTM

PR #432, opened 2024-08-02
feat: support distinguished name state(S)

### LINKED ISSUE #431: Support additional short name S for state/province
Can you please open an issue in notation-go to implement this?

_Originally posted by @priteshbandi in https://github.com/notaryproject/specifications/issues/307#issuecomment-2264522207_
            
This issue is for implementing the specification updates done by PR https://github.com/notaryproject/specifications/issues/307

JeyJeyGao:
Resolves #431 

### DIFF
--- a/internal/pkix/pkix.go
+++ b/internal/pkix/pkix.go
@@ -38,7 +38,7 @@ func ParseDistinguishedName(name string) (map[string]string, error) {
 			return nil, fmt.Errorf("distinguished name (DN) %q has multi-valued RDN attributes, remove multi-valued RDN attributes as they are not supported", name)
 		}
 		for _, attribute := range rdn.Attributes {
-			// stateOrProvincename 'S' is an alias for 'ST'
+			// stateOrProvince name 'S' is an alias for 'ST'
 			if attribute.Type == "S" {
 				attribute.Type = "ST"
 			}
--- a/internal/pkix/pkix.go
+++ b/internal/pkix/pkix.go
@@ -38,6 +38,10 @@ func ParseDistinguishedName(name string) (map[string]string, error) {
 			return nil, fmt.Errorf("distinguished name (DN) %q has multi-valued RDN attributes, remove multi-valued RDN attributes as they are not supported", name)
 		}
 		for _, attribute := range rdn.Attributes {
+			// stateOrProvincename 'S' is an alias for 'ST'
+			if attribute.Type == "S" {
+				attribute.Type = "ST"
+			}
 			if attrKeyValue[attribute.Type] == "" {
 				attrKeyValue[attribute.Type] = attribute.Value
 			} else {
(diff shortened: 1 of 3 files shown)


shizhMSFT (changes requested):
This PR only covers the parsing parts. It does not include the comparation between parsed values. For example, does `ST=Shanghai` equal `S=Shanghai`?

JeyJeyGao:
> This PR only covers the parsing parts. It does not include the comparation between parsed values. For example, does `ST=Shanghai` equal `S=Shanghai`?

Yes, you are right? Updated.

shizhMSFT (approved):
LGTM

internal/pkix/pkix.go
```diff
@@ -39,6 +38,10 @@ func ParseDistinguishedName(name string) (map[string]string, error) {
 			return nil, fmt.Errorf("distinguished name (DN) %q has multi-valued RDN attributes, remove multi-valued RDN attributes as they are not supported", name)
 		}
 		for _, attribute := range rdn.Attributes {
+			// stateOrProvince name 'S' is an alias for 'ST'
+			if attribute.Type == "S" {
+				attribute.Type = "ST"
+			}
```
priteshbandi, line 44: Is it possible to have both `S` and `ST` ?

priteshbandi, line 44: We need to add validation for not supporting both ST and S in same trusted identity.

JeyJeyGao: >We need to add validation for not supporting both ST and S in same trusted identity.

line 45-50 has confirmed that there is no duplicated S or ST
```go
			if attrKeyValue[attribute.Type] == "" {
				attrKeyValue[attribute.Type] = attribute.Value
			} else {
				return nil, fmt.Errorf("distinguished name (DN) %q has duplicate RDN attribute for %q, DN can only have unique RDN attributes", name, attribute.Type)
			}
```


> Also, certificate shouldn't contain S; do we need check for that?

Actually, there is not a string value "S" or "ST" written in certificate, instead, they are defined by OID, for example,
Here is an ANS.1 encoding of an certificate subject:
```sh
        SEQUENCE
        {
            SET
            {
                SEQUENCE
                {
                    OBJECT IDENTIFIER=CountryName (2.5.4.6)
                    PRINTABLE STRING='US'
                }
            }
            SET
            {
                SEQUENCE
                {
                    OBJECT IDENTIFIER=StateOrProvinceName (2.5.4.8)
                    PRINTABLE STRING='Washington'
                }
            }
            SET
            {
                SEQUENCE
                {
                    OBJECT IDENTIFIER=LocalityName (2.5.4.7)
                    PRINTABLE STRING='Redmond'
                }
            }
            SET
            {
                SEQUENCE
                {
                    OBJECT IDENTIFIER=OrganizationName (2.5.4.10)
                    PRINTABLE STRING='Microsoft Corporation'
                }
            }
            SET
            {
                SEQUENCE
                {
                    OBJECT IDENTIFIER=CommonName (2.5.4.3)
                    PRINTABLE STRING='Microsoft Time-Stamp PCA 2010'
                }
            }
        }
```
> OBJECT IDENTIFIER=StateOrProvinceName (2.5.4.8)

This OID is rendered as a string by the application; for Windows, it will be "S"; for Golang, it will be "ST". As a result, Golang will never see a certificate with the "S" field. It is only possible if the user writes "S" in the trust identity.

Two-Hearts (approved):
LGTM

priteshbandi (approved):
LGTM

JeyJeyGao:
> LGTM;
> 
> Please sign your commits

Re-signed. Thanks!

PR #437, opened 2024-08-14
refactor!: remove blob sign/verify related code

Two-Hearts:
As discussed, this PR removes the blob sign/verify related code for v1.2.0 release.

This PR targets on branch `release-1.2` ONLY.


verifier/trustpolicy/oci.go
```diff
@@ -53,41 +53,31 @@ type OCITrustPolicy struct {
 }
 
 // Document represents a trustPolicy.json document
+//
 // Deprecated: Document exists for historical compatibility and should not be used.
 // To create OCI Document, use OCIDocument.
```
JeyJeyGao, line 58: Should we remove the comment?

verifier/trustpolicy/oci.go
```diff
@@ -53,41 +53,31 @@ type OCITrustPolicy struct {
...
 type Document = OCIDocument
 
 // TrustPolicy represents a policy statement in the policy document
+//
 // Deprecated: TrustPolicy exists for historical compatibility and should not be used.
 // To create OCI TrustPolicy, use OCITrustPolicy.
```
JeyJeyGao, line 64: Should we remove the comment?

verifier/trustpolicy/oci.go
Two-Hearts: I was thinking about it as well. However, this file is called `oci.go`, and its changes are all related to `OCI` instead of `blob`. I kind of think it's a good refactoring, it makes the trust policy more extensible. Therefore, I kept it.

verifier/trustpolicy/oci.go
Two-Hearts: Tracking the discussion above: https://github.com/notaryproject/notation-go/pull/437#discussion_r1716264902

verifier/trustpolicy/oci.go
JeyJeyGao: I see. However, we are still using the deprecated struct `trustpolicy.TrustPolicy` in the test file. Maybe we should fix that in the stable release.

JeyJeyGao (approved):
LGTM

verifier/trustpolicy/oci.go
Two-Hearts: That's a good point, we should fix that in the `rc` version. Actually, I think we could keep the `trustpolicy.TrustPolicy` as undeprecated and remove the `OCITrustPolicy` for v1.2.0, meanwhile keeping the `oci.go` and `trustpolicy.go` files as they are good refactoring to the `trustpolicy` package.

JeyJeyGao (approved):
LGTM

verifier/trustpolicy/oci.go
Two-Hearts: Updated.

JeyJeyGao (dismissed):
LGTM

verifier/trustpolicy/oci.go
```diff
@@ -25,17 +25,17 @@ import (
...
 
 	// TrustPolicies include each policy statement
-	TrustPolicies []OCITrustPolicy `json:"trustPolicies"`
+	TrustPolicies []TrustPolicy `json:"trustPolicies"`
 }
 
-// OCITrustPolicy represents a policy statement in the policy document for OCI artifacts
-type OCITrustPolicy struct {
+// TrustPolicy represents a policy statement in the policy document
+type TrustPolicy struct {
```
shizhMSFT, line 38: The order of the fields was changed by #394

shizhMSFT: This change can be treated as a breaking change. Should be fixed in both this PR and the `main` branch.

Two-Hearts: @shizhMSFT Oh, that's a good catch. 
I guess it's a breaking change due to the way to initialize a struct in Golang? i.e., they could just follow the field order, without the name, while generating a new struct.

JeyJeyGao (dismissed):
LGTM

shizhMSFT (approved):
LGTM

JeyJeyGao (dismissed):
LGTM

signer/signer.go
```diff
@@ -34,7 +34,7 @@ import (
 )
 
 // signingAgent is the unprotected header field used by signature.
-const signingAgent = "Notation/1.0.0"
```
shizhMSFT, line 37: @Two-Hearts Can you also send out an PR to the `main` branch as well as the `release-1.1` branch? We need to release 1.1.2. Also, we need to send out an PR to the `main` branch for `RELEASE_CHECKLIST.md`. Because of this field, the `notation-go` library requires a release process similar to `notation` CLI instead of `notation-core-go`.

Also, should this be `notation-go/1.2.0` instead?

Good thing is that it is for `GenericSigner`, which is mainly for testing purposes.

signer/signer.go
```diff
@@ -34,14 +34,15 @@ import (
...
+const signingAgent = "Notation/1.2.0"
```
shizhMSFT, line 37: Should it be `notation-go/1.2.0-rc.1`?

Two-Hearts: Updated.

signer/signer.go
Two-Hearts: I've already created an issue to track this: https://github.com/notaryproject/notation-go/issues/438. Will create PR to fix it later.

It impacts all plugins with `SIGNATURE_GENERATOR.RAW` capability, such as `AKV plugin`. When debugging, it is shown in the signing debug log.

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

JeyJeyGao (approved):
LGTM

priteshbandi (approved):
LGTM

Issue #446, opened 2024-08-30
Bug: Notation panics on init when $HOME and $XDG_CONFIG_HOME unset


JasonTheDeveloper:
### What is not working as expected?

The bug relates to an issue raised in https://github.com/fluxcd/flux2/issues/4961

This gist of the issue is when $HOME or $XDC_CONFIG_HOME are unset, `dir/path` panics when initialising https://github.com/notaryproject/notation-go/blob/974c2916fa135b977b02db1714efb6ed2ecec0f3/dir/path.go#L82-L97

### What did you expect to happen?

I would expect notation to not panic, breaking any application importing the notation SDK when $HOME and $XDC_CONFIG_HOME are unset.

### How can we reproduce it?

Sets can be found in the reference issue 

### Describe your environment

OS: Arch Linux, Debian Bookworm
Golang: 1.22.5

### What is the version of your notation-go Library?

v1.1.0

PR #449, opened 2024-09-02
fix: dir no longer panics when HOME and XDG_CONFIG_HOME are not set

JasonTheDeveloper:
This PR addresses the issue #446

In this PR I:

- I removed the `init()` function from `dir/path`
- When `userConfigDir()` returns an error, instead of `panic(err)` I default to the current directory instead
- Split `loadUserPath()` into two new functions used to setup and return the values for `UserConfigDir` and `UserLibexecDir`
- Added additional unit tests for the two new functions and to test the default directory is used when `HOME` is set to `""`

### DIFF
--- a/dir/path.go
+++ b/dir/path.go
@@ -92,7 +92,8 @@ func UserConfigDirPath() string {
 	if UserConfigDir == "" {
 		userDir, err := userConfigDir()
 		if err != nil {
-		panic(err)
+			// fallback to current directory
+			userDir = "."
 		}
 		// set user config
 		UserConfigDir = filepath.Join(userDir, notation)
--- a/dir/path.go
+++ b/dir/path.go
@@ -86,6 +86,8 @@ func userConfigDirPath() string {
 		if err != nil {
 			// fallback to current directory
 			userDir = "."
+			UserConfigDir = "." + notation
+			return UserConfigDir
 		}
 		// set user config
 		UserConfigDir = filepath.Join(userDir, notation)
--- a/dir/path.go
+++ b/dir/path.go
@@ -79,10 +79,6 @@ const (
 
 var userConfigDir = os.UserConfigDir // for unit test
 
-func init() {
-	loadUserPath()
-}
-
 // loadUserPath function defines UserConfigDir and UserLibexecDir.
 func loadUserPath() {
 	// set user config
--- a/dir/path.go
+++ b/dir/path.go
@@ -87,8 +87,6 @@ func userConfigDirPath() string {
 			// fallback to current directory
 			UserConfigDir = "." + notation
 			return UserConfigDir
-			UserConfigDir = "." + notation
-			return UserConfigDir
 		}
 		// set user config
 		UserConfigDir = filepath.Join(userDir, notation)
--- a/dir/path.go
+++ b/dir/path.go
@@ -81,15 +81,32 @@ var userConfigDir = os.UserConfigDir // for unit test
 
 // loadUserPath function defines UserConfigDir and UserLibexecDir.
 func loadUserPath() {
-	// set user config
-	userDir, err := userConfigDir()
-	if err != nil {
-		panic(err)
-	}
-	UserConfigDir = filepath.Join(userDir, notation)
+	UserConfigDir = ""
+	UserLibexecDir = ""
 
-	// set user libexec
-	UserLibexecDir = UserConfigDir
+	_ = UserLibexecDirPath()
+}
+
+// UserConfigDirPath returns the user level {NOTATION_CONFIG} path.
+func UserConfigDirPath() string {
+	if UserConfigDir == "" {
+		userDir, err := userConfigDir()
+		if err != nil {
+		panic(err)
+		}
+		// set user config
+		UserConfigDir = filepath.Join(userDir, notation)
+	}
+	return UserConfigDir
+}
+
+// UserLibexecDirPath returns the user level {NOTATION_LIBEXEC} path.
+func UserLibexecDirPath() string {
+	if UserLibexecDir == "" {
+		// set user libexec
+		UserLibexecDir = UserConfigDirPath()
+	}
+	return UserLibexecDir
 }
 
 // LocalKeyPath returns the local key and local cert relative paths.
--- a/dir/path.go
+++ b/dir/path.go
@@ -84,11 +84,11 @@ func loadUserPath() {
 	UserConfigDir = ""
 	UserLibexecDir = ""
 
-	_ = UserLibexecDirPath()
+	_ = userLibexecDirPath()
 }
 
-// UserConfigDirPath returns the user level {NOTATION_CONFIG} path.
-func UserConfigDirPath() string {
+// userConfigDirPath returns the user level {NOTATION_CONFIG} path.
+func userConfigDirPath() string {
 	if UserConfigDir == "" {
 		userDir, err := userConfigDir()
 		if err != nil {
@@ -101,11 +101,11 @@ func UserConfigDirPath() string {
 	return UserConfigDir
 }
 
-// UserLibexecDirPath returns the user level {NOTATION_LIBEXEC} path.
-func UserLibexecDirPath() string {
+// userLibexecDirPath returns the user level {NOTATION_LIBEXEC} path.
+func userLibexecDirPath() string {
 	if UserLibexecDir == "" {
 		// set user libexec
-		UserLibexecDir = UserConfigDirPath()
+		UserLibexecDir = userConfigDirPath()
 	}
 	return UserLibexecDir
 }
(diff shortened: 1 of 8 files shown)


dir/path.go
```diff
@@ -79,21 +79,35 @@ const (
...
+	_ = UserLibexecDirPath()
+}
+
+// UserConfigDirPath returns the user level {NOTATION_CONFIG} path.
+func UserConfigDirPath() string {
+	if UserConfigDir == "" {
+		userDir, err := userConfigDir()
+		if err != nil {
+			// fallback to current directory
+			userDir = "."
```
JeyJeyGao, line 96: By including this PR and updating the Notation CLI, if the `$HOME` and `$XDG_CONFIG_HOME` variables are not set on Linux, we will use the current working directory as the default config directory. This is different from the previous implementation, where the application would panic in this case.

Comment for PM @yizha1 @FeynmanZhou

dir/path.go
```diff
@@ -79,21 +79,35 @@ const (
 
 var userConfigDir = os.UserConfigDir // for unit test
 
-func init() {
-	loadUserPath()
-}
-
 // loadUserPath function defines UserConfigDir and UserLibexecDir.
 func loadUserPath() {
```
JeyJeyGao, line 83: Looks like we don't need this function now. If we don't need it, we can remove it as it is an internal function.

dir/path.go
```diff
@@ -79,21 +79,35 @@ const (
...
+// UserConfigDirPath returns the user level {NOTATION_CONFIG} path.
+func UserConfigDirPath() string {
+	if UserConfigDir == "" {
+		userDir, err := userConfigDir()
+		if err != nil {
+			// fallback to current directory
+			userDir = "."
+		}
+		// set user config
+		UserConfigDir = filepath.Join(userDir, notation)
```
JeyJeyGao, line 99: Thank you for adding the lazy initialization! 👍

dir/path.go
```diff
@@ -79,21 +79,35 @@ const (
...
-}
-
 // loadUserPath function defines UserConfigDir and UserLibexecDir.
 func loadUserPath() {
-	// set user config
-	userDir, err := userConfigDir()
-	if err != nil {
-		panic(err)
+	UserConfigDir = ""
+	UserLibexecDir = ""
```
JeyJeyGao, line 85: The string default value is empty string, so we may don't need this set the empty value?

dir/path.go
```diff
@@ -79,21 +79,35 @@ const (
...
+
+	_ = UserLibexecDirPath()
```
JeyJeyGao, line 87: We may don't need to initialize it as it is lazy now.

dir/path.go
```diff
@@ -79,21 +79,35 @@ const (
...
+}
+
+// UserConfigDirPath returns the user level {NOTATION_CONFIG} path.
+func UserConfigDirPath() string {
```
JeyJeyGao, line 91:
```suggestion
func userConfigDirPath() string {
```

Can we set it as an internal function? If use want to get the config dir, they can call `ConfigFS().SysPath()`

dir/path.go
```diff
@@ -79,21 +79,35 @@ const (
...
+		UserConfigDir = filepath.Join(userDir, notation)
 	}
-	UserConfigDir = filepath.Join(userDir, notation)
+	return UserConfigDir
+}
 
-	// set user libexec
-	UserLibexecDir = UserConfigDir
+// UserLibexecDirPath returns the user level {NOTATION_LIBEXEC} path.
+func UserLibexecDirPath() string {
```
JeyJeyGao, line 105:
```suggestion
func userLibexecDirPath() string {
```
Can we set it as an internal function? If use want to get the config dir, they can call `PluginFS().SysPath()`

JasonTheDeveloper: Resolved in [5281e07](https://github.com/notaryproject/notation-go/pull/449/commits/5281e07b3601e70fe467b000158ea7862e44b713).

dir/path.go
JasonTheDeveloper: Resolved in [5281e07](https://github.com/notaryproject/notation-go/pull/449/commits/5281e07b3601e70fe467b000158ea7862e44b713).

dir/path.go
JasonTheDeveloper: Agreed. I left it in for the sake of the unit tests but I've since removed it. Resolved in [46c35d4](https://github.com/notaryproject/notation-go/pull/449/commits/46c35d45a2c830ad16861f7ab00b6a9ccb8aec9d).

dir/path.go
JasonTheDeveloper: This was more to reset the values but I've removed the func in [46c35d4](https://github.com/notaryproject/notation-go/pull/449/commits/46c35d45a2c830ad16861f7ab00b6a9ccb8aec9d).

dir/path.go
JasonTheDeveloper: Resolved in [46c35d4](https://github.com/notaryproject/notation-go/pull/449/commits/46c35d45a2c830ad16861f7ab00b6a9ccb8aec9d).

JeyJeyGao (dismissed):
LGTM

dir/path.go
yizha1: We need to update this document https://notaryproject.dev/docs/user-guides/how-to/directory-structure/ to show how the config directory is determined.

dir/path.go
```diff
@@ -79,21 +79,27 @@ const (
...
 
-func init() {
-	loadUserPath()
+// userConfigDirPath returns the user level {NOTATION_CONFIG} path.
+func userConfigDirPath() string {
+	if UserConfigDir == "" {
+		userDir, err := userConfigDir()
+		if err != nil {
+			// fallback to current directory
+			userDir = "."
```
shizhMSFT, line 88: This will result in `./notation`. Would it be better to have `.notation` as `UserConfigDir` as a more Unix/Linux way?

JasonTheDeveloper: The current behaviour if `$HOME` is set would result in `$HOME/notation`. Would you rather change the notation folder be named `.notation` instead? So, if `$HOME` is set, the result would be `$HOME/.notation`. Or are you asking if `$HOME` is not set, the result should be `./.notation` otherwise the result should be `$HOME/notation`?

shizhMSFT: It's not quite right. According to [os.UserConfigDir()](https://pkg.go.dev/os#UserConfigDir) doc,
> On Unix systems, it returns $XDG_CONFIG_HOME as specified by https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html if non-empty, else $HOME/.config. On Darwin, it returns $HOME/Library/Application Support. On Windows, it returns %AppData%. On Plan 9, it returns $home/lib.

If `$HOME` is set, the result would be `$HOME/.config/notation` on Linux.

I am asking if `$HOME` is not set, the result should be `./.notation`. Otherwise, the result should be `$HOME/.config/notation`.

JasonTheDeveloper: How about something like this?

```go
if UserConfigDir == "" {
	notationDir := notation
	userDir, err := userConfigDir()
	if err != nil {
		// fallback to current directory
		userDir = "."
		notationDir = fmt.Sprintf(".%s", notation) 
	}
	// set user config
	UserConfigDir = filepath.Join(userDir, notationDir)
}
```

shizhMSFT: It's getting more complex. How about
```suggestion
			UserConfigDir = "." + notation
			return UserConfigDir
```

dir/path.go
JasonTheDeveloper: @JeyJeyGao see https://github.com/notaryproject/notation-go/pull/449#discussion_r1741356288. In summary, rather than `panic`, if `$HOME` and `$XDG_CONFIG_HOME` are not set, `./.notation` is used instead. Otherwise, `$HOME/.config/notation` is used.

Two-Hearts (approved):
LGTM

JeyJeyGao (approved):
LGTM

dir/path.go
JasonTheDeveloper: I created a new issue to track the documentation change requirements https://github.com/notaryproject/notaryproject.dev/issues/413

JeyJeyGao:
@JasonTheDeveloper 

> The base branch requires all commits to be signed

You need to fix it.

JasonTheDeveloper:
> @JasonTheDeveloper
> 
> > The base branch requires all commits to be signed
> 
> You need to fix it.

All my commits are signed though? DCO also [passed](https://github.com/notaryproject/notation-go/pull/449/checks?check_run_id=29589053252) the sign off checks. Not sure why GitHub is saying `The base branch requires all commits to be signed.`

image

priteshbandi (commented):
LGTM

priteshbandi (approved):

PR #451, opened 2024-09-04
fix: dir no longer panics when HOME and XDG_CONFIG_HOME are not set

Two-Hearts:
This PR is for `release-1.2` branch only.

This PR cherry picks the fix of issue #446 from main.

The fix:
- Removed the `init()` function from `dir/path`
- When `userConfigDir()` returns an error, instead of `panic(err)`, default to the current directory instead
- Split `loadUserPath()` into two new functions used to setup and return the values for `UserConfigDir` and `UserLibexecDir`
- Added additional unit tests for the two new functions and to test the default directory is used when `HOME` is set to `""`

### DIFF
--- a/dir/fs.go
+++ b/dir/fs.go
@@ -51,10 +51,10 @@ func NewSysFS(root string) SysFS {
 
 // ConfigFS is the config SysFS
 func ConfigFS() SysFS {
-	return NewSysFS(UserConfigDir)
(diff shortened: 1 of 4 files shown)


JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

PR #462, opened 2024-09-20
feat: crl cache

Two-Hearts:
This PR adds an implementation of CRL cache along with {NOTATION_CACHE} directory for callers to initialize the cache.

### DIFF
--- a/dir/fs.go
+++ b/dir/fs.go
@@ -59,7 +59,7 @@ func PluginFS() SysFS {
 	return NewSysFS(filepath.Join(userLibexecDirPath(), PathPlugins))
 }
 
-// CacheFS is the cache SysFS
-func CacheFS() SysFS {
-	return NewSysFS(userCacheDirPath())
+// CRLFileCacheFS is the crl file cache SysFS
+func CRLFileCacheFS() SysFS {
+	return NewSysFS(filepath.Join(userCacheDirPath(), PathCRLFileCache))
 }
--- a/dir/fs.go
+++ b/dir/fs.go
@@ -58,3 +58,8 @@ func ConfigFS() SysFS {
 func PluginFS() SysFS {
 	return NewSysFS(filepath.Join(userLibexecDirPath(), PathPlugins))
 }
+
+// CacheFS is the cache SysFS
+func CacheFS() SysFS {
+	return NewSysFS(userCacheDirPath())
+}
--- a/dir/fs.go
+++ b/dir/fs.go
@@ -61,8 +61,7 @@ func PluginFS() SysFS {
 
 // CacheFS is the cache SysFS.
 //
-// To get the directory for crl file cache, one should use
-// `CacheFS().SysFS(PathCRLFileCache)`.
+// To get the root of crl file cache, use `CacheFS().SysFS(PathCRLCache)`.
 func CacheFS() SysFS {
 	return NewSysFS(userCacheDirPath())
 }
--- a/dir/fs.go
+++ b/dir/fs.go
@@ -59,7 +59,7 @@ func PluginFS() SysFS {
 	return NewSysFS(filepath.Join(userLibexecDirPath(), PathPlugins))
 }
 
-// CRLFileCacheFS is the crl file cache SysFS
-func CRLFileCacheFS() SysFS {
-	return NewSysFS(filepath.Join(userCacheDirPath(), PathCRLFileCache))
+// CacheFS is the cache SysFS
+func CacheFS() SysFS {
+	return NewSysFS(userCacheDirPath())
 }
--- a/dir/fs.go
+++ b/dir/fs.go
@@ -59,7 +59,10 @@ func PluginFS() SysFS {
 	return NewSysFS(filepath.Join(userLibexecDirPath(), PathPlugins))
 }
 
-// CacheFS is the cache SysFS
+// CacheFS is the cache SysFS.
+//
+// To get the directory for crl file cache, one should use
+// `CacheFS().SysFS(PathCRLFileCache)`.
 func CacheFS() SysFS {
 	return NewSysFS(userCacheDirPath())
 }
--- a/dir/path.go
+++ b/dir/path.go
@@ -66,8 +66,6 @@ const (
 	PathOCITrustPolicy = "trustpolicy.oci.json"
 	// PathBlobTrustPolicy is the Blob trust policy file relative path.
 	PathBlobTrustPolicy = "trustpolicy.blob.json"
-	// PathPlugins is the plugins directory relative path.
-	PathPlugins = "plugins"
 	// LocalKeysDir is the directory name for local key relative path.
 	LocalKeysDir = "localkeys"
 	// LocalCertificateExtension defines the extension of the certificate files.
@@ -76,6 +74,12 @@ const (
 	LocalKeyExtension = ".key"
 	// TrustStoreDir is the directory name of trust store.
 	TrustStoreDir = "truststore"
+)
+
+// The relative path to {NOTATION_LIBEXEC}
+const (
+	// PathPlugins is the plugins directory relative path.
+	PathPlugins = "plugins"
 )
 
 // The relative path to {NOTATION_CACHE}
--- a/dir/path.go
+++ b/dir/path.go
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 // Package dir implements Notation directory structure.
-// [directory spec]: https://github.com/notaryproject/notation/blob/main/specs/directory.md
+// [directory spec]: https://notaryproject.dev/docs/user-guides/how-to/directory-structure/
 //
 // Example:
 //
@@ -31,7 +31,7 @@
 //   - Set custom configurations directory:
 //     dir.UserConfigDir = '/path/to/configurations/'
 //
-// Only user level directory is supported for RC.1, and system level directory
+// Only user level directory is supported, and system level directory
 // may be added later.
 package dir
 
@@ -44,6 +44,7 @@ import (
 var (
 	UserConfigDir  string // Absolute path of user level {NOTATION_CONFIG}
 	UserLibexecDir string // Absolute path of user level {NOTATION_LIBEXEC}
+	UserCacheDir   string // Absolute path of user level {NOTATION_CACHE}
 )
 
 const (
@@ -77,7 +78,12 @@ const (
 	TrustStoreDir = "truststore"
 )
 
-var userConfigDir = os.UserConfigDir // for unit test
+// for unit tests
+var (
+	userConfigDir = os.UserConfigDir
+
+	userCacheDir = os.UserCacheDir
+)
 
 // userConfigDirPath returns the user level {NOTATION_CONFIG} path.
 func userConfigDirPath() string {
@@ -101,6 +107,21 @@ func userLibexecDirPath() string {
 		UserLibexecDir = userConfigDirPath()
 	}
 	return UserLibexecDir
+}
+
+// userCacheDirPath returns the user level {NOTATION_CACHE} path.
+func userCacheDirPath() string {
+	if UserCacheDir == "" {
+		userDir, err := userCacheDir()
+		if err != nil {
+			// fallback to current directory
+			UserCacheDir = filepath.Join("."+notation, "cache")
+			return UserCacheDir
+		}
+		// set user config
+		UserCacheDir = filepath.Join(userDir, notation)
+	}
+	return UserCacheDir
 }
 
 // LocalKeyPath returns the local key and local cert relative paths.
--- a/dir/path.go
+++ b/dir/path.go
@@ -118,7 +118,7 @@ func userCacheDirPath() string {
 			UserCacheDir = filepath.Join("."+notation, "cache")
 			return UserCacheDir
 		}
-		// set user config
+		// set user cache
 		UserCacheDir = filepath.Join(userDir, notation)
 	}
 	return UserCacheDir
--- a/dir/path.go
+++ b/dir/path.go
@@ -84,8 +84,8 @@ const (
 
 // The relative path to {NOTATION_CACHE}
 const (
-	// PathCRLFileCache is the crl file cache directory relative path.
-	PathCRLFileCache = "crl"
+	// PathCRLCache is the crl file cache directory relative path.
+	PathCRLCache = "crl"
 )
 
 // for unit tests
--- a/dir/path.go
+++ b/dir/path.go
@@ -78,6 +78,12 @@ const (
 	TrustStoreDir = "truststore"
 )
 
+// The relative path to {NOTATION_CACHE}
+const (
+	// PathCRLFileCache is the crl file cache directory relative path.
+	PathCRLFileCache = "crl"
+)
+
 // for unit tests
 var (
 	userConfigDir = os.UserConfigDir
--- a/internal/file/file.go
+++ b/internal/file/file.go
@@ -117,7 +117,7 @@ func TrimFileExtension(fileName string) string {
 	return strings.TrimSuffix(fileName, filepath.Ext(fileName))
 }
 
-// WriteFile writes content to a temporary file and move it to path with all
+// WriteFile writes content to a temporary file and moves it to path with all
 // parent directories created.
 // If path already exists and is a file, WriteFile overwrites it.
 func WriteFile(path string, content []byte) (writeErr error) {
--- a/internal/file/file.go
+++ b/internal/file/file.go
@@ -117,7 +117,8 @@ func TrimFileExtension(fileName string) string {
 	return strings.TrimSuffix(fileName, filepath.Ext(fileName))
 }
 
-// WriteFile writes content to path with all parent directories created.
+// WriteFile writes content to a temporary file and move it to path with all
+// parent directories created.
 // If path already exists and is a file, WriteFile overwrites it.
 func WriteFile(path string, content []byte) (writeErr error) {
 	if err := os.MkdirAll(filepath.Dir(path), 0700); err != nil {
--- a/internal/file/file.go
+++ b/internal/file/file.go
@@ -15,12 +15,18 @@ package file
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
+)
+
+const (
+	// tempFileName is the prefix of the temporary file
+	tempFileName = "notation-*"
 )
 
 // ErrNotRegularFile is returned when the file is not an regular file.
@@ -109,4 +115,35 @@ func CopyDirToDir(src, dst string) error {
 // when input is xyz.tar.gz, output is xyz.tar
 func TrimFileExtension(fileName string) string {
 	return strings.TrimSuffix(fileName, filepath.Ext(fileName))
+}
+
+// WriteFile writes content to path with all parent directories created.
+// If path already exists and is a file, WriteFile overwrites it.
+func WriteFile(path string, content []byte) (writeErr error) {
+	if err := os.MkdirAll(filepath.Dir(path), 0700); err != nil {
+		return err
+	}
+	tempFile, err := os.CreateTemp("", tempFileName)
+	if err != nil {
+		return fmt.Errorf("failed to create temp file: %w", err)
+	}
+	defer func() {
+		// remove the temp file in case of error
+		if writeErr != nil {
+			tempFile.Close()
+			os.Remove(tempFile.Name())
+		}
+	}()
+
+	if _, err := tempFile.Write(content); err != nil {
+		return fmt.Errorf("failed to write content: %w", err)
+	}
+
+	// close before moving
+	if err := tempFile.Close(); err != nil {
+		return fmt.Errorf("failed to close temp file: %w", err)
+	}
+
+	// rename is atomic on UNIX-like platforms
+	return os.Rename(tempFile.Name(), path)
 }
--- a/internal/file/file.go
+++ b/internal/file/file.go
@@ -117,13 +117,9 @@ func TrimFileExtension(fileName string) string {
 	return strings.TrimSuffix(fileName, filepath.Ext(fileName))
 }
 
-// WriteFile writes content to a temporary file and moves it to path with all
-// parent directories created.
+// WriteFile writes content to a temporary file and moves it to path.
 // If path already exists and is a file, WriteFile overwrites it.
 func WriteFile(path string, content []byte) (writeErr error) {
-	if err := os.MkdirAll(filepath.Dir(path), 0700); err != nil {
-		return err
-	}
 	tempFile, err := os.CreateTemp("", tempFileNamePrefix)
 	if err != nil {
 		return fmt.Errorf("failed to create temp file: %w", err)
--- a/internal/file/file.go
+++ b/internal/file/file.go
@@ -136,7 +136,7 @@ func WriteFile(path string, content []byte) (writeErr error) {
 	}()
 
 	if _, err := tempFile.Write(content); err != nil {
-		return fmt.Errorf("failed to write content: %w", err)
+		return fmt.Errorf("failed to write content to temp file: %w", err)
 	}
 
 	// close before moving
--- a/internal/file/file.go
+++ b/internal/file/file.go
@@ -25,8 +25,8 @@ import (
 )
 
 const (
-	// tempFileName is the prefix of the temporary file
-	tempFileName = "notation-*"
+	// tempFileNamePrefix is the prefix of the temporary file
+	tempFileNamePrefix = "notation-*"
 )
 
 // ErrNotRegularFile is returned when the file is not an regular file.
@@ -123,7 +123,7 @@ func WriteFile(path string, content []byte) (writeErr error) {
 	if err := os.MkdirAll(filepath.Dir(path), 0700); err != nil {
 		return err
 	}
-	tempFile, err := os.CreateTemp("", tempFileName)
+	tempFile, err := os.CreateTemp("", tempFileNamePrefix)
 	if err != nil {
 		return fmt.Errorf("failed to create temp file: %w", err)
 	}
--- a/verifier/crl/crl.go
+++ b/verifier/crl/crl.go
@@ -128,7 +128,7 @@ func (c *FileCache) Get(ctx context.Context, url string) (*corecrl.Bundle, error
 }
 
 // Set stores the CRL bundle in c with url as key.
-func (c *FileCache) Set(ctx context.Context, url string, bundle *corecrl.Bundle) error {
+func (c *FileCache) Set(ctx context.Context, url string, bundle *corecrl.Bundle) (setErr error) {
 	logger := log.GetLogger(ctx)
 	logger.Infof("Storing crl bundle to file cache with key %q ...", url)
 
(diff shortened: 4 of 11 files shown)


verifier/crl/crl.go
```diff
@@ -0,0 +1,133 @@
...
+		if os.IsNotExist(err) {
+			logger.Infof("CRL file cache miss. Key %q does not exist", url)
+			return nil, corecrl.ErrCacheMiss
+		}
+		return nil, fmt.Errorf("failed to get crl bundle from file cache with key %q: %w", url, err)
+	}
+	defer f.Close()
+	dec := gob.NewDecoder(f)
+	var bundle corecrl.Bundle
+	err = dec.Decode(&bundle)
```
shizhMSFT, line 86: What if the cache is tempered but the decode is succeeded? or incompatible cache is served (maybe generated by different versions of notation)? Will it cause any future panics?

It is OK that it fails the validation, but it should not panic.

shizhMSFT: Basic validation against [x509.ParseRevocationList()](https://pkg.go.dev/crypto/x509#ParseRevocationList) is necessary.

shizhMSFT (changes requested):

verifier/crl/crl.go
```diff
@@ -0,0 +1,133 @@
...
+	f, err := os.Open(filepath.Join(c.root, c.fileName(url)))
+	if err != nil {
+		if os.IsNotExist(err) {
+			logger.Infof("CRL file cache miss. Key %q does not exist", url)
+			return nil, corecrl.ErrCacheMiss
+		}
+		return nil, fmt.Errorf("failed to get crl bundle from file cache with key %q: %w", url, err)
+	}
+	defer f.Close()
+	dec := gob.NewDecoder(f)
```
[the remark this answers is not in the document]
Two-Hearts: @shizhMSFT okay, perhaps `encoding/json`? Or do you have any other recommendations/suggestions?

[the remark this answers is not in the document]
shizhMSFT: It's up to you as long as it is not golang specific. You can choose JSON, PEM, tar, etc. The point is that do not serialize the full struct. Please only store `BaseCRL.Raw` and `DeltaCRL.Raw` in the file.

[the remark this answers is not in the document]
Two-Hearts: Updated.

verifier/crl/crl.go
Two-Hearts: Updated.

dir/fs.go
```diff
@@ -58,3 +58,8 @@ func ConfigFS() SysFS {
 func PluginFS() SysFS {
 	return NewSysFS(filepath.Join(userLibexecDirPath(), PathPlugins))
 }
+
+// CacheFS is the cache SysFS
+func CacheFS() SysFS {
+	return NewSysFS(userCacheDirPath())
+}
```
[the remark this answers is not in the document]
Two-Hearts: This is intended to be used by callers of notation-go. When caller wants to use `crl.NewFileCache(root string)`, they will need to pass `dir.CacheFS().SysPath()` as `root`.

verifier/verifier.go
```diff
@@ -205,8 +211,13 @@ func (v *verifier) setRevocation(verifierOptions VerifierOptions) error {
...
 	// both RevocationCodeSigningValidator and RevocationClient are nil
+	crlFetcher, err := corecrl.NewHTTPFetcher(&http.Client{Timeout: 5 * time.Second})
+	if err != nil {
+		return err
+	}
 	revocationCodeSigningValidator, err = revocation.NewWithOptions(revocation.Options{
 		OCSPHTTPClient:   &http.Client{Timeout: 2 * time.Second},
+		CRLFetcher:       crlFetcher,
 		CertChainPurpose: purpose.CodeSigning,
 	})
```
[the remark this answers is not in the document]
Two-Hearts: Removed. As discussed, Notation CLI will pass in a `CRLFetcher` with a file cache.

verifier/verifier.go
```diff
@@ -182,8 +183,13 @@ func (v *verifier) setRevocation(verifierOptions VerifierOptions) error {
...
 	var err error
 	if revocationTimestampingValidator == nil {
+		crlFetcher, err := corecrl.NewHTTPFetcher(&http.Client{Timeout: 5 * time.Second})
+		if err != nil {
+			return err
+		}
 		revocationTimestampingValidator, err = revocation.NewWithOptions(revocation.Options{
 			OCSPHTTPClient:   &http.Client{Timeout: 2 * time.Second},
+			CRLFetcher:       crlFetcher,
 			CertChainPurpose: purpose.Timestamping,
```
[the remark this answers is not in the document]
Two-Hearts: Removed. As discussed, Notation CLI will pass in a CRLFetcher with a file cache.

dir/fs.go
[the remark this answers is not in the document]
Two-Hearts: @shizhMSFT You could see the usage in this PR: https://github.com/notaryproject/notation/pull/1043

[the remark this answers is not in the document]
shizhMSFT: I know this is used for caching folder but the cache folder is not just for CRL. What if we need caching for other components such as OCSP (see https://github.com/notaryproject/notation-core-go/issues/230)?

verifier/crl/crl.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// fileCacheContent is the actual content saved in a FileCache
+type fileCacheContent struct {
+	// RawBaseCRL is baseCRL.Raw
+	RawBaseCRL []byte `json:"rawBaseCRL"`
+
+	// RawDeltaCRL is deltaCRL.Raw
+	RawDeltaCRL []byte `json:"rawDeltaCRL,omitempty"`
+}
```
shizhMSFT, line 68: nit: we can simplify it as 
```suggestion
type fileCacheContent struct {
	// BaseCRL is the ASN.1 encoded base CRL
	BaseCRL []byte `json:"baseCRL"`

	// DeltaCRL is the ASN.1 encoded delta CRL
	DeltaCRL []byte `json:"deltaCRL,omitempty"`
}
```

verifier/crl/crl.go
```diff
@@ -0,0 +1,185 @@
...
+	if bundle.DeltaCRL != nil {
+		content.RawDeltaCRL = bundle.DeltaCRL.Raw
+	}
+
+	// save content to tmp file
+	tmpFile, err := os.CreateTemp("", tmpFileName)
+	if err != nil {
+		return fmt.Errorf("failed to store crl bundle in file cache: failed to create temp file: %w", err)
+	}
+	err = json.NewEncoder(tmpFile).Encode(content)
```
shizhMSFT, line 154: `tmpFile` is never closed.

Two-Hearts: Updated.

verifier/crl/crl.go
Two-Hearts: Updated.

dir/fs.go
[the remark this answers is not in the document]
Two-Hearts: Good point. Updated.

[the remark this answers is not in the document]
Two-Hearts: We have some options here:
1. Keep the `CacheFS()`, and when caller wants a crl file cache, they would do `dir.CacheFS().SysPath(dir.PathCRLFileCache)`, where `dir.PathCRLFileCache` is just `crl`.
2. We replace the `CacheFS()` with `CRLFileCacheFS()`, the caller would then use `dir.CRLFileCacheFS().SysPath()` to initialize crl cache. The tradeoff here is if there's another cache in the future, we will need a new type for it.

What do you think? @shizhMSFT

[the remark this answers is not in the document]
shizhMSFT: Option 1 LGTM but we need to document it well.

dir/path.go
```diff
@@ -77,7 +76,24 @@ const (
...
+// The relative path to {NOTATION_LIBEXEC}
+const (
+	// PathPlugins is the plugins directory relative path.
+	PathPlugins = "plugins"
+)
+
+// The relative path to {NOTATION_CACHE}
+const (
+	// PathCRLFileCache is the crl file cache directory relative path.
+	PathCRLFileCache = "crl"
```
shizhMSFT, line 88: Since it is always "file", we may name it
```suggestion
	PathCRLCache = "crl"
```

verifier/crl/crl.go
```diff
@@ -0,0 +1,189 @@
...
+		content.DeltaCRL = bundle.DeltaCRL.Raw
+	}
+
+	// save content to tmp file
+	tmpFile, err := os.CreateTemp("", tmpFileName)
+	if err != nil {
+		return fmt.Errorf("failed to store crl bundle in file cache: failed to create temp file: %w", err)
+	}
+	defer os.Remove(tmpFile.Name())
+	defer tmpFile.Close()
```
shizhMSFT, line 157: Please do proper error handling.
- Check error returned by `Close()` since it is writing path.
- Only remove tmp file if there is an error since the tmp file has been renamed / moved.

dir/fs.go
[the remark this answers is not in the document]
Two-Hearts: Cool. Added comments both for `CacheFS()` and `crl.NewFileCache()` to reflect this.

dir/path.go
Two-Hearts: Updated.

dir/path.go
```diff
@@ -77,7 +76,24 @@ const (
 	TrustStoreDir = "truststore"
 )
 
-var userConfigDir = os.UserConfigDir // for unit test
+// The relative path to {NOTATION_LIBEXEC}
+const (
+	// PathPlugins is the plugins directory relative path.
+	PathPlugins = "plugins"
```
JeyJeyGao, line 82: Why we need to export it?

verifier/crl/crl.go
```diff
@@ -0,0 +1,195 @@
...
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"time"
+
+	corecrl "github.com/notaryproject/notation-core-go/revocation/crl"
```
JeyJeyGao, line 30: Can we name it 'crl' directly?

verifier/crl/crl.go
```diff
@@ -0,0 +1,195 @@
...
+// or the content has expired, corecrl.ErrCacheMiss is returned.
+func (c *FileCache) Get(ctx context.Context, url string) (*corecrl.Bundle, error) {
+	logger := log.GetLogger(ctx)
+	logger.Infof("Retrieving crl bundle from file cache with key %q ...", url)
+
+	// get content from file cache
+	f, err := os.Open(filepath.Join(c.root, c.fileName(url)))
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			logger.Infof("CRL file cache miss. Key %q does not exist", url)
```
JeyJeyGao, line 92: Should it be debug log?

verifier/crl/crl.go
```diff
@@ -0,0 +1,195 @@
...
+			return nil, corecrl.ErrCacheMiss
+		}
+		return nil, fmt.Errorf("failed to get crl bundle from file cache with key %q: %w", url, err)
```
JeyJeyGao, line 95: We may need log the error as core-go will drop it

verifier/crl/crl.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	defer f.Close()
+
+	// decode content to crl Bundle
+	var content fileCacheContent
+	err = json.NewDecoder(f).Decode(&content)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode file retrieved from file cache: %w", err)
```
JeyJeyGao, line 103: All of the error should be logged as core-go will drop it.

verifier/crl/crl.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+	}
+
+	// check expiry
+	if err := checkExpiry(ctx, bundle.BaseCRL.NextUpdate); err != nil {
+		return nil, err
+	}
+	if bundle.DeltaCRL != nil {
+		if err := checkExpiry(ctx, bundle.DeltaCRL.NextUpdate); err != nil {
+			return nil, err
```
JeyJeyGao, line 123: If the delta CRL expires and returns a cache miss, core-go will redownload both the base CRL and delta CRL, which is unnecessary. Therefore, for the cache implementation, they could choose not to handle CRL expiry and let core-go manage it.

verifier/crl/crl.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// Get retrieves CRL bundle from c given url as key. If the key does not exist
+// or the content has expired, corecrl.ErrCacheMiss is returned.
+func (c *FileCache) Get(ctx context.Context, url string) (*corecrl.Bundle, error) {
+	logger := log.GetLogger(ctx)
+	logger.Infof("Retrieving crl bundle from file cache with key %q ...", url)
+
+	// get content from file cache
+	f, err := os.Open(filepath.Join(c.root, c.fileName(url)))
```
shizhMSFT, line 89: nit: Would you consider `os.ReadFile` to simplify the code and release the file lock eariler?

verifier/crl/crl.go
shizhMSFT: `core-go` won't drop it by default

verifier/crl/crl.go
```diff
@@ -0,0 +1,195 @@
...
+		return fmt.Errorf("failed to store crl bundle in file cache: failed to encode content: %w", err)
+	}
+	if err := tempFile.Close(); err != nil {
+		return fmt.Errorf("failed to store crl bundle in file cache: failed to close temp file: %w", err)
+	}
+
+	// rename is atomic on UNIX-like platforms
+	if err := os.Rename(tempFile.Name(), filepath.Join(c.root, c.fileName(url))); err != nil {
+		return fmt.Errorf("failed to store crl bundle in file cache: %w", err)
+	}
```
shizhMSFT, line 173: You may want to refactor this to simplify the code flow. We can chat later offline.

dir/path.go
Two-Hearts: It's not exported by this PR. It was in the wrong place.

verifier/crl/crl.go
Two-Hearts: I'd prefer `corecrl`. This package is already named `crl`. Other packages may need both of them in the future, they will refer core-go crl as `corecrl` anyway.

verifier/crl/crl.go
Two-Hearts: That's not the case though. Remember the discard error flag?

verifier/crl/crl.go
Two-Hearts: https://github.com/notaryproject/notation-core-go/blob/0786f51de7375cfc9a565dc37f66da79cf49f2fe/revocation/crl/cache.go#L24:
`// if the key does not exist or the content is expired, return ErrCacheMiss.`

This is the `Cache` interface of core-go. As an implementation of this interface, we should check expiry.

verifier/crl/crl.go
Two-Hearts: Updated.

verifier/crl/crl.go
```diff
@@ -0,0 +1,193 @@
...
+	root string
+}
+
+// fileCacheContent is the actual content saved in a FileCache
+type fileCacheContent struct {
+	// BaseCRL is the ASN.1 encoded base CRL
+	BaseCRL []byte `json:"baseCRL"`
+
+	// DeltaCRL is the ASN.1 encoded delta CRL
+	DeltaCRL []byte `json:"deltaCRL,omitempty"`
```
priteshbandi, line 67: Should this be array of DeltaCRLs ?

verifier/crl/crl.go
```diff
@@ -0,0 +1,193 @@
...
+	"path/filepath"
+	"time"
+
+	corecrl "github.com/notaryproject/notation-core-go/revocation/crl"
+	"github.com/notaryproject/notation-go/log"
+)
+
+const (
+	// tempFileName is the prefix of the temporary file
+	tempFileName = "notation-*"
```
priteshbandi, line 36: s/tempFileName/tempFileNamePrefix

verifier/crl/crl.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// rename is atomic on UNIX-like platforms
+	if err := os.Rename(tempFile.Name(), filepath.Join(c.root, c.fileName(url))); err != nil {
+		return fmt.Errorf("failed to store crl bundle in file cache: %w", err)
+	}
+	return nil
+}
+
+// fileName returns the filename of the content stored in c
+func (c *FileCache) fileName(url string) string {
```
priteshbandi, line 176: Should we normalize url to remove any white space?

verifier/crl/crl.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+	}
+
+	return &bundle, nil
+}
+
+// Set stores the CRL bundle in c with url as key.
+func (c *FileCache) Set(ctx context.Context, url string, bundle *corecrl.Bundle) (setErr error) {
+	logger := log.GetLogger(ctx)
+	logger.Debugf("Storing crl bundle to file cache with key %q ...", url)
```
priteshbandi, line 131: nit: url is not the key; its hash of url

verifier/crl/crl.go
```diff
@@ -0,0 +1,193 @@
...
+	return &FileCache{
+		root: root,
+	}, nil
+}
+
+// Get retrieves CRL bundle from c given url as key. If the key does not exist
+// or the content has expired, corecrl.ErrCacheMiss is returned.
+func (c *FileCache) Get(ctx context.Context, url string) (*corecrl.Bundle, error) {
+	logger := log.GetLogger(ctx)
+	logger.Debugf("Retrieving crl bundle from file cache with key %q ...", url)
```
priteshbandi, line 86: nit: url is not the key

verifier/crl/crl.go
```diff
@@ -0,0 +1,193 @@
...
+	var content fileCacheContent
+	if err := json.Unmarshal(contentBytes, &content); err != nil {
+		return nil, fmt.Errorf("failed to decode file retrieved from file cache: %w", err)
+	}
+	var bundle corecrl.Bundle
+	bundle.BaseCRL, err = x509.ParseRevocationList(content.BaseCRL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse base CRL of file retrieved from file cache: %w", err)
+	}
+	if content.DeltaCRL != nil {
```
priteshbandi, line 108: Should we validate that baseCRL doesn't needs delta CRL here?

verifier/crl/crl.go
Two-Hearts: It corresponds to a single [x509.RevocationList](https://pkg.go.dev/crypto/x509#RevocationList).

verifier/crl/crl.go
Two-Hearts: Since this is the `key` of the cache, normalizing it would change its sha256 hash value. I think we should keep what the user passed in? (just in case, there might be other ways to retrieve the cache value from file system in the future).

verifier/crl/crl.go
Two-Hearts: This is from the definition of `Cache` interface: https://github.com/notaryproject/notation-core-go/blob/e90546bd90a8074357dba2cbf19f2e755bd4be2a/revocation/crl/cache.go#L19. 

From the user's point of view, url is the `key` to the cache. They would use the same url as `key` to fetch content from the cache.
As for how notation internally stores the key and the content is implementation details and should be hidden from external caller. This is intended, otherwise, the user might compute the url sha256 themselves and pass in the hash value.

verifier/crl/crl.go
Two-Hearts: Tracking the discussion above: https://github.com/notaryproject/notation-go/pull/462#discussion_r1772749429

verifier/crl/crl.go
Two-Hearts: I'm not sure if this should be a part of the cache implementation? It returns the bundle stored in the cache, the caller gets decide whether to use the full bundle or not. Notation-core-go is only using the `baseCRL` at the moment.

verifier/crl/crl.go
Two-Hearts: Updated.

verifier/crl/crl.go
Two-Hearts: Updated after discussion.

verifier/crl/crl.go
Two-Hearts: @priteshbandi I created an issue to track questions regarding `deltaCRL`: https://github.com/notaryproject/notation-go/issues/463. We will revisit its logic once notation-core-go supports it.

verifier/crl/crl.go
Two-Hearts: Tracked by issue: https://github.com/notaryproject/notation-go/issues/463

verifier/crl/crl.go
Two-Hearts: Added `url` validation to make sure a valid url is passed by caller.

verifier/crl/crl.go
JeyJeyGao: Will we use the default option? If so, the cache error will cause the operation to fail.

verifier/crl/crl_test.go
```diff
@@ -0,0 +1,372 @@
...
+		t.Fatal(err)
+	}
+	baseCRL, err := x509.ParseRevocationList(crlBytes)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ctx := context.Background()
+	root := t.TempDir()
+	cache, err := NewFileCache(root)
```
JeyJeyGao, line 52: var cache corecrl.Cache

Can we add it to enforce the FileCache to implement Cache interface?

internal/file/file.go
```diff
@@ -110,3 +116,34 @@ func CopyDirToDir(src, dst string) error {
 func TrimFileExtension(fileName string) string {
 	return strings.TrimSuffix(fileName, filepath.Ext(fileName))
 }
+
+// WriteFile writes content to path with all parent directories created.
```
shizhMSFT, line 120: nit:
```suggestion
// WriteFile writes content to a temporary file and move it to path with all parent directories created.
```

verifier/crl/crl.go
shizhMSFT: @priteshbandi Could you elaborate more on "array of DeltaCRLs"?

internal/file/file.go
```diff
@@ -110,3 +116,34 @@ func CopyDirToDir(src, dst string) error {
...
+// If path already exists and is a file, WriteFile overwrites it.
+func WriteFile(path string, content []byte) (writeErr error) {
+	if err := os.MkdirAll(filepath.Dir(path), 0700); err != nil {
+		return err
+	}
```
shizhMSFT, line 125: It seems not necessary to call `MkdirAll` each time `WriteFile`, especially when you have this logic at `crl.go:71`.

verifier/crl/crl.go
```diff
@@ -0,0 +1,180 @@
...
+import (
+	"context"
+	"crypto/sha256"
+	"crypto/x509"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/fs"
+	nurl "net/url"
```
shizhMSFT, line 26: Why is this package aliased to `nurl`? this should be reverted for readability.

shizhMSFT: You can keep the package `url` as `url` and rename the `url` variable name to other names like `crlURL` per context.

verifier/crl/crl.go
```diff
@@ -0,0 +1,180 @@
...
+// Get retrieves CRL bundle from c given url as key. If the key does not exist
+// or the content has expired, corecrl.ErrCacheMiss is returned.
+func (c *FileCache) Get(ctx context.Context, url string) (*corecrl.Bundle, error) {
+	logger := log.GetLogger(ctx)
+	logger.Debugf("Retrieving crl bundle from file cache with key %q ...", url)
+
+	// sanity check
+	if _, err := nurl.Parse(url); err != nil {
+		return nil, fmt.Errorf("invalid url: %w", err)
+	}
```
shizhMSFT, line 88: This check is unnecessary.

shizhMSFT: `url` is just a cache key.

shizhMSFT: bad url should just be a cache miss

verifier/crl/crl.go
shizhMSFT: I agree that validating delta CRL against base CRL should not be in the scope of caching.

verifier/crl/crl.go
```diff
@@ -0,0 +1,180 @@
...
+
+// Set stores the CRL bundle in c with url as key.
+func (c *FileCache) Set(ctx context.Context, url string, bundle *corecrl.Bundle) error {
+	logger := log.GetLogger(ctx)
+	logger.Debugf("Storing crl bundle to file cache with key %q ...", url)
+
+	// sanity check
+	if _, err := nurl.Parse(url); err != nil {
+		return fmt.Errorf("invalid url: %w", err)
+	}
```
shizhMSFT, line 137: This check is not necessary

internal/file/file.go
Two-Hearts: This method is under internal/file at the moment, which is shared by packages. The [os.WriteFile](https://pkg.go.dev/os#WriteFile) has the similar behavior though. 
If we remove `MkdirAll`, caller has to create the dir before writing every time.
An alternative would be moving this function back to `crl` package. What do you think? @shizhMSFT

internal/file/file.go
Two-Hearts: Updated.

verifier/crl/crl.go
Two-Hearts: removed.

verifier/crl/crl.go
Two-Hearts: removed.

verifier/crl/crl.go
Two-Hearts: Yes, we will use the default option.

verifier/crl/crl.go
```diff
@@ -0,0 +1,180 @@
...
+	if bundle == nil {
+		return errors.New("failed to store crl bundle in file cache: bundle cannot be nil")
```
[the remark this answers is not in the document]
Two-Hearts: What if other library uses this cache? Will core-go be the only caller of this cache implementation?

internal/file/file.go
```diff
@@ -110,3 +116,34 @@ func CopyDirToDir(src, dst string) error {
...
+	tempFile, err := os.CreateTemp("", tempFileNamePrefix)
+	if err != nil {
+		return fmt.Errorf("failed to create temp file: %w", err)
+	}
+	defer func() {
+		// remove the temp file in case of error
+		if writeErr != nil {
```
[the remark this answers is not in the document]
Two-Hearts: This is indeed subtle. Please see the returned value of this function, any error occurred will be `writeErr`.

verifier/crl/crl_test.go
Two-Hearts: This won't work, because the test cases below are using `cache.root` of the `cache` struct. Let me add a new test case to enforce the implementation of the interface.

Two-Hearts: Added a new test regarding this.

internal/file/file.go
Two-Hearts: Updated.

shizhMSFT (approved):
LGTM

JeyJeyGao (approved):

priteshbandi (approved):
LGTM

PR #471, opened 2024-10-29
fix: added tsa trust store root cert validation

### LINKED ISSUE #470: TSA trust store root certificate validation
We currently require tsa trust store to only store TSA's root certificates. Although it is checked during tsa counter signature verification, we could fail fast at the step of getting certificates from the trust store.

Two-Hearts:
This PR adds tsa trust store root cert validation while getting certificates from trust store. This is to fail fast if cert in TSA trust store is not a root CA certificate.

Resolves #470


shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

PR #475, opened 2024-11-08
fix: crl cache log and err msg

Two-Hearts:
This PR removes redundant logs in crl cache `Get` and `Set` methods. Instead, these logs should be logged in Notation CLI, see https://github.com/notaryproject/notation/pull/1078 for reference.

This PR also fixes a couple of error messages in crl cache.


JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

PR #478, opened 2024-11-12
fix: timestamping

Two-Hearts:
This PR adds a sanity check on timstamp value against the signing time: timestamp value should always be bounded after the signing time. 
This is to say, one cannot timestamp a signature before the signature itself been created. If it happens, fail the verification.


shizhMSFT (approved):
LGTM

JeyJeyGao (approved):
LGTM

priteshbandi:
Do we need to account for clock skew such as when the local system clock runs faster than the timestamp server??

Two-Hearts:
> Do we need to account for clock skew such as when the local system clock runs faster than the timestamp server??

This change has been tested with several public tsa servers, such as digicert and globalsign. 
Since we cannot tell the difference between a clock skew and an intentionally manipulated local system clock, we should fail the verification on both cases following the `secure by default` practice.

In addition, this check also covers the scenario where the signer generates a Notary Project compliant signature using another tool and send it to tsa for timestamping `later on`, then attach the signature themselves. When verifying such signature envelopes in Notation, we expect the tsa timestamp to be later than the signing time as well.
@priteshbandi

priteshbandi (approved):
LGTM

PR #479, opened 2024-11-21
fix: add warning message for non-revokable certificate

JeyJeyGao:
Fix:
- added warning message for non-revokable certificate

### DIFF
--- a/verifier/verifier.go
+++ b/verifier/verifier.go
@@ -840,7 +840,7 @@ func revocationFinalResult(certResults []*revocationresult.CertRevocationResult,
 		}
 
 		if i < len(certResults)-1 && certResult.Result == revocationresult.ResultNonRevokable {
-			logger.Debugf("Certificate #%d in the chain with subject %v neither has an OCSP nor a CRL revocation method.", (i + 1), cert.Subject.String())
+			logger.Warnf("Certificate #%d in the chain with subject %v neither has an OCSP nor a CRL revocation method.", (i + 1), cert.Subject.String())
 		}
 	}
 	if revokedFound {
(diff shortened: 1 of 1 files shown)


verifier/verifier.go
```diff
@@ -838,6 +838,10 @@ func revocationFinalResult(certResults []*revocationresult.CertRevocationResult,
 				revokedCertSubject = problematicCertSubject
 			}
 		}
+
+		if i < len(certResults)-1 && certResult.Result == revocationresult.ResultNonRevokable {
+			logger.Warnf("Certificate #%d in the chain with subject %v neither has an OCSP nor a CRL revocation method.", (i + 1), cert.Subject.String())
```
Two-Hearts, line 843: We should only display the warning under `strict` mode?

Two-Hearts: In other words, I don't see a particular reason to set it as a warning under `log` mode.

shizhMSFT (approved):
LGTM

Two-Hearts (approved):
LGTM

JeyJeyGao:
#480
Added an issue to improve the performance of error message.

verifier/verifier.go
priteshbandi: Alternative methods exist for verifying revocation status, such as through the use of plugins. Therefore, displaying this warning may be inaccurate.

JeyJeyGao: https://github.com/notaryproject/notation-go/blob/5a323330d0392dcf703930c1ff4b46a6197ccdc0/verifier/verifier.go#L505-L518

If the plugin has `CapabilityRevocationCheckVerifier`, it will call the plugin instead of triggering this code.

PR #482, opened 2024-11-29
fix: enable timestamping cert chain revocation check during signing

### LINKED ISSUE #483: TSA certificate chain revocation check after timestamping
We should consider adding an option for users to enable tsa cert chain revocation check after timestamping.
Currently this is done during verification only, but we can fail fast on the signing side.

Two-Hearts:
This PR adds an option to enable tsa cert chain revocation check during signing.

Resolves #483

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -16,3 +16,6 @@ coverage:
     project:
       default:
         target: 80%
+    patch:
+      default:
+        target: 80%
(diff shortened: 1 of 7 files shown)


JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

PR #484, opened 2024-11-29
fix: limit the plugin output size

### LINKED ISSUE #187: plugin output size limitation
The plugin is an external binary which may generate very large command line output and will be loaded to memory by notation-go library. 
Do we need a plugin output size limitation?

JeyJeyGao:
Fix:
- set the plugin output limit for STDOUT and STDERR to be 10MiB

Test:
- when the plugin output size exceeds 64MiB, the output pipe breaks, and the plugin process outputs an error to STDERR

Spec changes: https://github.com/notaryproject/specifications/pull/320
Resolves #187

### DIFF
--- a/internal/io/limitedwriter.go
+++ b/internal/io/limitedwriter.go
@@ -1,0 +1,53 @@
+// Copyright The Notary Project Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package io provides a LimitWriter that writes to an underlying writer up to
+// a limit.
+
+package io
+
+import (
+	"errors"
+	"io"
+)
+
+// ErrLimitExceeded is returned when the write limit is exceeded.
+var ErrLimitExceeded = errors.New("write limit exceeded")
+
+// LimitedWriter is a writer that writes to an underlying writer up to a limit.
+type LimitedWriter struct {
+	W io.Writer // underlying writer
+	N int64     // remaining bytes
+}
+
+// LimitWriter returns a new LimitWriter that writes to w.
+//
+// parameters:
+// w: the writer to write to
+// limit: the maximum number of bytes to write
+func LimitWriter(w io.Writer, limit int64) *LimitedWriter {
+	return &LimitedWriter{W: w, N: limit}
+}
+
+// Write writes p to the underlying writer up to the limit.
+func (l *LimitedWriter) Write(p []byte) (int, error) {
+	if l.N <= 0 {
+		return 0, ErrLimitExceeded
+	}
+	if int64(len(p)) > l.N {
+		p = p[:l.N]
+	}
+	n, err := l.W.Write(p)
+	l.N -= int64(n)
+	return n, err
+}
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -27,11 +27,15 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/notaryproject/notation-go/internal/io"
 	"github.com/notaryproject/notation-go/internal/slices"
 	"github.com/notaryproject/notation-go/log"
 	"github.com/notaryproject/notation-go/plugin/proto"
 	"github.com/notaryproject/notation-plugin-framework-go/plugin"
 )
+
+// maxPluginOutputSize is the maximum size of the plugin output.
+const maxPluginOutputSize = 64 * 1024 * 1024 // 64 MiB
 
 var executor commander = &execCommander{} // for unit test
 
@@ -218,12 +222,14 @@ func (c execCommander) Output(ctx context.Context, name string, command plugin.C
 	var stdout, stderr bytes.Buffer
 	cmd := exec.CommandContext(ctx, name, string(command))
 	cmd.Stdin = bytes.NewReader(req)
-	cmd.Stderr = &stderr
-	cmd.Stdout = &stdout
+	// The limit writer will be handled by the caller in run() by comparing the
+	// bytes written with the expected length of the bytes.
+	cmd.Stderr = io.LimitWriter(&stderr, maxPluginOutputSize)
+	cmd.Stdout = io.LimitWriter(&stdout, maxPluginOutputSize)
 	err := cmd.Run()
 	if err != nil {
 		if errors.Is(ctx.Err(), context.DeadlineExceeded) {
-			return nil, stderr.Bytes(), fmt.Errorf("'%s %s' command execution timeout: %w", name, string(command), err);
+			return nil, stderr.Bytes(), fmt.Errorf("'%s %s' command execution timeout: %w", name, string(command), err)
 		}
 		return nil, stderr.Bytes(), err
 	}
--- a/internal/io/limitedwriter_test.go
+++ b/internal/io/limitedwriter_test.go
@@ -1,0 +1,67 @@
+// Copyright The Notary Project Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package io
+
+import (
+	"bytes"
+	"errors"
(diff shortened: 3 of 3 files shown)


internal/io/limitwriter.go
```diff
@@ -0,0 +1,49 @@
...
+
+// Package io provides a LimitWriter that writes to an underlying writer up to
+// a limit.
+
+package io
+
+import "io"
+
+// LimitWriter is a writer that writes to an underlying writer up to a limit.
+type LimitWriter struct {
```
shizhMSFT, line 22: Please follow the naming convention of the `io` package. That is, `LimitedWriter` and `LimitWriter()` just like [io.LimitedReader](https://pkg.go.dev/io#LimitedReader) and [io.LimitReader](https://pkg.go.dev/io#LimitReader).

shizhMSFT: You should also follow the similar fields.

internal/io/limitwriter.go
```diff
@@ -0,0 +1,49 @@
...
+// limit: the maximum number of bytes to write
+func NewLimitWriter(w io.Writer, limit int64) *LimitWriter {
+	return &LimitWriter{w: w, limit: limit}
+}
+
+// Write writes p to the underlying writer up to the limit.
+func (lw *LimitWriter) Write(p []byte) (int, error) {
+	remaining := lw.limit - lw.written
+	if remaining <= 0 {
+		return 0, nil
```
shizhMSFT, line 41: This should return an error. `ErrShortWrite` might be a proper error.

Two-Hearts:
I actually have a couple of questions regarding this PR and the issue #187 itself: 
1. We are essentially limiting the behavior of a plugin due to a certain level of `untrust`. Do we need to reflect this in the Notary project spec? (such as, where does the 10MiB comes from?)
2. What security concern do we have if the plugin output size is larger than 10 MiB? According to Notation threat model: https://github.com/notaryproject/specifications/blob/main/threatmodels/notation-threatmodel.md, plugins are inside the `trust boundary`. If we do not trust the plugin's output, how do we interpret the trust boundary then?

Two-Hearts (changes requested):
I actually have a couple of questions regarding this PR and the issue https://github.com/notaryproject/notation-go/issues/187 itself:

1. We are essentially limiting the behavior of a plugin due to a certain level of untrust. Do we need to reflect this in the Notary project spec? (such as, where does the 10MiB comes from?)
2. What security concern do we have if the plugin output size is larger than 10 MiB? According to Notation threat model: https://github.com/notaryproject/specifications/blob/main/threatmodels/notation-threatmodel.md, plugins are inside the trust boundary. If we do not trust the plugin's output, how do we interpret the trust boundary then?

internal/io/limitwriter.go
JeyJeyGao: Updated.

internal/io/limitwriter.go
JeyJeyGao: Updated.

JeyJeyGao:
> I actually have a couple of questions regarding this PR and the issue #187 itself:
> 
> 1. We are essentially limiting the behavior of a plugin due to a certain extent of `untrust`. Do we need to reflect this in the Notary project spec? (such as, where does the 10MiB comes from?)
> 2. What security concern do we have if the plugin output size is larger than 10 MiB? According to Notation threat model: https://github.com/notaryproject/specifications/blob/main/threatmodels/notation-threatmodel.md, plugins are inside the trust boundary. If we do not trust the plugin's output, how do we interpret the trust boundary then?

1. The limit has been updated to 64 MiB, which is almost impossible to be reached by a normal plugin. The limitation depends on the implementation; different implementations may have different concerns that lead to different limitation decisions. The limitation is more like a notation-specific setting rather than a specification requirement. If we need to update the specification, we can mention that the plugin output may have size limitations depending on the implementation.
2. We don’t have security concerns regarding this issue. The goal is to improve system reliability when a plugin has a bug or generates unexpected output. Notation-go can generate proper logs without killing the process directly, ensuring proper handling.

Two-Hearts:
> 1. The limit has been updated to 64 MiB, which is almost impossible to be reached by a normal plugin. The limitation depends on the implementation; different implementations may have different concerns that lead to different limitation decisions. The limitation is more like a notation-specific setting rather than a specification requirement. If we need to update the specification, we can mention that the plugin output may have size limitations depending on the implementation.

Note: any extra restriction on the plugin without reflecting in the spec may lead to potential breaking changes for Notation.

> 2. We don’t have security concerns regarding this issue. The goal is to improve system reliability when a plugin has a bug or generates unexpected output. Notation-go can generate proper logs without killing the process directly, ensuring proper handling.

If we indeed do not have any security concern regarding this issue, then we should leave the control to plugin authors and users. If a plugin has a bug, its usage should be dropped until the bug is fixed right?  I don't think this is in Notation's scope, just like we do not restrict the output content of a plugin.

JeyJeyGao:
> > 1. The limit has been updated to 64 MiB, which is almost impossible to be reached by a normal plugin. The limitation depends on the implementation; different implementations may have different concerns that lead to different limitation decisions. The limitation is more like a notation-specific setting rather than a specification requirement. If we need to update the specification, we can mention that the plugin output may have size limitations depending on the implementation.
> 
> Note: any extra restriction on the plugin without reflecting in the spec may lead to potential breaking changes for Notation. It's not about the number, whether it's 10 or 64. It's about best practice. The reason is simple, plugin builders follow the spec: https://github.com/notaryproject/specifications/blob/main/specs/plugin-extensibility.md to build their plugins. Any extra restriction that's not in that spec can be a hidden rule and unexpected.
> 
> > 2. We don’t have security concerns regarding this issue. The goal is to improve system reliability when a plugin has a bug or generates unexpected output. Notation-go can generate proper logs without killing the process directly, ensuring proper handling.
> 
> If we indeed do not have any security concern regarding this issue, then we should leave the control to plugin authors and users. If a plugin has a bug, its usage should be dropped until the bug is fixed right? I don't think this is in Notation's scope, just like we do not censor the output content of a plugin.

As per our online discussion, I created a corresponding Spec PR: https://github.com/notaryproject/specifications/pull/320

internal/io/limitedwriter.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// Write writes p to the underlying writer up to the limit.
+func (l *LimitedWriter) Write(p []byte) (int, error) {
+	if l.N <= 0 {
+		return 0, io.ErrShortWrite
+	}
+	if int64(len(p)) > l.N {
+		p = p[:l.N]
+	}
```
Two-Hearts, line 43: I'm trying to understand, when p is larger than the limit, we are truncating p up to the limit? Should we return an error in this case?

JeyJeyGao: https://cs.opensource.google/go/go/+/refs/tags/go1.23.3:src/io/io.go;l=472-482
Please have a look at the LimitedReader implementation, which is similar to this case.

By calling
> n, err := l.W.Write(p)

the value of n will be less than expected, and the caller will know there is an issue.

internal/io/limitedwriter.go
```diff
@@ -0,0 +1,47 @@
...
+// w: the writer to write to
+// limit: the maximum number of bytes to write
+func LimitWriter(w io.Writer, limit int64) *LimitedWriter {
+	return &LimitedWriter{W: w, N: limit}
+}
+
+// Write writes p to the underlying writer up to the limit.
+func (l *LimitedWriter) Write(p []byte) (int, error) {
+	if l.N <= 0 {
+		return 0, io.ErrShortWrite
```
Two-Hearts, line 39: The error `io.ErrShortWrite` is defined as `a write accepted fewer bytes than requested but failed to return an explicit error.`
But in a `Limit writer` context, at line 39, it means the limit of this writer has been reached. Therefore, the error message should be something like:
`limit %d of limit writer has been reached`

plugin/plugin.go
```diff
@@ -218,12 +222,12 @@ func (c execCommander) Output(ctx context.Context, name string, command plugin.C
 	var stdout, stderr bytes.Buffer
 	cmd := exec.CommandContext(ctx, name, string(command))
 	cmd.Stdin = bytes.NewReader(req)
-	cmd.Stderr = &stderr
-	cmd.Stdout = &stdout
+	cmd.Stderr = io.LimitWriter(&stderr, maxPluginOutputSize)
+	cmd.Stdout = io.LimitWriter(&stdout, maxPluginOutputSize)
 	err := cmd.Run()
```
Two-Hearts, line 227: Now I understand how we trigger the limit writer limit check. Please add some comments in the code to illustrate it as it's buried deeply under `cmd.Run()`.

internal/io/limitedwriter.go
```diff
@@ -0,0 +1,52 @@
...
+// a limit.
+
+package io
+
+import (
+	"errors"
+	"io"
+)
+
+var ErrLimitExceeded = errors.New("write limit exceeded")
```
shizhMSFT, line 24: Request for doc

shizhMSFT (approved):
LGTM with suggestions

Two-Hearts (approved):
LGTM

internal/io/limitedwriter.go
JeyJeyGao: Updated.

shizhMSFT (approved):
LGTM

Two-Hearts (approved):
LGTM

priteshbandi (approved):
LGTM

PR #502, opened 2025-01-09
chore: clean up

Two-Hearts:
While using the library, it's necessary for a further clean up:
1. Fix comments with grammar issue.
2. Fix 1 error message in blob trust policy.
3. Removing/adding empty lines based on golang convention.


JeyJeyGao (approved):
LGTM

Issue #513, opened 2025-02-20
Timestamp verification against signing time


Two-Hearts:
In Notation v1.3.0, we added a new check during verification, which requires the timestamp be bounded after the signing time. This may cause timestamp verification failure due to time skew.
We could remove this new check for now and revisit the necessity of this logic in further discussions.

priteshbandi:
Do we want to show warning in this case?

Two-Hearts:
> Do we want to show warning in this case?

Yeah, add a warning could be another option. We could revisit this check after the patch release.

Two-Hearts:
> Do we want to show warning in this case?

Yeah, add a warning could be another option. We could revisit this check after the v1.3.1 patch release.

PR #514, opened 2025-02-20
fix: remove signing time check during authenticTimestamp

Two-Hearts:
This PR removes the signing time check against timestamp during authentic timestamp verification. The possibility of time skew requires further discussion on the check.

Resolves: #513


shizhMSFT (approved):
LGTM

shizhMSFT:
This PR reverts #478

JeyJeyGao (approved):
LGTM

PR #518, opened 2025-02-21
fix: timestamp against signing time

Two-Hearts:
This PR removes the timestamping check against signing time during authenticTimestamp verification. This is due to potential time skew and signing time is considered unauthenticated per Notary Project spec: https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#standard-attributes.


JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

shizhMSFT:
This PR reverts #478

PR #527, opened 2025-03-24
fix: update error message for signing key config

Two-Hearts:
This PR contains some error message enhancements found while using the library.

### DIFF
--- a/config/errors.go
+++ b/config/errors.go
@@ -18,16 +18,16 @@ import (
 	"fmt"
 )
 
-// ErrorKeyNameEmpty is used when key name is empty.
-var ErrorKeyNameEmpty = errors.New("key name cannot be empty")
+// ErrKeyNameEmpty is used when key name is empty.
+var ErrKeyNameEmpty = errors.New("key name cannot be empty")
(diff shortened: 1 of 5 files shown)


config/errors.go
```diff
@@ -0,0 +1,35 @@
...
+	KeyName string
+}
+
+// Error returns the error message.
+func (e ErrorKeyNotFound) Error() string {
+	if e.KeyName != "" {
+		return fmt.Sprintf("signing key %s not found", e.KeyName)
+	}
+	return "signing key not found"
+}
```
shizhMSFT, line 35: Since they are exported, can we follow the golang naming conversation that.
- `ErrFoo` if the error is a variable.
- `BarError` if the error is a type.

shizhMSFT (changes requested):

config/errors.go
Two-Hearts: Updated.

shizhMSFT (approved):
LGTM

JeyJeyGao (approved):
LGTM

Issue #528, opened 2025-03-24
Error type needs to follow Golang Convention


Two-Hearts:
Currently, error types defined in errors.go is not following the Golang naming convention.

For example, `ErrorPushSignatureFailed` should be named as `PushSignatureFailedError`

Reference: https://go.dev/wiki/Errors#:~:text=PrintParseError(err)%20%7D-,Naming,int%20%7D%20var%20ErrBadAction%20%3D%20errors.

qba73:
@Two-Hearts would you accept a PR to address this issue?

Two-Hearts:
> [@Two-Hearts](https://github.com/Two-Hearts) would you accept a PR to address this issue?

@qba73 Sure! Thanks!
(Note: since these errors are exported Golang types, directly modifying them is considered as a breaking change. So a PR that deprecates them and adds new error types would be a better way at the moment.)

FeynmanZhou:
Hi @qba73 , thanks for your interest! Just assigned this issue to you. Looking forward to your first PR to address it. You are also welcome to join the [slack channel](https://cloud-native.slack.com/archives/CQUH8U287) to connect with the community.

qba73:
@FeynmanZhou PR is ready for review

PR #529, opened 2025-04-02
refactor: update error names to align with Go convention

qba73:
This PR fixes https://github.com/notaryproject/notation-go/issues/528 and adds the following changes:

- add new custom errors (naming aligned with [Go convention](https://go.dev/doc/effective_go#errors))
- mark old custom error deprecated

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -15,6 +15,8 @@ package notation
 
 // ErrorPushSignatureFailed is used when failed to push signature to the
 // target registry.
+//
+// Deprecated: Use PushSignatureFailedError instead.
 type ErrorPushSignatureFailed struct {
 	Msg string
 }
@@ -26,8 +28,23 @@ func (e ErrorPushSignatureFailed) Error() string {
 	return "failed to push signature to registry"
 }
 
+// PushSignatureFailedError is used when failed to push signature to the
+// target registry.
+type PushSignatureFailedError struct {
+	Msg string
+}
+
+func (e PushSignatureFailedError) Error() string {
+	if e.Msg != "" {
+		return "failed to push signature to registry with error: " + e.Msg
+	}
+	return "failed to push signature to registry"
+}
+
 // ErrorVerificationInconclusive is used when signature verification fails due
 // to a runtime error (e.g. a network error)
+//
+// Deprecated: Use VerificationInconclusiveError instead.
 type ErrorVerificationInconclusive struct {
 	Msg string
 }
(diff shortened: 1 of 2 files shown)


FeynmanZhou:
@qba73 Thanks for your contribution! Would you mind signing off your git commit? See the details at https://github.com/notaryproject/notation-go/pull/529/checks?check_run_id=39824512392

errors.go
```diff
@@ -15,6 +15,8 @@ package notation
 
 // ErrorPushSignatureFailed is used when failed to push signature to the
 // target registry.
+//
+// Deprecated: Use PushSignatureFailedError instead.
 type ErrorPushSignatureFailed struct {
```
shizhMSFT, line 20: Since `ErrorPushSignatureFailed` and `PushSignatureFailedError` are the exact same thing, can you do
```
type ErrorPushSignatureFailed = PushSignatureFailedError
```
The same comment applies to other types.

shizhMSFT (changes requested):

errors.go
qba73: > Since `ErrorPushSignatureFailed` and `PushSignatureFailedError` are the exact same thing, can you do
> 
> ```
> type ErrorPushSignatureFailed = PushSignatureFailedError
> ```
> 
> The same comment applies to other types.

@shizhMSFT updated, please let me know if it looks like you want

qba73:
> @qba73 Thanks for your contribution! Would you mind signing off your git commit? See the details at https://github.com/notaryproject/notation-go/pull/529/checks?check_run_id=39824512392

Done

errors.go
shizhMSFT: LoL. I mean you can do something like
```go
// ErrorPushSignatureFailed is used when failed to push signature to the
// target registry.
//
// Deprecated: Use PushSignatureFailedError instead.
type ErrorPushSignatureFailed = PushSignatureFailedError

// PushSignatureFailedError is used when failed to push signature to the
// target registry.
type PushSignatureFailedError struct {
	Msg string
}

func (e PushSignatureFailedError) Error() string {
	if e.Msg != "" {
		return "failed to push signature to registry with error: " + e.Msg
	}
	return "failed to push signature to registry"
}
```
instead of putting it in the comment block.

qba73: done

Two-Hearts (approved):
LGTM

JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #542, opened 2025-04-27
feat: `artifactType` support in signature manifest

### LINKED ISSUE #540: Add support of artifactType under signature manifest
Add support of `artifactType` field under signature manifest as defined by OCI image spec v1.1: https://github.com/opencontainers/image-spec/blob/v1.1.0/manifest.md

Two-Hearts:
This PR is an implementation of the Notary Project spec https://github.com/notaryproject/specifications/pull/325.

Resolves #540


JeyJeyGao (dismissed):
LGTM

registry/repository.go
```diff
@@ -274,19 +231,35 @@ func signatureReferrers(ctx context.Context, target content.ReadOnlyGraphStorage
...
+				node.ArtifactType = image.ArtifactType
+			} else if image.ArtifactType == "" && image.Config.MediaType == ArtifactTypeNotation {
+				// 2. artifacteType does not exist,
+				// and config.mediaType is "application/vnd.cncf.notary.signature"
+				node.ArtifactType = image.Config.MediaType
+			} else {
+				// not a valid Notary Project signature
+				logger.Infof("not a valid Notary Project signature with artifactType %q and config.mediaType %q", image.ArtifactType, image.Config.MediaType)
+				continue
+			}
```
shizhMSFT, line 256: It is better to have a switch statement on `image.ArtifactType`.

Note: `node` is not always a Notary Project signature. It can be a SBoM or other artifacts. We should not have info log for those artifacts.

Two-Hearts: Updated.

registry/repository.go
```diff
@@ -274,19 +231,46 @@ func signatureReferrers(ctx context.Context, target content.ReadOnlyGraphStorage
...
 			node.Annotations = image.Annotations
 		default:
 			continue
 		}
-		// only keep nodes of "application/vnd.cncf.notary.signature"
+
+		// only keeping nodes of "application/vnd.cncf.notary.signature"
 		if node.ArtifactType == ArtifactTypeNotation {
```
shizhMSFT, line 274: I noticed that the nodes are actually filtered by line 248 and line 256 in your changed code. Do you still need this check?

Two-Hearts: This check is for both `application/vnd.oci.artifact.manifest.v1+json` and `application/vnd.oci.image.manifest.v1+json`. 
Updated by adding the check to `application/vnd.oci.artifact.manifest.v1+json` as well.

JeyJeyGao (dismissed):
LGTM

JeyJeyGao (dismissed):
LGTM

registry/repository.go
```diff
@@ -274,22 +235,47 @@ func signatureReferrers(ctx context.Context, target content.ReadOnlyGraphStorage
...
+			// check if image is a valid Notary Project signature
+			switch image.ArtifactType {
+			case ArtifactTypeNotation:
+				// 1. artifactType is "application/vnd.cncf.notary.signature",
+				// and config.mediaType is "application/vnd.oci.empty.v1+json"
+				if image.Config.MediaType == ocispec.MediaTypeEmptyJSON {
+					node.ArtifactType = image.ArtifactType
+				} else {
+					// not a valid Notary Project signature
+					logger.Infof("not a valid Notary Project signature with artifactType %q, but config.mediaType is %q", image.ArtifactType, image.Config.MediaType)
```
shizhMSFT, line 256: nit: Should it be a warning?

shizhMSFT (approved):
LGTM with suggestions

registry/repository.go
Two-Hearts: Yeah, it should be a warning.

Two-Hearts: Updated.

JeyJeyGao (approved):
LGTM

PR #545, opened 2025-05-07
add govulncheck workflow

qba73:
This PR adds [govulncheck](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) workflow to run security audits.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,16 @@
+name: Security audit
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/audit.yml
JeyJeyGao, line 1: Maybe naming it `govulncheck.yml` will make it clearer?

qba73: applied

JeyJeyGao (approved):
LGTM

PR #572, opened 2026-03-12
feat: add NewPluginPrimitiveSigner and KeySpecFromPlugin for raw-signature use cases

dallasd1:
Add two helper functions to the `signer` package so external callers like the notation CLI can produce raw plugin-backed signatures and certificate chains without going through the JWS/COSE envelope path.

The default behavior in the signing path for JWS/COSE signing remains unchanged. 

- `signer.NewPluginPrimitiveSigner`  returns a `signature.Signer` that calls `plugin.GenerateSignature` and returns raw signature bytes + certificate chain

- `signer.KeySpecFromPlugin`  calls `plugin.DescribeKey` and returns the decoded keySpec, enforcing that the plugin's response references the same keyID that was requested

### DIFF
--- a/signer/plugin.go
+++ b/signer/plugin.go
@@ -325,7 +325,14 @@ func parseCertChain(certChain [][]byte) ([]*x509.Certificate, error) {
 	return certs, nil
 }
 
-// PluginPrimitiveSigner implements signature.Signer
+// PluginPrimitiveSigner implements signature.Signer by delegating raw
+// signature generation to a plugin.SignPlugin.
+//
+// Unlike PluginSigner, which produces a complete JWS/COSE signature
+// envelope for OCI artifact signing, PluginPrimitiveSigner returns the raw
+// signature bytes plus the signing certificate chain. This is required
+// by signature envelopes that build their own structure around a
+// pre-computed signature, such as PKCS#7.
 type PluginPrimitiveSigner struct {
 	ctx          context.Context
 	plugin       plugin.SignPlugin
@@ -334,8 +341,13 @@ type PluginPrimitiveSigner struct {
 	keySpec      signature.KeySpec
 }
 
-// Sign signs the digest by calling the underlying plugin.
+// Sign signs the payload by calling the underlying plugin's GenerateSignature command
+// and returns the raw signature bytes together with the signing certificate
+// chain returned by the plugin.
 func (s *PluginPrimitiveSigner) Sign(payload []byte) ([]byte, []*x509.Certificate, error) {
+	if s == nil || s.plugin == nil {
+		return nil, nil, errors.New("PluginPrimitiveSigner not initialized: use NewPluginPrimitiveSigner")
+	}
 	// Execute plugin sign command.
 	keySpec, err := proto.EncodeKeySpec(s.keySpec)
 	if err != nil {
@@ -370,37 +382,57 @@ func (s *PluginPrimitiveSigner) Sign(payload []byte) ([]byte, []*x509.Certificat
 	return resp.Signature, certs, nil
 }
 
-// KeySpec returns the keySpec of a keyID by calling describeKey and do some
-// keySpec validation.
+// KeySpec returns the KeySpec that the signer was constructed with.
+// The keySpec is supplied at construction time (typically via KeySpecFromPlugin).
 func (s *PluginPrimitiveSigner) KeySpec() (signature.KeySpec, error) {
+	if s == nil || s.plugin == nil {
+		return signature.KeySpec{}, errors.New("PluginPrimitiveSigner not initialized: use NewPluginPrimitiveSigner")
+	}
 	return s.keySpec, nil
 }
 
 // NewPluginPrimitiveSigner creates a new PluginPrimitiveSigner that delegates
-// signing to a plugin. This is used for dm-verity PKCS#7 signing where raw
-// signature bytes are needed instead of JWS/COSE envelopes.
-func NewPluginPrimitiveSigner(ctx context.Context, p plugin.SignPlugin, keyID string, keySpec signature.KeySpec, pluginConfig map[string]string) *PluginPrimitiveSigner {
+// signing to a plugin. It is intended for callers that need raw signature
+// bytes (PKCS#7) rather than a JWS/COSE envelope.
+func NewPluginPrimitiveSigner(ctx context.Context, p plugin.SignPlugin, keyID string, keySpec signature.KeySpec, pluginConfig map[string]string) (signature.Signer, error) {
+	if p == nil {
+		return nil, errors.New("nil plugin")
+	}
+	if keyID == "" {
+		return nil, errors.New("keyID not specified")
+	}
+	if _, err := proto.HashAlgorithmFromKeySpec(keySpec); err != nil {
+		return nil, fmt.Errorf("invalid keySpec: %w", err)
+	}
 	return &PluginPrimitiveSigner{
 		ctx:          ctx,
 		plugin:       p,
 		keyID:        keyID,
 		keySpec:      keySpec,
 		pluginConfig: pluginConfig,
-	}
+	}, nil
 }
 
-// GetKeySpecFromPlugin retrieves the key specification from a plugin by calling DescribeKey.
-func GetKeySpecFromPlugin(ctx context.Context, p plugin.SignPlugin, keyID string, pluginConfig map[string]string) (signature.KeySpec, error) {
-	req := &plugin.DescribeKeyRequest{
+// KeySpecFromPlugin retrieves and validates the key specification from a
+// plugin by calling DescribeKey. It enforces that the plugin's response
+// references the same keyID that was requested.
+func KeySpecFromPlugin(ctx context.Context, p plugin.SignPlugin, keyID string, pluginConfig map[string]string) (signature.KeySpec, error) {
+	if p == nil {
+		return signature.KeySpec{}, errors.New("nil plugin")
+	}
+	if keyID == "" {
+		return signature.KeySpec{}, errors.New("keyID not specified")
+	}
+	resp, err := p.DescribeKey(ctx, &plugin.DescribeKeyRequest{
 		ContractVersion: plugin.ContractVersion,
 		KeyID:           keyID,
 		PluginConfig:    pluginConfig,
-	}
-
-	resp, err := p.DescribeKey(ctx, req)
+	})
 	if err != nil {
-		return signature.KeySpec{}, err
+		return signature.KeySpec{}, fmt.Errorf("failed to describe key %q: %w", keyID, err)
 	}
-
+	if resp.KeyID != "" && resp.KeyID != keyID {
+		return signature.KeySpec{}, fmt.Errorf("keyID in describeKey response %q does not match request %q", resp.KeyID, keyID)
+	}
 	return proto.DecodeKeySpec(resp.KeySpec)
 }
--- a/signer/plugin.go
+++ b/signer/plugin.go
@@ -162,7 +162,7 @@ func (s *PluginSigner) generateSignature(ctx context.Context, desc ocispec.Descr
 	logger := log.GetLogger(ctx)
 	logger.Debug("Generating signature by plugin")
 	genericSigner := GenericSigner{
-		signer: &pluginPrimitiveSigner{
+		signer: &PluginPrimitiveSigner{
 			ctx:          ctx,
 			plugin:       s.plugin,
 			keyID:        s.keyID,
@@ -325,8 +325,8 @@ func parseCertChain(certChain [][]byte) ([]*x509.Certificate, error) {
 	return certs, nil
 }
 
-// pluginPrimitiveSigner implements signature.Signer
-type pluginPrimitiveSigner struct {
+// PluginPrimitiveSigner implements signature.Signer
+type PluginPrimitiveSigner struct {
 	ctx          context.Context
 	plugin       plugin.SignPlugin
 	keyID        string
@@ -335,7 +335,7 @@ type pluginPrimitiveSigner struct {
 }
 
 // Sign signs the digest by calling the underlying plugin.
-func (s *pluginPrimitiveSigner) Sign(payload []byte) ([]byte, []*x509.Certificate, error) {
+func (s *PluginPrimitiveSigner) Sign(payload []byte) ([]byte, []*x509.Certificate, error) {
 	// Execute plugin sign command.
 	keySpec, err := proto.EncodeKeySpec(s.keySpec)
 	if err != nil {
@@ -372,6 +372,35 @@ func (s *pluginPrimitiveSigner) Sign(payload []byte) ([]byte, []*x509.Certificat
 
 // KeySpec returns the keySpec of a keyID by calling describeKey and do some
 // keySpec validation.
-func (s *pluginPrimitiveSigner) KeySpec() (signature.KeySpec, error) {
+func (s *PluginPrimitiveSigner) KeySpec() (signature.KeySpec, error) {
 	return s.keySpec, nil
+}
+
+// NewPluginPrimitiveSigner creates a new PluginPrimitiveSigner that delegates
+// signing to a plugin. This is used for dm-verity PKCS#7 signing where raw
+// signature bytes are needed instead of JWS/COSE envelopes.
+func NewPluginPrimitiveSigner(ctx context.Context, p plugin.SignPlugin, keyID string, keySpec signature.KeySpec, pluginConfig map[string]string) *PluginPrimitiveSigner {
+	return &PluginPrimitiveSigner{
+		ctx:          ctx,
+		plugin:       p,
+		keyID:        keyID,
+		keySpec:      keySpec,
+		pluginConfig: pluginConfig,
+	}
+}
+
+// GetKeySpecFromPlugin retrieves the key specification from a plugin by calling DescribeKey.
+func GetKeySpecFromPlugin(ctx context.Context, p plugin.SignPlugin, keyID string, pluginConfig map[string]string) (signature.KeySpec, error) {
+	req := &plugin.DescribeKeyRequest{
+		ContractVersion: plugin.ContractVersion,
+		KeyID:           keyID,
+		PluginConfig:    pluginConfig,
+	}
+
+	resp, err := p.DescribeKey(ctx, req)
+	if err != nil {
+		return signature.KeySpec{}, err
+	}
+
+	return proto.DecodeKeySpec(resp.KeySpec)
 }
--- a/signer/plugin.go
+++ b/signer/plugin.go
@@ -162,7 +162,7 @@ func (s *PluginSigner) generateSignature(ctx context.Context, desc ocispec.Descr
 	logger := log.GetLogger(ctx)
 	logger.Debug("Generating signature by plugin")
 	genericSigner := GenericSigner{
-		signer: &PluginPrimitiveSigner{
+		signer: &pluginPrimitiveSigner{
 			ctx:          ctx,
 			plugin:       s.plugin,
 			keyID:        s.keyID,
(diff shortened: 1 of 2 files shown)


signer/plugin.go
```diff
@@ -372,6 +372,35 @@ func (s *pluginPrimitiveSigner) Sign(payload []byte) ([]byte, []*x509.Certificat
...
+		ctx:          ctx,
+		plugin:       p,
+		keyID:        keyID,
+		keySpec:      keySpec,
+		pluginConfig: pluginConfig,
+	}
+}
+
+// GetKeySpecFromPlugin retrieves the key specification from a plugin by calling DescribeKey.
+func GetKeySpecFromPlugin(ctx context.Context, p plugin.SignPlugin, keyID string, pluginConfig map[string]string) (signature.KeySpec, error) {
```
dallasd1, line 393: This is needed for the CLI PR that will pull the signing type to include in the PKCS#7 envelope

bketelsen: the private getKeySpec validates the KeyID matching the requested keyID, but the public method doesn't. It should be validated like the existing private method does.

signer/plugin.go
```diff
@@ -372,6 +372,35 @@ func (s *pluginPrimitiveSigner) Sign(payload []byte) ([]byte, []*x509.Certificat
...
+	req := &plugin.DescribeKeyRequest{
+		ContractVersion: plugin.ContractVersion,
+		KeyID:           keyID,
+		PluginConfig:    pluginConfig,
+	}
+
+	resp, err := p.DescribeKey(ctx, req)
+	if err != nil {
+		return signature.KeySpec{}, err
```
bketelsen: wrap these errors in some context with fmt.Errorf("failed to describe key %q: %w", keyID, err)) to match the rest of plugin.go

signer/plugin.go
```diff
@@ -372,6 +372,35 @@ func (s *pluginPrimitiveSigner) Sign(payload []byte) ([]byte, []*x509.Certificat
...
 // keySpec validation.
-func (s *pluginPrimitiveSigner) KeySpec() (signature.KeySpec, error) {
+func (s *PluginPrimitiveSigner) KeySpec() (signature.KeySpec, error) {
 	return s.keySpec, nil
 }
+
+// NewPluginPrimitiveSigner creates a new PluginPrimitiveSigner that delegates
+// signing to a plugin. This is used for dm-verity PKCS#7 signing where raw
+// signature bytes are needed instead of JWS/COSE envelopes.
+func NewPluginPrimitiveSigner(ctx context.Context, p plugin.SignPlugin, keyID string, keySpec signature.KeySpec, pluginConfig map[string]string) *PluginPrimitiveSigner {
```
bketelsen: NewPluginSigner validates inputs and returns an error but this version does neither. It would silently accept a nil plugin and/or empty keyID and store them, which would cause a panic later.

signer/plugin_test.go
```diff
@@ -492,3 +492,53 @@ func basicSignTest(t *testing.T, ps *PluginSigner, envelopeType string, data []b
 	}
 	basicVerification(t, data, envelopeType, mockPlugin.certs[len(mockPlugin.certs)-1], &validMetadata)
 }
+
```
bketelsen, line 501: add test for nil plugin - see previous comment

bketelsen: and a test for empty keyID

signer/plugin_test.go
```diff
@@ -492,3 +492,53 @@ func basicSignTest(t *testing.T, ps *PluginSigner, envelopeType string, data []b
...
+	got, err := GetKeySpecFromPlugin(ctx, mp, "testKeyID", nil)
+	if err != nil {
+		t.Fatalf("GetKeySpecFromPlugin() error: %v", err)
+	}
+	if got != defaultKeySpec {
+		t.Fatalf("GetKeySpecFromPlugin() = %v, want %v", got, defaultKeySpec)
+	}
+}
+
```
bketelsen, line 618: Only happy path testing here. Need Sign() failure and at least one non-RSA key tested.

bketelsen (changes requested):
Some blockers - KeyID validation is dropped from new public method, no input validation for NewPluginPrimitiveSigner()

signer/plugin.go
```diff
@@ -325,8 +325,8 @@ func parseCertChain(certChain [][]byte) ([]*x509.Certificate, error) {
 	return certs, nil
 }
 
-// pluginPrimitiveSigner implements signature.Signer
-type pluginPrimitiveSigner struct {
+// PluginPrimitiveSigner implements signature.Signer
+type PluginPrimitiveSigner struct {
```
shizhMSFT: > Now that this is `PluginPrimitiveSigner`, every field is *almost* part of the public API by reflection / `unsafe` use. Keep them unexported (they already are), but please also confirm that `signer.PluginPrimitiveSigner{}` (zero value) cannot be misused — e.g., add a runtime guard in `Sign` and `KeySpec` that returns an error if `s.plugin == nil`. Otherwise external callers using `&PluginPrimitiveSigner{}` literals will hit nil-deref panics rather than clean errors.

signer/plugin.go
```diff
@@ -372,6 +372,35 @@ func (s *pluginPrimitiveSigner) Sign(payload []byte) ([]byte, []*x509.Certificat
...
+// signature bytes are needed instead of JWS/COSE envelopes.
+func NewPluginPrimitiveSigner(ctx context.Context, p plugin.SignPlugin, keyID string, keySpec signature.KeySpec, pluginConfig map[string]string) *PluginPrimitiveSigner {
+	return &PluginPrimitiveSigner{
+		ctx:          ctx,
+		plugin:       p,
+		keyID:        keyID,
+		keySpec:      keySpec,
+		pluginConfig: pluginConfig,
+	}
+}
```
shizhMSFT:
```suggestion
// NewPluginPrimitiveSigner creates a new PluginPrimitiveSigner that delegates
// signing to a plugin. This is intended for callers that need raw signature
// bytes (e.g., dm-verity PKCS#7) rather than a JWS/COSE envelope.
//
// The keySpec must match what the plugin reports via DescribeKey. Use
// KeySpecFromPlugin to obtain it, or construct via NewPluginSigner if you
// just need standard envelope signing.
func NewPluginPrimitiveSigner(ctx context.Context, p plugin.SignPlugin, keyID string, keySpec signature.KeySpec, pluginConfig map[string]string) (signature.Signer, error) {
	if p == nil {
		return nil, errors.New("nil plugin")
	}
	if keyID == "" {
		return nil, errors.New("keyID not specified")
	}
	if _, err := proto.HashAlgorithmFromKeySpec(keySpec); err != nil {
		return nil, fmt.Errorf("invalid keySpec: %w", err)
	}
	return &PluginPrimitiveSigner{
		ctx:          ctx,
		plugin:       p,
		keyID:        keyID,
		keySpec:      keySpec,
		pluginConfig: pluginConfig,
	}, nil
}
```
Three changes: (1) validate inputs the same way `NewPluginSigner` does, (2) return `signature.Signer` (interface) rather than the concrete type so we can refactor later, (3) return an error so callers must handle setup failure. Same pattern as `NewPluginSigner` for consistency.

shizhMSFT (commented):
> [!CAUTION]
> **AI-generated review.** This review was produced by an AI agent (Claude Opus 4.7, `shizh-reviewer` skill) on behalf of @shizhMSFT. It is posted as `COMMENT` (not `REQUEST_CHANGES`) — please weigh findings on their merits and treat them as starting points, not gating concerns.

Thanks @dallasd1. `notation-go` is more flexed than `notation-core-go`, but exporting an internal type into the public API of the most-imported library in the Notary Project ecosystem still deserves careful thought. A few concerns and one design question.

**Design question first**

> Do we actually need to export `pluginPrimitiveSigner`?

The stated reason is that the dm-verity flow lives in `notaryproject/notation` (the CLI) and needs a `signature.Signer`. Two alternatives that don't expand `notation-go`'s public API:

1. **Move the dm-verity command into a `signer/dmverity` subpackage of `notation-go`** and call the existing `pluginPrimitiveSigner` internally. The CLI then composes via a public function like `signer.NewDMVeritySigner(plugin, keyID, opts)` that returns `[]byte` directly. This keeps `pluginPrimitiveSigner` private.
2. **Add a single narrow public function** `signer.SignPrimitive(ctx, plugin, keyID, payload, pluginConfig) ([]byte, []*x509.Certificate, error)` that internally constructs and calls `pluginPrimitiveSigner`. The caller never holds the type; we can refactor later without breaking compat.

Both reduce the API surface from "a struct + constructor + helper + KeySpec method" to one function. Per `clig.dev` reasoning applied to libraries: **public API surface is a permanent contract**; we should add the smallest amount that solves the use case. Right now this PR exports more than is strictly needed (see comments below). Worth at least answering "did we consider these?" before merge.

**Concerns (if we proceed with the export)**

1. **`GetKeySpecFromPlugin` drops the `keyID` round-trip check.** The existing `PluginSigner.getKeySpec` (signer/plugin.go:155–157) compares `s.keyID` to `descKeyResp.KeyID` and rejects mismatches. The new helper does not. A malicious or buggy plugin can return a different key's spec; the caller will then sign with the wrong spec/algorithm. This is the kind of TOCTOU-adjacent issue that has bitten supply-chain tooling before.
2. **`NewPluginPrimitiveSigner` performs no input validation.** `NewPluginSigner` (signer/plugin.go:66–78) rejects `nil` plugin and empty keyID. The new constructor accepts both and only fails later, deep inside `Sign()`. It also accepts an arbitrary `signature.KeySpec` value from the caller without sanity-checking it — the suggestion below adds local validation via `proto.HashAlgorithmFromKeySpec`. Note: this is purely a "did you pass me a valid (Type, Size) pair?" check; it does **not** cross-validate the keySpec against what the plugin would report — for that, the caller must use `KeySpecFromPlugin` first.
3. **`*PluginPrimitiveSigner` returned from constructor.** Constructors that return concrete pointers commit us to every exported field/method/struct-tag. Return `signature.Signer` (the interface the type already implements). Less to maintain; same usefulness.

**Non-blocking design tension**

- **`context.Context` baked into the public struct.** The original code already does this internally, so exporting it doesn't *introduce* the anti-pattern — it propagates it. `Sign(payload []byte)` ignores its surroundings and uses `s.ctx`, so timeouts/cancellations from the calling goroutine cannot reach the plugin RPC. Resolving this properly (either passing `ctx` to `Sign(ctx, payload)` or hiding the ctx behind constructor scope) means changing `signature.Signer`, which is out of scope for this PR. Worth flagging as a tracking issue for the next signing-API revision.

**Naming**

- `GetKeySpecFromPlugin` → `KeySpecFromPlugin` (Go style: no `Get` prefix on getters; cf. https://go.dev/wiki/CodeReviewComments#getters)
- `NewPluginPrimitiveSigner` is fine, but consider `NewSignaturePrimitiveSigner` or just folding into `NewPluginSigner` as a mode flag. "PluginPrimitive" reads as two adjectives stacked; users will get it wrong.

**Verdict:** COMMENT (AI review only — not gating).

signer/plugin.go
```diff
@@ -372,6 +372,35 @@ func (s *pluginPrimitiveSigner) Sign(payload []byte) ([]byte, []*x509.Certificat
...
+		PluginConfig:    pluginConfig,
+	}
+
+	resp, err := p.DescribeKey(ctx, req)
+	if err != nil {
+		return signature.KeySpec{}, err
+	}
+
+	return proto.DecodeKeySpec(resp.KeySpec)
+}
```
shizhMSFT:
```suggestion
// KeySpecFromPlugin retrieves and validates the key specification from a plugin
// by calling DescribeKey. It enforces that the plugin's response references the
// same keyID that was requested; this guards against a misbehaving plugin
// returning a spec for a different key.
func KeySpecFromPlugin(ctx context.Context, p plugin.SignPlugin, keyID string, pluginConfig map[string]string) (signature.KeySpec, error) {
	if p == nil {
		return signature.KeySpec{}, errors.New("nil plugin")
	}
	if keyID == "" {
		return signature.KeySpec{}, errors.New("keyID not specified")
	}
	resp, err := p.DescribeKey(ctx, &plugin.DescribeKeyRequest{
		ContractVersion: plugin.ContractVersion,
		KeyID:           keyID,
		PluginConfig:    pluginConfig,
	})
	if err != nil {
		return signature.KeySpec{}, err
	}
	if resp.KeyID != keyID {
		return signature.KeySpec{}, fmt.Errorf("keyID in describeKey response %q does not match request %q", resp.KeyID, keyID)
	}
	return proto.DecodeKeySpec(resp.KeySpec)
}
```
Two changes:
1. Drop the `Get` prefix per Go style.
2. Add the same `keyID` round-trip validation that `PluginSigner.getKeySpec` performs (lines 155–157). Without it, a plugin can respond with a different key's spec and we'd silently sign with the wrong algorithm. This is the same defense-in-depth check `PluginPrimitiveSigner.Sign()` already does at lines 362–364 *after* signing — please mirror it *before* signing.

signer/plugin_test.go
```diff
@@ -492,3 +492,53 @@ func basicSignTest(t *testing.T, ps *PluginSigner, envelopeType string, data []b
...
+func TestGetKeySpecFromPlugin_Error(t *testing.T) {
+	ctx := context.Background()
+	mp := &mockPlugin{}
+	_, err := GetKeySpecFromPlugin(ctx, mp, "testKeyID", nil)
+	if err == nil {
```
shizhMSFT: > `&mockPlugin{}` returns `proto.DescribeKeyResponse{KeySpec: ks}` where `ks, _ := proto.EncodeKeySpec(p.keySpec)` — for a zero-valued `keySpec` this returns `{}`. The test passes today because `proto.DecodeKeySpec("")` errors, but the comment "expected error for empty keySpec" is misleading: we're really testing decode-side error handling, not the helper's input contract. Once the validation suggested above is added, please also add tests for: nil plugin, empty keyID, mismatched keyID round-trip, and a plugin whose `DescribeKey` itself errors. The `Error` test name should match the failure mode (`TestKeySpecFromPlugin_KeyIDMismatch`, etc.).

signer/plugin_test.go
```diff
@@ -492,3 +492,53 @@ func basicSignTest(t *testing.T, ps *PluginSigner, envelopeType string, data []b
...
+
+	// verify KeySpec
+	ks, err := s.KeySpec()
+	if err != nil {
+		t.Fatalf("KeySpec() error: %v", err)
+	}
+	if ks != defaultKeySpec {
+		t.Fatalf("KeySpec() = %v, want %v", ks, defaultKeySpec)
+	}
+
```
shizhMSFT, line 519: > Once `NewPluginPrimitiveSigner` returns `(signature.Signer, error)` per the suggestion above, please also assert it returns an error for `(nil, "key", spec, nil)` and `(plugin, "", spec, nil)`. Mirroring `NewPluginSigner`'s test coverage.

signer/plugin_test.go
```diff
@@ -166,7 +166,7 @@ func (p *mockPlugin) GenerateEnvelope(ctx context.Context, req *proto.GenerateEn
 			return nil, err
 		}
 
-		primitivePluginSigner := &pluginPrimitiveSigner{
+		primitivePluginSigner := &PluginPrimitiveSigner{
```
shizhMSFT: > nit: the local variable was `primitivePluginSigner` and is now `&PluginPrimitiveSigner{...}` — fine, but consider renaming the variable to `pluginPrimitiveSigner` (lowercase first letter for a local var) so the only thing that changed is the type, not the variable's spelling. Helps git blame.

signer/plugin_test.go
dallasd1: More failure tests and ECDSA tests added

signer/plugin_test.go
dallasd1: Tests for nil plugin and empty KeyID added

signer/plugin.go
dallasd1: NewPluginPrimitiveSigner now validates input and returns error. This update also returns the signature.Signer type, so there's no need to export pluginPrimitiveSigner

signer/plugin.go
dallasd1: Errors are now wrapped to match the rest of plugin.go

signer/plugin.go
dallasd1: KeySpecFromPlugin now validates the input and KeyIDs

signer/plugin.go
dallasd1: I've changed the return type to signature.Signer and unexported pluginPrimitiveSigner

signer/plugin.go
dallasd1: All three suggestions have been implemented

signer/plugin.go
dallasd1: GetKeySpecFromPlugin renamed to KeySpecFromPlugin. Validation checks have also been added

signer/plugin_test.go
dallasd1: Extra tests for nil plugin, empty KeyID, mismatched KeyID, DescribeKey have all been added

signer/plugin_test.go
dallasd1: pluginPrimitiveSigner has been unexported

signer/plugin.go
```diff
@@ -325,17 +325,29 @@ func parseCertChain(certChain [][]byte) ([]*x509.Certificate, error) {
...
 }
 
-// Sign signs the digest by calling the underlying plugin.
-func (s *pluginPrimitiveSigner) Sign(payload []byte) ([]byte, []*x509.Certificate, error) {
+// Sign signs the payload by calling the underlying plugin's GenerateSignature command
+// and returns the raw signature bytes together with the signing certificate
+// chain returned by the plugin.
+func (s *PluginPrimitiveSigner) Sign(payload []byte) ([]byte, []*x509.Certificate, error) {
+	if s == nil || s.plugin == nil {
+		return nil, nil, errors.New("PluginPrimitiveSigner not initialized: use NewPluginPrimitiveSigner")
```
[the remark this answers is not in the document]
dallasd1: Error strings have been expanded and are in lower case

signer/plugin.go
```diff
@@ -370,8 +382,57 @@ func (s *pluginPrimitiveSigner) Sign(payload []byte) ([]byte, []*x509.Certificat
...
 }
 
-// KeySpec returns the keySpec of a keyID by calling describeKey and do some
-// keySpec validation.
-func (s *pluginPrimitiveSigner) KeySpec() (signature.KeySpec, error) {
+// KeySpec returns the KeySpec that the signer was constructed with.
+// The keySpec is supplied at construction time (typically via KeySpecFromPlugin).
+func (s *PluginPrimitiveSigner) KeySpec() (signature.KeySpec, error) {
+	if s == nil || s.plugin == nil {
+		return signature.KeySpec{}, errors.New("PluginPrimitiveSigner not initialized: use NewPluginPrimitiveSigner")
```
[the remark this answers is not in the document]
dallasd1: Error strings have been expanded and are in lower case

signer/plugin.go
```diff
@@ -370,8 +382,57 @@ func (s *pluginPrimitiveSigner) Sign(payload []byte) ([]byte, []*x509.Certificat
...
+	}
+	resp, err := p.DescribeKey(ctx, &plugin.DescribeKeyRequest{
+		ContractVersion: plugin.ContractVersion,
+		KeyID:           keyID,
+		PluginConfig:    pluginConfig,
+	})
+	if err != nil {
+		return signature.KeySpec{}, fmt.Errorf("failed to describe key %q: %w", keyID, err)
+	}
+	if resp.KeyID != "" && resp.KeyID != keyID {
```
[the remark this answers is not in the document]
dallasd1: This has been updated to not allow an empty KeyID

dallasd1:
Thanks for the review @bketelsen! Your feedback should be addressed concerning the validation checks and test coverage.

signer/plugin_test.go
```diff
@@ -492,3 +501,174 @@ func basicSignTest(t *testing.T, ps *PluginSigner, envelopeType string, data []b
...
+	ctx := context.Background()
+	mp := newMockPlugin(defaultKeyCert.key, defaultKeyCert.certs, defaultKeySpec)
+
+	got, err := KeySpecFromPlugin(ctx, mp, "testKeyID", nil)
+	if err != nil {
+		t.Fatalf("KeySpecFromPlugin() error: %v", err)
+	}
+	if got != defaultKeySpec {
+		t.Fatalf("KeySpecFromPlugin() = %v, want %v", got, defaultKeySpec)
+	}
```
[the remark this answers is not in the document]
dallasd1: Fix added using describeKeyIdOverride along with a new test

dallasd1:
Thanks for the feedback @shizhMSFT! The comments have been addressed along with unexporting `PluginPrimitiveSigner` in favor of returning `signature.Signer`

signer/plugin_test.go
dallasd1: Error handling has been updated

PR #581, opened 2026-06-03
test: add unit tests for plugin payload attribute validation

### LINKED ISSUE #194: Unit test payload attribute validation
TODO: add detail

arpitjain099:
## What this does

Adds unit tests for the plugin payload attribute validation that #178 introduced. As noted in #194, those checks shipped without unit coverage because they were hard to exercise through the full signing flow without refactoring business logic. This PR tests the two helper functions directly instead, so no production code changes are needed.

Closes #194

## Tests added

In a new `signer/payload_test.go`:

`TestIsPayloadDescriptorValid` is table-driven and covers a plugin returning the same subject descriptor it was given (valid), appending an annotation (valid), and the tamper cases that the function exists to catch: a changed digest, a changed size, a changed mediaType, and an existing annotation being overridden. These map to the spec requirement that a plugin must not alter the payload subject during signing.

`TestAreUnknownAttributesAdded` covers a payload where every key is an allowed descriptor field (nothing reported), an unknown attribute injected inside the `targetArtifact` descriptor, an unknown attribute added at the payload top level, and both at once. This matches the rule from the plugin signing workflow that no extra top-level attributes may be added to the subject.

## Verification

- `go test ./signer/...` passes; statement coverage of `isPayloadDescriptorValid` and `areUnknownAttributesAdded` goes from 0 percent to full.
- `gofmt -l` and `go vet ./signer/...` are clean.
- I confirmed the tests are real by temporarily weakening each function (dropping the descriptor equality check, and returning no unknown attributes); the relevant subtests failed, then passed again once the code was restored. The diff in this PR touches only the new test file.


arpitjain099:
Still relevant, and the stale label caught me out rather than the change being abandoned. Going through Copilot's three points:

On rebinding `tc := tc` in the two table loops: this module is `go 1.23.0`, and Go 1.22 changed range variables to be per-iteration, so the rebinding is a no-op on any toolchain that can build this repo. `grep -r 'tc := tc'` also finds nothing in the tree, so adding it here would be the only instance and would read as cargo cult. Happy to add it if the project would rather keep the older idiom for consistency with `t.Parallel()` use elsewhere, but I would rather not add a line that does nothing.

On the `TestAreUnknownAttributesAdded` name: that one is fair. The function returns the slice of unknown keys rather than a bool, so the name promises the wrong shape. I can rename it to `TestAreUnknownAttributesAdded_ReportsUnknownAttributes`.

The tests themselves cover `areUnknownAttributesAdded` over the empty, exact-match, extra-key and missing-key cases, which is the validation path a plugin response goes through. Let me know if you want the rename pushed and I will do it, or if you would rather this be closed.

PR #582, opened 2026-07-08
fix: skip sub-directories when copying a plugin directory

arpitjain099:
While reading the plugin install path I noticed that `internal/file.CopyDirToDir` doesn't behave the way its comment describes. It's meant to copy only the top-level regular files of a directory, but the skip-subdirectories guard never fires.

The `WalkDir` callback names its path parameter `path`, so the check `d.Name() != filepath.Base(path)` compares the current entry's base name against itself and is always false. The walk therefore descends into every sub-directory and copies nested files, flattening them into the destination. `parsePluginFromDir` in the same package gets this right by comparing against the walk root, so the two disagree on what a plugin directory contains: `Install` validates the top level but then copies nested files too.

The fix compares the walked path against the source root, matching the existing `parsePluginFromDir` pattern. I added a regression test that fails before the change (a nested file gets copied) and passes after. Verified with `go test ./internal/file/... ./plugin/...` on linux/amd64.

Thanks for taking a look.


arpitjain099:
Commenting to keep this off the stale path, and to say the change is still current.

Re-checked `main` today: `internal/file/file.go:94` is unchanged.

```go
// skip sub-directories
if d.IsDir() && d.Name() != filepath.Base(path) {
    return fs.SkipDir
}
```

`path` is the callback's own parameter, so `d.Name()` and `filepath.Base(path)` are the same string for every entry and the guard never fires. `CopyDirToDir` therefore walks into sub-directories and flattens nested files into the destination, which is the opposite of what its doc comment says, and it disagrees with `parsePluginFromDir` in the same package about what a plugin directory contains.

The branch is rebased and mergeable, CI is green, and the diff is two files: the one-line comparison against the walk root, matching the `parsePluginFromDir` pattern, plus a regression test that fails without it.

No rush from my side, but a look from anyone on the review list would be appreciated, or tell me what you would like changed and I will turn it around.

